Issue #14, opened 2017-08-03
CPU-Manager-For-Kubernetes fails to launch when using kubeconfig multus


lmdaly:
I am trying to use CPU-Manager-For-Kubernetes with Mutlus Kubeconfig & TPR - however the annotations are not being propagated to pods created by CMK and Mutlus errors out dues to the lack of annotations. Below is the output of CMK cluster init & Mutlus. CMK works when running with Mutlus Delegates and if deployed manually with annotations in all pods.

CMK init-install-discover pods were not able to come up.
Following is the output while deploying CMK with flannel network object.
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
$ kubectl get pods
NAME                                          READY     STATUS     RESTARTS   AGE
kcm-cluster-init-pod                          1/1       Running    0          10m
kcm-init-install-discover-pod-192.168.14.38   0/2       Init:0/1   0          10m

$ kubectl describe pods kcm-init-install-discover-pod-192.168.14.38
Name:           kcm-init-install-discover-pod-192.168.14.38
Namespace:      default
Node:           192.168.14.38/192.168.14.38
Start Time:     Fri, 21 Jul 2017 10:04:57 -0400
Labels:         <none>
Status:         Pending
IP:
Controllers:    <none>
Init Containers:
  init:
    Container ID:
    Image:              96.118.23.36:8089/kcm:v0.6.1
    Image ID:
    Port:
    Command:
      /bin/bash
      -c
    Args:
      /kcm/kcm.py init --num-dp-cores=2 --num-cp-cores=1
    State:              Waiting
      Reason:           PodInitializing
    Ready:              False
    Restart Count:      0
    Volume Mounts:
      /etc/kcm from kcm-conf-dir (rw)
      /host/proc from host-proc (ro)
      /opt/bin from kcm-install-dir (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-dlbk8 (ro)
    Environment Variables:
      KCM_PROC_FS:      /host/proc
Containers:
  install:
    Container ID:
    Image:              96.118.23.36:8089/kcm:v0.6.1
    Image ID:
    Port:
    Command:
      /bin/bash
      -c
    Args:
      /kcm/kcm.py install
    State:              Waiting
      Reason:           PodInitializing
    Ready:              False
    Restart Count:      0
    Volume Mounts:
      /etc/kcm from kcm-conf-dir (rw)
[...]

PR #15, opened 2017-08-11
Bug fix to provide default CNI networking.

rkamudhan:
Refer CPU-Manager-For-Kubernetes fails to launch when using kubeconfig multus #14 for more info


swatisehgal:
Tested with CMK. Works fine!

PR #33, opened 2017-12-02
Use typed error when kubernetes network is missing

dcbw:
Matching error text is error-prone, so use a typed error instead.


rkamudhan:
Agreed to the changes.

rkamudhan:
@dcbw LGTM

Testing team comment:
I’ve checked out this PR and I had no issues with using it. AFAIK after looking at this commit the fix is more about some specifics of the Golang. I did not encounter any issue connected to the part of the code that has been changed here.

rkamudhan:
@dcbw Just now I noticed that we not doing the same implementation for the cmdDel

PR #34, opened 2017-12-05
reorganization and additions to the README

egernst:
Hey Kural et al,

I started playing with the multus CNI plugin late last week and learned quite a bit.  It's a pretty slick plugin and fills the gap in multi-network support nicely.

A few things became a lot more clear to me after using the plugin and trying it with both CRD and default conf.  Most of this was already documented in the readme, but not very linearly.  Anyway, I spent a bit of time and reordered and also introduced some concepts/topics at the start.   All the other content is still there, but some of it I moved into secondary documents which are referenced from the first, to keep it clearer, imo.  I'm hopeful that this'll make it easier for the next wave of people using Multus.

Hope you find this helpful.

--Eric

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,15 +63,40 @@ Please read [CNI](https://github.com/containernetworking/cni) for more informati
    <img src="doc/images/multus_cni_pod.png" width="1008" />
 </p>
 
-## Build
+## Getting Started
 
-**This plugin requires Go 1.8 to build.**
+### Requirements to build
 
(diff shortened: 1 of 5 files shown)


egernst:
A "todo" I didn't have time for was to maybe add a more advanced examples area where you can highlight changes that are needed to run some of these plugins in multus on multiple nodes (ie, host-local IPAM won't suffice).  I think that can be added later...

rkamudhan:
@egernst I will go through the readme files will merge it by next week

Issue #45, opened 2018-02-19
Error checking when using kubeconfig is wrong


One-ders:
The function getK8sNetwork returns an error. After commit 7066439e38ce3e442699b40c4bbe3485a2409fd7
it was changed so that the error checking is done on the type of the returned error, instead of the string value. In cmdAdd a check is done for the type '*NoK8sNetworkError', but the function returns NoK8sNetworkError (no star), so the statement will never be true. What will happend is that the else will be taken, and if there is a default net in the delegates list, the pod will be started without any anotated nets, just with the default network, this is wrong if you wnated the pod to have annotated networks, and made an error in the config.

Also the cmdDel does not even check for the type of error, but for the string value, that is not returned anymore, since the update of getK8sNetwork. 

Attaching a git diff for an example fix.

[multus-error.patch.tar.gz](https://github.com/Intel-Corp/multus-cni/files/1736427/multus-error.patch.tar.gz)

Fixes 1

Signed-off-by: anders.franzen@yahoo.se

rkamudhan:
Hi Anders. I will check the issue and try to fix it with your patch.

One-ders:
Great!!!

dougbtv:
Also put the `.patch` into a gist here @ https://gist.github.com/dougbtv/c58c08c9498070c2cdb9fd10dbc801e4

dougbtv:
Fixed in #60

PR #60, opened 2018-04-23
Fix error type for NoK8sNetworkError

s1061123:
This fix changes type type for NoK8sNetworkError due to error handling, in #45.
Currently NoK8sNetworkError is not pointer type, just a string, hence we never
get *NoK8sNetworkError. This fix changes it based @One-ders patch.

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -336,15 +336,17 @@ func cmdDel(args *skel.CmdArgs) error {
 
 	if in.Kubeconfig != "" {
 		podDelegate, r := k8s.GetK8sNetwork(args, in.Kubeconfig)
-		if r != nil && r.Error() == "nonet" {
-			nopodnet = true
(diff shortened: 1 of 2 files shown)


multus/multus.go
```diff
@@ -336,17 +336,19 @@ func cmdDel(args *skel.CmdArgs) error {
 
 	if in.Kubeconfig != "" {
 		podDelegate, r := k8s.GetK8sNetwork(args, in.Kubeconfig)
-		if r != nil && r.Error() == "nonet" {
-			nopodnet = true
-			if !defaultcninetwork {
-				return fmt.Errorf("Multus: Err in getting k8s network from the poc spec, check the pod spec or set delegate for the default network, Refer the README.md: %v", r)
+		if r != nil {
```
rkamudhan: Excellent optimization :smile:

PR #67, opened 2018-04-25
Examples for CRD w/ Defacto Standard

dougbtv:
Notably -- this should be considered as just "a working example" I believe that we should refine this later with what we agree should be the environment for the demo. However, this is (generally speaking) the process I've been using to make sure everything is in working order.

Includes:

* An `./examples` directory with a README showing process to use the examples.
* An update to the primary README with a link to the examples directory.
* Reference config file, reference Multus + Flannel deployment daemonset (with RBACs).
* Example clusterrole to give access to the `system:node:*` users.
* Example CRD and CRD objects.
* Example pod spec with an annotation.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,62 @@
+# Example Configurations & Pod Specs
+
+In the `./examples` folder some example configurations are provided for using Multus, especially with CRDs, and doubly so in reference to their usage with the [defacto standard for CRDs](https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit) as proposed by the Network Plumbing Working Group.
+
+## Examples overview
+
+Generally, the examples here show a setup using Multus with CRD support. The examples here demonstrate a setup with Multus as the meta-plugin used by Kubernetes, and delgating to either Flannel (which will be the default pod network), or to macvlan. The CRDs are intended to be alignment with the defacto standard.
+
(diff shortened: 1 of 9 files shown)


examples/README.md
```diff
@@ -0,0 +1,62 @@
...
+    - Refer to the primary README.md for more details on compilation.
+* Allow `system:node` access to enable Multus to pull CRD objects.
+    - See "RBAC configuration section below for details."
+* Create the Flannel + Multus setup with the daemonset provided
+    - As in: `kubectl create -f multus-with-flannel.yml`
+    - Optionally, verify that the `/etc/cni/net.d/*.conf` exists on each node.
+* Create the CRDs
+    - Create the CRD itself, `kubectl create -f crd.yml`
+    - Create the network attachment configurations (i.e. CNI configurations packed into CRD objects)
+        + `kubectl create -f flannel-conf.yml`
```
s1061123: Looking above `multus-with-flannel.yml`, this flannel conf seems to be second one, correct?
If we only use macvlan for secondary as `sample-pod.yml`, we could remove `flannel-conf.yml`.

How about to have another sidecar network yaml, like vlan?
I will send vlan CNI config offline...

rkamudhan:
LGTM, what about rebase and merge it to the dev/network-plumbing-working-group-crd-change branch itself ?

dougbtv:
Oops! I had the wrong target in the PR, thanks for the heads up @rkamudhan 

@s1061123 , I'll add another different config -- good eye, the flannel-conf is redundant.

rkamudhan:
@dougbtv It looks to me, Can we merge this branch ?

dougbtv:
@rkamudhan -- sounds like a good plan, I still want to clean up the example to reduce some redundancy, but, I'm going to merge for now. Thanks Kural!

PR #71, opened 2018-05-03
[bugfix][docs] Updates examples to match spec criteria

### LINKED ISSUE #70: Old config details commited by mistake
https://github.com/Intel-Corp/multus-cni/blob/696f863d93215a765219ddd7044e64405d735684/examples/flannel-conf.yml#L6-L12

dougbtv:
Here's updates to the examples that are compatible with the "3 modes of configuration operation" as recently modified. 

Fixes #70

### DIFF
--- a/examples/flannel-conf.yml
+++ b/examples/flannel-conf.yml
@@ -2,11 +2,11 @@ apiVersion: "kubernetes.cni.cncf.io/v1"
 kind: Network
 metadata:
   name: flannel-conf
-plugin: flannel
-args: '[
-        {
-                "delegate": {
-                        "isDefaultGateway": true
-                }
-        }
-]'
+spec: 
+  config: '{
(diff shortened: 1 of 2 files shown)


examples/flannel-conf.yml
```diff
@@ -2,11 +2,11 @@ apiVersion: "kubernetes.cni.cncf.io/v1"
...
-        {
-                "delegate": {
-                        "isDefaultGateway": true
-                }
-        }
-]'
+spec: 
+  config: '{
+    "cniVersion": "0.3.0",
+    type: "flannel",
```
rkamudhan: It is not a valid json. Nit-pick: Please use the https://jsonlint.com/  for checking the documents

dougbtv: Thanks for the eye! I ran it through a linter after a fix, too.

PR #72, opened 2018-05-03
Demo examples including updates

dougbtv:
@s1061123 I went ahead and added your demo files into a `./examples/demo` directory -- but, I also went ahead and updated them so that they match the spec section that looks like:

```
metadata:
  name: flannel-conf
spec: 
  config: '{
            "cniVersion": "0.3.0",
```

E.g. now the `plugin:` top level element is gone, and the `"type": ` element is under `spec.config`. 

Tomo, feel free to commit to this branch to show the changes, or, we can later delete this branch if you want to do it another way.

*note* Don't merge this yet, just for discussion of the demo examples for now, can merge later if we keep this branch, or close out the PR if we don't need it.

### DIFF
--- a/examples/npwg-demo-1/01_crd.yml
+++ b/examples/npwg-demo-1/01_crd.yml
@@ -1,0 +1,22 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  # name must match the spec fields below, and be in the form: <plural>.<group>
+  name: networks.kubernetes.cni.cncf.io
+spec:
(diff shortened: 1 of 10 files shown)


s1061123:
Feng mentioned that we should rename 'demo' to 'demo1' or some for another demo and I agree with him.
I will rename it.

dougbtv:
Sounds like a good idea

s1061123:
Now it is ready, so removing 'WIP' tag.

rkamudhan:
@s1061123 @dougbtv I will look into it..

s1061123:
I've rebased and make it as one commit.

PR #73, opened 2018-05-03
Comma-delimited format validation

dougbtv:
Only modifies the existing process slightly, but, the gist: 

1. assumes that it's JSON
2. if it's not, validate the format against format as specified in de-facto standard
3. Send an error message if it's not, explain it couldn't be parsed

### DIFF
--- a/k8sclient/k8sclient.go
+++ b/k8sclient/k8sclient.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"regexp"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


k8sclient/k8sclient.go
```diff
@@ -99,8 +100,21 @@ func parsePodNetworkObject(podnetwork string) ([]map[string]interface{}, error)
...
 	// Parse the podnetwork string, and assume it is JSON.
 	if err := json.Unmarshal([]byte(podnetwork), &podNet); err != nil {
-		// If the JSON parsing fails, assume it is comma delimited.
+
+		// Check and see if it matches the specification for the comma delimited format.
+		// "Valid attachment names must be comprised of units of the DNS-1123 label format"
+		// [a-z0-9]([-a-z0-9]*[a-z0-9])?
+		// And we allow at (@), comma (,), and forward slash (/)
+		// But it must start and end alphanumerically.
+		matched, _ := regexp.MatchString("^[a-z0-9]([-a-z0-9/@,]*[a-z0-9])?$", podnetwork)
```
rkamudhan: podnetwork := "netA,netB,example-namespace/attachment-name" is valid network annotation. Please correct me, if I am miss understood concept here.

https://play.golang.org/p/tllaAF8S8K- just started testing it

s1061123: We also need to take care of space ' ', like "netA, netB". After here, parsePodNetworkObjectName() trims spaces, at that time, we don't do that. So regexp should take care of space as well as '@,/', otherwise user cannot use space between network name.

s1061123: In addition, should [-a-z] be [a-z]?
There is no notation [-a-z] in https://github.com/google/re2/wiki/Syntax

dougbtv:
@rkamudhan in the spec doc as it stands, the proposal I gave was based on the DNS1123 label format, matching what kube uses for validating namespaces, with the addition of `/` & `@` (@ mentioned only in the addendum/appendix, I think) allowed to separate "units" of those labels. I changed the examples that were out of date with upper case to have the lowercase.

k8sclient/k8sclient.go
dougbtv: @s1061123 regarding `[-a-z]` I based it on [this validation in kube](https://github.com/kubernetes/kubernetes/blob/8d7d7a5e0d4d7e75f5a860574346944b8cc0fc43/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L107) -- I believe that (at least in PCRE) you can put a `-` as the first or last item in a character class in order to have it be a literal dash, instead of being a range, or you can escape it.

Good call regarding the spaces -- next change I'm submitting will have it split by commas, and then I'm going to let your routine for parsePodNetworkObjectName and then validate the items therein as I realized that it became tricky to validate good looking names, e.g. this regex would've also accepted bad input like `foo@@@@bar////quux`, too.

dougbtv:
Latest push takes into account:

* Allowing spaces between elements separated by commas (thanks Tomo)
* Validates the DNS1123 label format for each element as separated by comma, slash & @ by moving the validation logic into the `parsePodNetworkObjectName()` and leveraging work done there
  * Previous commit would've allowed some bad formatting like `foo@@@bar//quux`
* Does not yet allow upper/lower case, added item to spec for discussion there.

s1061123:
Looks good to me, Doug.
Thank you for your incorporation!

PR #78, opened 2018-05-10
Refine multus config file (masterplugin and always_use_default)

s1061123:
This fix removes 'masterplugin' and 'always_use_default' from multus config file. default network, defined in
/etc/cni/net.d/*conf should be master and this network should be used as default from network plumbing working group draft.

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -71,20 +71,16 @@ func loadNetConf(bytes []byte) (*types.NetConf, error) {
 		netconf.ConfDir = defaultConfDir
 	}
 
-	if netconf.UseDefault {
-		if netconf.Kubeconfig == "" || !defaultcninetwork {
-			return nil, fmt.Errorf(`If you have set always_use_default, you must also set the delegates & the kubeconfig, refer to the README`)
-		}
-		return netconf, nil
-	}
-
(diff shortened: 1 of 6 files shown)


s1061123:
Removing 'masterplugin' from the code will be done as a part of refactoring.

multus/multus.go
```diff
@@ -71,21 +71,17 @@ func loadNetConf(bytes []byte) (*types.NetConf, error) {
...
+	if netconf.Kubeconfig == "" || !defaultcninetwork {
+		return nil, fmt.Errorf(`You must also set the delegates & the kubeconfig, refer to the README`)
 	}
 
 	if len(netconf.Delegates) == 0 && !defaultcninetwork {
 		return nil, fmt.Errorf(`delegates or kubeconfig option is must, refer README.md`)
 	}
 
+	// default network in multus conf as master plugin
+	netconf.Delegates[0]["masterplugin"] = true
```
dcbw: So this PR makes the first delegate in the Multus CNI JSON config the masterplugin that returns its IP to Kubernetes, right?

s1061123: Correct. First network in /etc/cni/net.d/*.conf will be masterplugin (e.g. default network).

rkamudhan:
Verifed and test.. LGTM.

Issue #86, opened 2018-07-04
Adding NPWG network status in Multus


rkamudhan:
**From the NPWG spec document
Section 5: Network Attachment Status Annotation**
To ensure that the result of a network attachment is available via the Kubernetes API, the implementation may publish the result of the attachment operation to an annotation on the pod object that requested the attachment.

The annotation’s name shall be `kubernetes.v1.cni.cncf.io/network-status` and its value shall be a JSON-encoded list of maps

PR #87, opened 2018-07-04
Adding network status features in multus

rkamudhan:
For more information refer #86

**How to test the PR ?**

1. Create a cni conf files, as below
```
# cat /etc/cni/net.d/00-multus.conf
{
	"name": "defaultnetwork",
	"type": "multus",
	"kubeconfig": "/etc/kubernetes/node-kubeconfig.yaml",
	"delegates": [{
		"cniVersion": "0.3.0",
		"name": "defaultnetwork",
		"type": "flannel",
		"isDefaultGateway": true
	}],
	"confDir": "/etc/cni/multus/net.d"
}
```
2. Create a network object `test-ptp-network.yaml` as below and execute `kubectl create -f test-ptp-network.yaml`
```
# cat test-ptp-network.yaml
apiVersion: "kubernetes.cni.cncf.io/v1"
kind: Network
metadata:
  name: test-ptp-network
spec:
  config: '{
    "cniVersion": "0.3.0",
    "name": "test-ptp-network",
    "type": "ptp",
    "ipam": {
      "type": "host-local",
      "subnet": "10.1.1.0/24"
    },
    "dns": {
                "nameservers": [ "10.1.1.1", "8.8.8.8" ]
    }
}'
```
3. create a pod `iamgroot` with following spec and execute `kubectl create -f  busybox-ptp.yaml`
```
# cat busybox-ptp.yaml
apiVersion: v1
kind: Pod
metadata:
  name: iamgroot
  annotations:
    kubernetes.v1.cni.cncf.io/networks: test-ptp-network

spec:
  containers:
  - name: iamgroot
    image: busybox
    command: ["top"]
    stdin: true
    tty: true
    ports:
    - containerPort: 81
      protocol: TCP
```
4. Pod annotation `kubernetes.v1.cni.cncf.io/networks-status` is created as below
```
# kubectl get pod iamgroot -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.v1.cni.cncf.io/networks: test-ptp-network
    kubernetes.v1.cni.cncf.io/networks-status: |-
      [{
          "name": "defaultnetwork",
          "interface": "eth0",
          "ips": [
              "192.168.49.2"
          ],
          "mac": "76:54:2d:ab:0a:f9",
          "default": true,
          "dns": {}
      },{
          "name": "test-ptp-network",
          "interface": "net1",
          "ips": [
              "10.1.1.200"
          ],
          "mac": "d2:0c:d3:e5:4a:b0",
          "dns": {
              "nameservers": [
                  "10.1.1.1",
                  "8.8.8.8"
              ]
          }
      }]
  creationTimestamp: 2018-07-04T22:59:45Z
  name: iamgroot
  namespace: default
  resourceVersion: "4002364"
  selfLink: /api/v1/namespaces/default/pods/iamgroot
  uid: f1cbd889-7fdd-11e8-8236-408d5c537d27
spec:
  containers:
  - command:
    - top
    image: busybox
    imagePullPolicy: Always
    name: iamgroot
    ports:
    - containerPort: 81
      protocol: TCP
    resources: {}
    stdin: true
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    tty: true
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: default-token-jsfpv
      readOnly: true
  dnsPolicy: ClusterFirst
  nodeName: bla..bla..bla..
  restartPolicy: Always
  schedulerName: default-scheduler
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
  volumes:
  - name: default-token-jsfpv
    secret:
      defaultMode: 420
      secretName: default-token-jsfpv
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2018-07-04T22:59:45Z
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: 2018-07-04T23:00:00Z
    status: "True"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: 2018-07-04T22:59:45Z
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: docker://499abdaff62a709b80f94d887d849f9191051eb377fc3a8ed1b1c0ae45f714f6
    image: busybox:latest
    imageID: docker-pullable://busybox@sha256:141c253bc4c3fd0a201d32dc1f493bcf3fff003b6df416dea4f41046e0f37d47
    lastState: {}
    name: iamgroot
    ready: true
    restartCount: 0
    state:
      running:
        startedAt: 2018-07-04T22:59:59Z
  hostIP: 10.237.222.58
  phase: Running
  podIP: 192.168.49.2
  qosClass: BestEffort
  startTime: 2018-07-04T22:59:45Z
```


rkamudhan:
@s1061123 @dcbw @dougbtv Please check the PR, and provide your feedback :eyes: :eyes:

k8sclient/k8sclient.go
```diff
@@ -69,6 +79,54 @@ func createK8sClient(kubeconfig string) (KubeClient, error) {
...
+	_, err = setPodNetworkAnnotation(k.Client, k.Podnamespace, pod, ns)
+	if err != nil {
+		return fmt.Errorf("SetNetworkStatus: failed to update the pod %v in out of cluster comm: %v", k.Podname, err)
+	}
+
+	return nil
+}
+
+func setPodNetworkAnnotation(client KubeClient, namespace string, pod *v1.Pod, ns string) (*v1.Pod, error) {
+	//pod annotations is empty
```
s1061123: Could you add a bit detail like "allocate allocation if pod annotation is empty"?

k8sclient/k8sclient.go
```diff
@@ -69,6 +79,54 @@ func createK8sClient(kubeconfig string) (KubeClient, error) {
...
+
+	_, err = setPodNetworkAnnotation(k.Client, k.Podnamespace, pod, ns)
+	if err != nil {
+		return fmt.Errorf("SetNetworkStatus: failed to update the pod %v in out of cluster comm: %v", k.Podname, err)
+	}
+
+	return nil
+}
+
+func setPodNetworkAnnotation(client KubeClient, namespace string, pod *v1.Pod, ns string) (*v1.Pod, error) {
```
s1061123: How about to change ns to status (to make it more clear), because sometime we may misunderstand that ns=namespace...?

rkamudhan: ok cool, Thanks for the suggest, I will make the changes.

k8sclient/k8sclient.go
rkamudhan: ok cool, Thanks for the suggest, I will make the changes.

rkamudhan:
@s1061123, @dougbtv  @dcbw: Please let me know, whether the PR is all good to merge to the `dev` branch..

rkamudhan:

dougbtv:
Just FYI I think the "unexpected operation" comes from admission.go @ [this line](https://github.com/kubernetes/kubernetes/blob/master/plugin/pkg/admission/noderestriction/admission.go#L219)

rkamudhan:
@dougbtv @s1061123 Any update on this PR status ?

k8sclient/k8sclient.go
```diff
@@ -69,6 +79,54 @@ func createK8sClient(kubeconfig string) (KubeClient, error) {
...
+
+func setPodNetworkAnnotation(client KubeClient, namespace string, pod *v1.Pod, networkstatus string) (*v1.Pod, error) {
+	//if pod annotations is empty, make sure it allocatable
+	if len(pod.Annotations) == 0 {
+		pod.Annotations = make(map[string]string)
+	}
+
+	pod.Annotations["kubernetes.v1.cni.cncf.io/networks-status"] = networkstatus
+
+	p, err := client.SetPod(namespace, pod)
```
dcbw: Instead of doing this, I would use the Kube util ExponentialBackoff() to try updating a couple times.  Just doing it once is somewhat error-prone in that if anything else updates the Pod object your update will get rejected.  Instead we can use RetryOnConflict:

```
import (
	"k8s.io/client-go/util/retry"
)


func (d *defaultKubeClient) UpdatePodStatus(pod *v1.Pod) (*v1.Pod, error) {
	return d.client.Core().Pods(pod.Namespace).UpdateStatus(pod)
}


	pod := pod.DeepCopy()
	var err error
	if resultErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
		if err != nil {
			// Re-get the pod unless it's the first attempt to update
			pod, err = client.GetPod(namespace, name)
			if err != nil {
				return err
			}
		}

		pod, err = client.UpdatePodStatus(pod)
		return err
	}); resultErr != nil {
		return nil, fmt.Errorf("status update failed for pod %s/%s: %v", namespace, name, resultErr)
	}
	return pod, nil
```

k8sclient/k8sclient.go
```diff
@@ -69,6 +79,54 @@ func createK8sClient(kubeconfig string) (KubeClient, error) {
...
+	return nil
+}
+
+func setPodNetworkAnnotation(client KubeClient, namespace string, pod *v1.Pod, networkstatus string) (*v1.Pod, error) {
+	//if pod annotations is empty, make sure it allocatable
+	if len(pod.Annotations) == 0 {
+		pod.Annotations = make(map[string]string)
+	}
+
+	pod.Annotations["kubernetes.v1.cni.cncf.io/networks-status"] = networkstatus
```
dcbw: kubernetes->k8s now...

multus/multus.go
```diff
@@ -221,7 +242,7 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 	}
 
 	if in.Kubeconfig != "" {
-		delegates, err := k8s.GetK8sNetwork(args, in.Kubeconfig, kubeClient, in.ConfDir)
+		delegates, _, err := k8s.GetK8sNetwork(args, in.Kubeconfig, kubeClient, in.ConfDir)
```
dcbw: We should probably remove the status annotation here, no?  The pod object may still be alive and in the API, but the annotation should be removed because the networks won't apply.

multus/multus.go
```diff
@@ -190,6 +193,16 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
 		if delegate.MasterPlugin || result == nil {
 			result = tmpResult
 		}
+
+		//create the network status, only in case Multus as kubeconfig
+		if n.Kubeconfig != "" {
+			delegateNetStatus, err := types.LoadNetworkStatus(tmpResult, delegate.Name, delegate.MasterPlugin)
+			if err != nil {
+				return nil, fmt.Errorf("Multus: Err rethe networks status: %v", err)
```
dcbw: Not quite sure what you meant with "rethe" here; maybe "Multus: error loading the network status"?

multus/multus.go
```diff
@@ -190,6 +193,16 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
...
+			}
+
+			n.AddNetworkStatus(delegateNetStatus)
```
dcbw: Instead of adding a field to our NetConf, since all this is in cmdAdd, would it be simpler to just declare `var netStatus    []*NetworkStatus` at the top of cmdAdd(), and append each status to it, then check later for `if len(netStatus) > 0 { <update pod status>`?

k8sclient/k8sclient.go
rkamudhan: @dcbw I was exactly looking for some solution like that. I will try this.

k8sclient/k8sclient.go
rkamudhan: Got it

multus/multus.go
rkamudhan: yes, I got it .. I am using `AddNetworkStatus` only in the `cmdAdd()` will include the netstatus length checking along with `kubeconfig`

multus/multus.go
```diff
@@ -180,6 +180,7 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
 	}
 
 	var result, tmpResult cnitypes.Result
+	var netStatus []*types.NetworkStatus
```
rkamudhan: removed the usage of `AddNetworkStatus()`

glide.yaml
```diff
@@ -18,10 +18,21 @@ import:
...
+  version: 072894a440bdee3a891dea811fe42902311cd2a3
+  subpackages:
+  - core/v1
 - package: k8s.io/client-go
-  version: 4a3ab2f5be5177366f8206fd79ce55ca80e417fa
+  version: 7d04d0e2a0a1a4d4a1cd6baa432a2301492e4e65
   subpackages:
   - kubernetes
   - tools/clientcmd
+  - util/retry
```
rkamudhan: Hi @dcbw ,

I am trying this glide update issue for nearly 10 hours :confused: from yesterday, no luck :disappointed: (It must be a small issue or silly mistake from my side, I am missing something here) . 

I am getting exactly same error, 
Setting version for golang.org/x/time to f51c12702a4d776e4c1fa9b0fabab841babae631.
[ERROR] Error scanning k8s.io/client-go/pkg/api/v1: open /root/.glide/cache/src/https-k8s.io-client-go/pkg/api/v1: no such file or directory

only 3 package - k8s.io/apimachinery, k8s.io/api and k8s.io/client-go is used in our project.

Please check the concept correctness for updating the pod in Multus.

Once I solved this issue. I can quickly compile the code and merge it.

s1061123:
@rkamudhan I found that some pods (e.g. coredns) running under namespace:kube-system failed to launch due to permission.

```
Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "edd02842b2f5de4f4c9593806dfafd021689a2af905bda575027e914a01aa457" network for pod "coredns-78fcdf6894-25pn7": NetworkPlugin cni failed to set up pod "coredns-78fcdf6894-25pn7_kube-system" network: Multus: Err set the networks status: SetNetworkStatus: failed to update the pod coredns-78fcdf6894-25pn7 in out of cluster comm: setPodNetworkAnnotation: failed to set the pod coredns-78fcdf6894-25pn7 in out of cluster comm: pods "coredns-78fcdf6894-25pn7" is forbidden: User "system:node:kube-master" cannot update pods in the namespace "kube-system"
```

I suppose we could fix as follows:

Option A) We skip to add network status into kube-system pods
  I prefer this option because we don't touch current k8s.

Option B) We add permission "system:node:kube-master" to modify under namespace:kube-system
  We should consider some security stuff if we do so, I guess.

I suppose we could make new issue for that and fix it later (after its merge).

rkamudhan:
@s1061123, I am ok with both the option. Right now the glide issue is not fixed.. So I can't compile this code, run now.. Is there, I am doing some silly mistake in the glide ?

glide.yaml
s1061123: I suppose that glide does 'deep update', including Godeps in vendoring pkgs. `glide update --quick` does quick update, just in Glide.yaml so we don't see any error.

dcbw:
@rkamudhan sent you an email with a resolution to the glide issue.  My suggested fix is the top two commits in https://github.com/dcbw/multus-cni/commits/dev/npwg-multus-network-status-feature (based off your branch)

dougbtv:
I'm trying to replicate the glide issue, and I'm running into something that looks at least remotely similar...

```
$ glide --version
glide version v0.13.1
$ rm -Rf vendor
$ glide update --strip-vendor
[... snip ...]
[ERROR] Error scanning k8s.io/client-go/pkg/api/v1: cannot find package "." in:
  /home/doug/.glide/cache/src/https-k8s.io-client-go/pkg/api/v1
[... snip ...]
[ERROR] Failed to retrieve a list of dependencies: Error resolving imports
$ echo $?
1
```

I also tried to change the version like so:

```
- package: k8s.io/client-go
  version: v8.0.0
```

s1061123:
Hi all,

Trying with `glide update --quick` works without errors, `glide update --quick` updates only in glide.yaml hence I guess that vendor package's dependency makes some error.

In near future, we should replace it with deps, so I suppose it could be okey. What do you think about it?

dcbw:
> [... snip ...] [ERROR] Error scanning k8s.io/client-go/pkg/api/v1: cannot find package "." in: /home/doug/.glide/cache/src/https-k8s.io-client-go/pkg/api/v1 [... snip ...]

@dougbtv I think you need to update k8sclient/k8sclient.go and testing/testing.go to stop using client-go/pkg/api/v1 since in recent versions of client-go that moved to k8s.io/api/v1

rkamudhan:
@dcbw I added your branch code along with unsetting the network status. 
@dougbtv @s1061123  skipping pod network status in `kube-system` namespace

rkamudhan:
@dougbtv @s1061123 @dcbw Please test it.

dcbw:
@rkamudhan LGTM, a couple things we can adjust later.

rkamudhan:
@dougbtv @s1061123 @dcbw @fepanResolved the conflict in the network status with latest changes. Please review it. We will merge it and release multus v3.0

Issue #93, opened 2018-07-09
Including build version and CNI spec support, Network Object spec version in Multus binary


rkamudhan:
Feature enhancement for the cloud admins to include version details in binary.

s1061123:
#206 adds the version, so close it.

Issue #92, opened 2018-07-09
Updating build script and adding release script for Multus


rkamudhan:
Expectation to develop release script for different arch.

s1061123:
It is already added I suppose, hence I close issue. Please open again if it isn't.

PR #94, opened 2018-07-10
adding release script for multus repo

rkamudhan:
See more info in #92 and #93

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_script:
 #  - gocyclo -over 15 ./multus
 
 script:
-  - ./build
+  - ./build.sh
   - sudo ./test.sh
   - mkdir -p ${TRAVIS_BUILD_DIR}/dist
   - tar cvfz ${TRAVIS_BUILD_DIR}/dist/multus-cni_amd64.tar.gz --warning=no-file-changed --exclude="dist" --exclude="vendor" .
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ script:
 before_deploy:
   - go get -u github.com/laher/goxc
   - mkdir -p $TRAVIS_BUILD_DIR/dist
-  - goxc -d=$TRAVIS_BUILD_DIR/dist -pv=$TRAVIS_TAG -bc=linux -tasks=clean-destination,xc,archive,rmbin
+  - goxc -d=$TRAVIS_BUILD_DIR/dist -pv=$TRAVIS_TAG -bc=linux -tasks=clean-destination,xc,archive,rmbin -build-ldflags="-extldflags -static -X main.version=$TRAVIS_TAG"
 
 deploy:
   provider: releases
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -23,7 +23,7 @@ before_script:
 #  - gocyclo -over 15 ./multus
 
 script:
-  - ./build
+  - ./build.sh
```
s1061123: In line 36 of .travis.yml, travis packages tgz that is built by 'before_script:' section and put it in github release page and we should change this script too, otherwise tgz file in github is not changed.

s1061123: Okey. Let me clarify...

In [.travis.yml:36-46](https://github.com/intel/multus-cni/blob/dev/network-plumbing-working-group-crd-change/.travis.yml#L36-L46), I've wrote script for travisCI. This script is invoked only tagging as matched '^v[0-9].*$', like v0.2.1. 

This travis script will upload files, that is built at TravisCI's build phase (*1), to github release. So in https://github.com/intel/multus-cni/releases/tag/v2.0,  multus-cni_v2.0_linux_386.tar.gz,  multus-cni_v2.0_linux_amd64.tar.gz and  multus-cni_v2.0_linux_arm.tar.gz are automatically put by TravisCI.

(*1): [.travis.yml:31-41](https://github.com/intel/multus-cni/blob/dev/network-plumbing-working-group-crd-change/.travis.yml#L31-L34) builds multus binary tgz for x86, x86_64 and arm for each commit. This binary is used at above travis upload script and also put to [snapshot](https://bintray.com/redhat-nfvpe/multus-cni-crd-snapshots/snapshot) in bintray.com every time.

Hence I suppose your script, 'release.sh' can be replaced with above script, otherwise we put two binaries into release page (one is from travisCI as I mentioned above and another is from release.sh).

release.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -xe
+
+SRC_DIR="${SRC_DIR:-$PWD}"
+
+TAG=$(git describe --tags --dirty)
+RELEASE_DIR=release-${TAG}
+#Todo include  version with flag
+BUILDFLAGS="-ldflags '-extldflags -static -X main.version=${TAG}'"
```
dougbtv: Thanks for getting the `-X main.version=${TAG}` in there! At some point I'd love a way for a user to be able to pull that up too in case we're trouble shooting and need to figure out what version they have. This can be a follow on (I think there's a trellow card for it)

rkamudhan:

.travis.yml
rkamudhan: @s1061123  @dougbtv: yes, I understood the script now. I will make the changes now to include the `-ldflag` in the `goxc`. I think we can keep the `release.sh` for the user local use, and the travis for the github maintainer ? What you guys think ?

s1061123: @rkamudhan, I agree with you. How about to add some comments in release.sh to mention it (i.e. local use) for users and further developers?

rkamudhan: ok cool. I will add it now.

s1061123:
closed due to #206 fix.

PR #95, opened 2018-07-12
k8sclient: Network -> NetworkAttachmentDefinition

dcbw:
To align with latest spec change clarifying the name of the CRD object.

@s1061123 @dougbtv @rkamudhan

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,9 +89,9 @@ Multus is compatible to work with both CRD and TPR(deprecated in K8s 1.7).
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: network-attachment-definitions.kubernetes.cni.cncf.io
+  name: network-attachment-definitions.k8s.cni.cncf.io
 spec:
-  group: kubernetes.cni.cncf.io
+  group: k8s.cni.cncf.io
   version: v1
   scope: Namespaced
   names:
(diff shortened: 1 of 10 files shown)


rkamudhan:
@dcbw @dougbtv @s1061123 Just check this is major change in terms of compatibility of the Network CRD from v2.0. Is that we have to change the release to v3.0 ?

dougbtv:
Indeed it is a pretty big change, seeing you use a different format. Ummm... I'm OK with just making it 2.2, but... Also open to the idea of 3.0, as this does look bigger than a minor version change. 

Anyone else have input here?

dcbw:
> Do you mind running a quick sed and pushing it up? I think this should get the references to the annotation from that side of the equation.

@dougbtv so I thought about this originally, but decided that maybe we could keep "networks" for the selection annotation.  Here's my reason... the annotation is selecting the network to attach the pod to, which the NAD also refers to.  It's an implementation detail that we route that through the NAD to find the *real* network, but hey I'm grasping at straws here.  I also thought that since the annotations are things cluster users might set, perhaps we want to keep them shorter.  But all that said, consistency is also a good thing.

dougbtv:
Oh, hold on -- I didn't quite think of it this way. So we can have the official nomenclature with the CRD definition itself -- but we can name the selection side differently? That... seems ideal to me, somehow.

dcbw:
@dougbtv @rkamudhan updated the CRD path in k8sclient.go and testing code

rkamudhan:
@dcbw we can merge the PR after #87 PR merge. It is easier for the rebase..

rkamudhan:
@dcbw I think my conflist support is based on the old implementation. I think, I have to change my code flow now. I will test the PR, and merge it tomorrow. And add my changes on top of it..

dcbw:
> @dcbw we can merge the PR after #87 PR merge. It is easier for the rebase..

@rkamudhan sure

README.md
```diff
@@ -89,42 +89,41 @@ Multus is compatible to work with both CRD and TPR(deprecated in K8s 1.7).
...
-    singular: network
-    # kind is normally the CamelCased singular type. Your resource manifests use this.
-    kind: Network
-    # shortNames allow shorter string to match your resource on the CLI
+    plural: network-attachment-definitions
+    singular: network-attachment-definition
+    kind: NetworkAttachmentDefinition
     shortNames:
-    - net
+    - net-attach-def
```
rkamudhan: any other short name ? net ? or nad ?

rkamudhan: "net" will be easier as a user

examples/macvlan-conf.yml
```diff
@@ -1,5 +1,5 @@
-apiVersion: "kubernetes.cni.cncf.io/v1"
-kind: Network
+apiVersion: "k8s.cni.cncf.io/v1"
```
rkamudhan: I think we have to update `k8s.v1.cni.cncf.io/networks` in all  the pod examples.

rkamudhan:
+1 LGTM, Expect "k8s.v1.cni.cncf.io/networks" update in the demo file. I tested this PR and tested conf file, config spec support, both for thin and thick plugin(Code is inserting the field `name` correctly). So all +1 for this PR. It is ready to merge.

README.md
rkamudhan: @dcbw any suggestion here. `net-attach-def` is quite lengthy for shortnames

PR #96, opened 2018-07-12
[docs] Adds documentation for logging enhancements

dougbtv:
To be merged into 

I realize that only the API is available at the moment, and that the additional logging needs to be added in other parts of the code, however, I figured we could get a jump on having some docs available.

Tomo, please feel free to correct if there's somethings that I missed from the code, or otherwise had incorrect.

For example, I wrote that the `STDERR` logging is always performed, even though I see there's places where you could in theory turn it off -- but, I see that it isn't exposed to the user via the configuration file, so I wrote that it happens always.


rkamudhan:

PR #102, opened 2018-07-24
[minor] Updates kubernetes.v1 to k8s.v1 in the sample files

dougbtv:
This is really a minor documentation fix, but, in tandem with the recent changes to the CRD name change, and the change from `s/kubernetes.v1/k8s.v1/` in the namespace, I went ahead and made the changes to the sample files as well.


rkamudhan:
+1 LGTM

PR #114, opened 2018-07-31
Coveralls Code Coverage

dougbtv:
Adds `.travis.yml` updates to run coveralls, using [goveralls](https://github.com/mattn/goveralls).

Also adds CI badges to README.md

meta topic: Should we start a `develop` branch to merge these commit into for release together later?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
 
 before_install:
   - sudo apt-get update -qq
+  - go get github.com/mattn/goveralls
 
 install:
   # workaround golint install error in https://github.com/golang/lint/issues/288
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -32,3 +32,5 @@ import:
   - util/retry
 - package: github.com/vishvananda/netns
 - package: github.com/pkg/errors
+- package: github.com/mattn/goveralls
+  version: ^0.0.2
```
s1061123: Looking commit history, https://github.com/mattn/goveralls/commits/master, after v0.0.2, there are several commits.
In addition, v0.0.2 is released at Nov 2017, so roughly no release for 1year even though master is updated.

So I'm wondering we use master instead of release...
I suppose we just keep in mind for now and we use v0.0.2 first.

dougbtv: Ahh shoot, this is a mistake! This shouldn't be here. Thanks Tomo, it's unnecessary.

dougbtv:
Guys -- Just realized there's some mistakes here, the whole inclusion of goveralls in glide + adding the vendored libraries there was unnecessary. All that's really necessary is the binary for goveralls when we run in Travis! ...Thanks for finding that Tomo. I'll clean up the included libraries here.

.travis.yml
```diff
@@ -10,6 +10,7 @@ env:
 
 before_install:
   - sudo apt-get update -qq
+  - go get github.com/mattn/goveralls
```
rkamudhan:
```
   - mkdir -p $GOPATH/src/golang.org/x
   - pushd $GOPATH/src/golang.org/x	  
   - git clone https://github.com/golang/tools.git
```
no need to add line 20,  it is already imported in the line 13

Issue #115, opened 2018-08-01
Should not do any operation for multiple DEL call from kubelet.


rkamudhan:
For more information, refer: 

https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444

PR #120, opened 2018-08-02
Handling the multiple cmd del call from kubelet

rkamudhan:
Refer #115 for more information

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -299,16 +299,14 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 	}
 	netns, err := ns.GetNS(args.Netns)
 	if err != nil {
-		if err != nil {
-			//  if NetNs is passed down by the CEO, or if it called multiple times
-			// so don't return an error if the device is already removed.
-			// https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444
-			_, ok := err.(ns.NSPathNotExistErr)
-			if ok {
-				return nil
-			}
-			return fmt.Errorf("failed to open netns %q: %v", netns, err)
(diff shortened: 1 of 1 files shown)


rkamudhan:
@dcbw @s1061123 @dougbtv  Added fix with reference to https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444

multus/multus.go
```diff
@@ -294,6 +294,24 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 		return err
 	}
 
+	if args.Netns == "" {
+		return nil
+	}
+	netns, err := ns.GetNS(args.Netns)
+	if err != nil {
+		if err != nil {
+			//  if NetNs is passed down by the CEO, or if it called multiple times
```
s1061123: one q. what is the CEO?

multus/multus.go
```diff
@@ -294,6 +294,24 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 		return err
 	}
 
+	if args.Netns == "" {
+		return nil
+	}
+	netns, err := ns.GetNS(args.Netns)
+	if err != nil {
+		if err != nil {
```
s1061123: "if err != nil {" is duplicated and nested. How about to make it one?

multus/multus.go
rkamudhan: Sorry that should Cloud Orchestration Engine

multus/multus.go
rkamudhan: I don't know how it came it.. My bad

rkamudhan:
@s1061123 Sorry, I didn't get back to you on this. I will take your feedback and update the code today. I am target it for the [v3.1](https://github.com/intel/multus-cni/milestone/4)

PR #121, opened 2018-08-06
Add logging message for debug/error

s1061123:
This changes adds initial logging messages:debug and error. 

Error messages:
fmt.Errorf() is wrapped by logging.Errorf() hence all error message is also put to log file. 

Debug messages:
Added logging.Debugf() at the head of almost func, so we could track the code by logging message.


s1061123:
@rkamudhan and @dougbtv , 

Fixing conflicts for rebase/merge commits.

s1061123:
In https://github.com/intel/multus-cni/blob/master/logging/logging.go#L118, it create the file if there is not.

Is that enough?

rkamudhan:
Just run this PR.. It is working fantastically as expected.. LGTM..

PR #125, opened 2018-08-13
Enable hairpinMode in the delegate config

### LINKED ISSUE #124: Hairpin is not set in the delegate config
I have noticed `hairpin` is missing in [the multus config](https://github.com/intel/multus-cni/blob/master/images/70-multus.conf) and I think that setting should be added by default when `flannel` is used as default delegate.

So, the config should look like this:

```json
{
  "name": "multus-cni-network",
  "type": "multus",
  "delegates": [
    {
      "type": "flannel",
      "name": "flannel.1",
      "delegate": {
        "isDefaultGateway": true,
        "hairpinMode": true
      }
    }
  ],
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig"
}
```

mhristache:
Fix #124

### DIFF
--- a/images/70-multus.conf
+++ b/images/70-multus.conf
@@ -6,7 +6,8 @@
       "type": "flannel",
       "name": "flannel.1",
       "delegate": {
-        "isDefaultGateway": true
+        "isDefaultGateway": true,
+        "hairpinMode": true
       }
     }
   ],


rkamudhan:
@mhristache Thanks for identifying the issue

PR #160, opened 2018-10-03
Add the option to auto generate 00-multus.conf

Kusanagi9999:
When `--multus-conf-file=auto` is used, 00-multus.conf will be
automatically generated from the CNI configuration file of the master
plugin (the first file in lexicographical order in cni-conf-dir).


Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN yum install -y $INSTALL_PKGS && \
     yum clean all && \
     rm -rf /tmp/*
 
+# Install jq : command-line JSON processor
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 > jq && \
```
dougbtv: @Kusanagi9999 -- I think for the time being if we can remove the use of jq / including the jq binary here, we could be OK to get this merged into master. Granted, it'll make the resulting config a little less human legible, but... I think that'll be OK, it'll still parse. One of my colleagues mentioned that idea of making a go application to pretty print JSON and doing a multi-stage build, but... I think that might be overkill for now.

Thank you a bunch for the pull request -- cool feature!

Kusanagi9999: OK, sounds good. I removed jq for now.

images/entrypoint.sh
```diff
@@ -134,6 +142,35 @@ fi
...
+    exit 1;
+  elif [ "$MASTER_PLUGIN" == "00-multus.conf" ]; then
+    echo "Warning: Multus is already configured: auto configuration skipped."
+  else
+    MASTER_PLUGIN_JSON="$(cat $CNI_CONF_DIR/$MASTER_PLUGIN)"
+    CONF=$(cat <<-EOF
+			{
+				"name": "multus-cni-network",
+				"type": "multus",
+				"kubeconfig": "$MULTUS_KUBECONFIG",
```
Kusanagi9999: This will generate ```Err in loading K8s Delegates k8s args: GetK8sClient: failed to get context for the kubeconfig /host/etc/cni/net.d/multus.d/multus.kubeconfig``` it should be ```etc/cni/net.d/multus.d/multus.kubeconfig```

PR #162, opened 2018-10-05
Improve grep in entrypoint.sh to only find .conf and .conflist files

Kusanagi9999:


Kusanagi9999:
@dougbtv ^^

Kusanagi9999:
Hey @dougbtv - I am working on a PR to add Multus to Kubespray (https://github.com/kubernetes-incubator/kubespray/pull/3166). I'd like to use the image from https://hub.docker.com/r/nfvpe/multus/tags/. Do you think you could update ```latest``` once this has been merged? Also, is there an ETA for Multus v3.2?

dougbtv:
@Kusanagi9999 -- sounds awesome re: Kubespray! Let me think about it a bit -- I wanted to have `:latest` be the latest stable (so I don't have to update the daemonset yaml often), and the images/image tags gets updated when we tag in the git rpeo, however... I think maybe what we could do is update the `.travis.yml` to also build a `nfvpe/multus-latest` image. 

OR: I can manually tag an image for you and push it up to say, `nfvpe/multus:v3.1.autoconf` which includes this PR? Would that work in the meantime? 

No solid date on a 3.2, however, it shouldn't be too long. I would guess within 30 days, we have a number of features we'll want in and tagged shortly, I'm sure of it.

Also, merging here in a moment! Sorry so slow getting back to this PR, been on the road.

Kusanagi9999:
@dougbtv, OK got it, nfvpe/multus:v3.1.autoconf would totally work for now until v3.2 is out. Thanks a lot!

Issue #153, opened 2018-09-26
Load default network from file, not in multus.conf


s1061123:
In addition to #131, we may need to support loading default network from file.

s1061123:
Fixed #169.

Issue #131, opened 2018-08-23
Specify default network by name


squeed:
It would be nice to be able to specify the default network by name in the multus configuration file. This has a couple advantanges:

1. It allows for CNI configurations that are more dynamic
2. It works well with existing self-hosted plugins, e.g. Calico, that write their CNI configuration file via a daemonset
3. It also solves the readiness problem fixed in #98 in a slightly neater way.

It could take advantage of the already existing `getCNIConfigFromFile` function, so this should be pretty easy.

Thoughts? I can have a PR for this relatively quickly.

dougbtv:
Hey Casey, just trying to verify my understanding of this so...

Like today, we have the default network CNI configuration packed into say `/etc/cni/net.d/00-multus.conf` that looks like:

```
    {
      "name": "multus-cni-network",
      "type": "multus",
      "delegates": [
        {
          "type": "flannel",
          "name": "flannel.1",
          "delegate": {
            "isDefaultGateway": true
          }
        }
      ],
      "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig"
    }
```

Instead, we'd have a way to otherwise specify the default network with something like...

```
    {
      "name": "multus-cni-network",
      "type": "multus",
      "default": "calico",
      "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig"
    }
```

And then it'd pull up say, Calico either via CRD object or file on disk?

squeed:
@dougbtv yeah, that was my idea - thanks for elaborating it!

rkamudhan:
@squeed Thanks for the thoughts on it. @dougbtv PR is help us to check a plugin readiness, it need not to be default network. 

I see your idea is making sense to use the on-disk file as the default network. so delegate, gives a opportunity to attach multiple networks, with out any Network CRD objects itself. So I am thinking we can change the name `delegates` to `defaultnetworks` and make it to accept both network name or array of networks conf ? Is that ok for you guys ?

dougbtv:
I really like this idea, and I'm all for it. I also have uses for making the default plugin more flexible, so, I'm all for it for sure. Thanks @squeed

squeed:
cc @dcbw, who wanted to take a look.

s1061123:
Let's close it because #169 address the issue and now we can specify network-attachment name to specify defaultNetwork as well as clusterNetwork.

PR #169, opened 2018-10-16
Add clusterNetwork/defaultNetwork into multus.conf

### LINKED ISSUE #157: Specify cluster default network with a new CRD object.
In additional to #131 and #153 , I think it would be better to specify the cluster default network with a CRD object instead of config files. My idea is to create a new CRD object. It may looks like:

```
kind: MultusDefaultNetwork
metadata:
  name: default
spec: 
  networkAttachememntDefinition: default/calico
```

Multus can get the name of default network from this CRD object, then retrieve the CNI config from that networkAttachememntDefinition object. 

There are some advantages I can think of:

1. When administrator want to switch the default network from one to another, he doesn't need to update the file on each node.
2. It ensures consistency of the default network config across the cluster.
3. It can be accessed though kube API.

s1061123:
To support CRD/file/directory, add clusterNetwork/defaultNetwork in multus.conf file.
Verified that kubernetes: v1.12.1.

Fix #157, #153, #131


README.md
```diff
@@ -135,11 +135,54 @@ $ kubectl exec -it samplepod -- ip a
...
+    "type": "multus",
+    "kubeconfig": "/etc/kubernetes/node-kubeconfig.yaml",
+    "confDir": "/etc/cni/multus/net.d",
+    "cniDir": "/var/lib/cni/multus",
+    "binDir": "/opt/cni/bin",
+    "logFile": "/var/log/multus.log",
+    "logLevel": "debug",
+    /* NOTE: you can set clusterNetwork+defaultNetworks OR delegates!! (this is only for manual) */
+    "clusterNetwork": "defaultCRD",
+    "defaultNetwork": ["sidecarCRD", "flannel"],
```
s1061123: defaultNetworks, not defaultNetwork

multus/multus.go
```diff
@@ -247,7 +247,16 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
 		}
 	})
 
-	numK8sDelegates, kc, err := k8s.TryLoadK8sDelegates(k8sArgs, n, kubeClient)
+	if n.ClusterNetwork != "" {
+		err = k8s.GetDefaultNetworks(k8sArgs, n, kubeClient)
+		if err != nil {
+			return nil, logging.Errorf("XXX")
```
s1061123: need to replace 'XXX' with more good error message.

s1061123:
@dcbw @squeed could you please take a look into it?

PR #182, opened 2018-11-01
Make conflistDel() behave like conflistAdd()

### LINKED ISSUE #179: Chained meta plugin not found on Delete
Adding interface(s) with chained plugins (e.g. tuning) work fine.  When Deleting, the chained plugin cannot be found:

```
2018-10-30T14:40:33-04:00 [error] error in getting result from DelNetworkList: failed to find plugin "tuning" in path [/home/mcambria/go/src/github.com/containernetworking/bin]
2018-10-30T14:40:33-04:00 [error] Multus: error in invoke Conflist Del - "mcc-test1": error in getting result from DelNetworkList: failed to find plugin "tuning" in path [/home/mcambria/go/src/github.com/containernetworking/bin]
Multus: error in invoke Conflist Del - "mcc-test1": error in getting result from DelNetworkList: failed to find plugin "tuning" in path [/home/mcambria/go/src/github.com/containernetworking/bin]

```

It looks like logic was added to make this work for conflistAdd() by finding the plugin vis CNI_PATH:

https://github.com/intel/multus-cni/blame/bdf901c122f923d89fd88e15fb6dd79d8a98d4da/multus/multus.go#L126

Similar logic is missing for Delete.  Is there a reason?

When I add the same logic to conflistDel(), the problem goes away.   But does adding this break something else?

mccv1r0:
conflistAdd() finds binaries differently than conflistDel().
Make the two call find binaries the same way.

Fixes #179

Signed-off-by: Michael Cambria <mcambria@redhat.com>

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -140,11 +140,12 @@ func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string, e
 	return result, nil
 }
 
-func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string) error {
+func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string, exec invoke.Exec) error {
 	logging.Debugf("conflistDel: %v, %s, %s", rt, string(rawnetconflist), binDir)
 	// In part, adapted from K8s pkg/kubelet/dockershim/network/cni/cni.go
-	binDirs := []string{binDir}
-	cniNet := libcni.CNIConfig{Path: binDirs}
+	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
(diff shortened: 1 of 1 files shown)


multus/multus.go
```diff
@@ -140,11 +140,13 @@ func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string, e
 	return result, nil
 }
 
-func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string) error {
+//func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, binDir string) error {
```
s1061123: How about to remove it?

s1061123:
merged, thanks again!

PR #186, opened 2018-11-07
Add mutating admission webhook

przemeklal:
Mutating webhook:
* Read pod spec, look for network annotations - if present ask API server if there's any resource associated with the network
* Inject resources requests/limits into first container in the spec for all network annotations that use them
* In case of errors block creation of given pod
* Add code for automated deployment of required resources designed to run as a Job
* Add tests for various mutation functionalities

Add common mutating and validating webhook features:
* Build Docker image with webhook server and webhook installer applications
* Update documentation with installation steps
* Add Kubernetes job specification for automated creation of certificates, secret, service and webhook configs
* Add deployment specification for running webhook server


przemeklal:
rebased

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+func parsePodNetworkSelectionElement(selection string) (*types.NetworkSelectionElement, error) {
+	var namespace, name, netInterface string
+	var networkSelectionElement *types.NetworkSelectionElement
+
+	units := strings.Split(selection, "/")
+	switch len(units) {
+	case 1:
+		namespace = "default"
```
booxter: This violates CRD document. If namespace is not specified then the pods namespace, not the default namespace, should be assumed.

See here: https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit#

"4.1.1 Comma-delimited Format
This format is intended to be a very simple, user-friendly format, composed only of NetworkAttachmentDefinition object references separated by commas.  The format of each object reference is either (a) <NetworkAttachmentDefinition object name> to reference a NetworkAttachmentDefinition object in the pod’s namespace, or (b) <NetworkAttachmentDefinition object namespace>/<NetworkAttachmentDefinition object name> to reference a NetworkAttachmentDefinition object in a different namespace."

Dockerfile.webhook
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.11 as builder
```
booxter: Is it possible to make it compatible with docker versions that don't support build phases? (For one, this is the version that is shipped with centos 7.x.)

If interested, here is the patch that fixes the build for centos for me: https://github.com/booxter/multus-cni/commit/c5aed8d8a2ba58b69ad760e3c08de25f9859f22f

webhook/deployment/deployment.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	logging.Debugf("signed certificate succesfully obtained")
+
+	/* create secret and push it to the API */
+	err = createSecret(certificate, key)
+	if err != nil {
+		logging.Errorf("error creating secret: %s", err)
+		os.Exit(1)
+	}
+	logging.Debugf("secret succesfully created")
```
booxter: nit: successfully (two s), here and below

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+
+		/* unmarshal list of network selection objects */
+		networks, _ := parsePodNetworkSelections(netSelections)
+
+		for _, n := range networks {
+			/* for each network in annotation ask API server for network-attachment-definition */
+			networkAttachmentDefinition, err := getNetworkAttachmentDefinition(n.Namespace, n.Name)
+			if err != nil {
+				/* if doesn't exist: deny pod */
+				explanation := logging.Errorf("could not find network attachment definition '%s/%s': %s", n.Name, n.Namespace, err)
```
booxter: I think the order of name and namespace in the error message should be swapped to reflect how we encode it in networks annotation (namespace/networkName).

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+				err = prepareAdmissionReviewResponse(false, explanation.Error(), ar)
+				if err != nil {
+					logging.Errorf("error preparing AdmissionReview response: %s", err)
+					http.Error(w, err.Error(), http.StatusBadRequest)
+					return
+				}
+				writeResponse(w, ar)
+				return
+			}
+			logging.Debugf("network attachment definition '%s/%s' found", n.Namespace, n.Name)
```
booxter: Same comment about order of arguments.

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+				return nil, logging.Errorf("error parsing network selection element: %v", err)
+			}
+			networkSelections = append(networkSelections, networkSelectionElement)
+		}
+	}
+
+	/* fill missing namespaces with default value */
+	for _, networkSelection := range networkSelections {
+		if networkSelection.Namespace == "" {
+			networkSelection.Namespace = "default"
```
booxter: It should be the pod's namespace not necessarily "default".

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		logging.Errorf(err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	writeResponse(w, ar)
+}
+
+func getNetworkAttachmentDefinition(namespace, name string) (*types.NetworkAttachmentDefinition, error) {
+	path := fmt.Sprintf("/apis/k8s.cni.cncf.io/v1/namespaces/%s/network-attachment-definitions/%s", namespace, name)
```
booxter: At some point, we should adopt the official network CRD client to fetch objects: https://github.com/K8sNetworkPlumbingWG/network-attachment-definition-client recently adopted by the networking WG (though it's not exactly ready as of now, there are 4 open PRs to merge before adopting the library).

webhook/server/server.go
```diff
@@ -0,0 +1,437 @@
...
+	rawNetworkAttachmentDefinition, err := clientset.ExtensionsV1beta1().RESTClient().Get().AbsPath(path).DoRaw()
+	if err != nil {
+		err := logging.Errorf("could not get Network Attachment Defintion %s/%s: %s", namespace, name, err)
```
booxter: Defintion -> Definition; also order of arguments should be name, then namespace to reflect actual notation of networks annotation.

booxter:
> This is still not 100% completed as there are some missing tests for automated deployment code and I'm not fully happy with all of this code yet, but I'd like to get some feedback on that work, whether it's going into the right direction or not.
> 
> I was also discussing with @ahalim-intel an option to validate that there's at least one node in the cluster that satisfies network resources request (which is done later by Kubernetes scheduler anyway), but we weren't really sure if that should be responsibility of the webhook, so it would be great to get some feedback on that as well.

I think it's not the job for the hook (same as it's not the job for the hook to validate that resources are allocatable / available anywhere in the cluster.)

przemeklal:
@booxter thanks for your review, I agree on all comments, so I'm starting to work on the changes right away

I'm also going to use your Dockerfile patch as part of this PR, so you won't have to create a new one for it in the future

Dockerfile.webhook
przemeklal: done

webhook/server/server.go
przemeklal: I'll be happy to use it once ready

webhook/server/server.go
przemeklal: typo fixed, in regard to ordering it should be "namespace/name", same as above right? if so, it's been already correctly formatted

webhook/server/server.go
przemeklal: done

webhook/server/server.go
przemeklal: done

webhook/server/server.go
przemeklal: done

webhook/deployment/deployment.go
przemeklal: fixed

webhook/server/server.go
przemeklal: fixed

webhook/server/server.go
booxter: You are right about order, sorry. ;)

przemeklal: no problem ;)

booxter:
@przemeklal thanks for the latest upload. Is the PR perhaps now ready to merge? I don't see major issues here in the production code itself (though I know you planned to expand test coverage here, so maybe this is it).

przemeklal:
thanks @booxter I think it would be OK to merge that at this point, coveralls check passes, so I'm good with that
@rkamudhan @dougbtv what do you guys think?

I also don't think that tests for code deploying various Kubernetes resources are the highest priority at the moment, so we could address that later if everyone agrees

booxter:
@przemeklal what's the plan for the PR in light of https://github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller created to keep (some of) this code? Are we still on track to merge this PR here, or are you going to scrap this PR (or validating part of it) and contribute it to the WG repo?

przemeklal:
Hi @booxter, plan is to split this PR into 2 separate applications, validating part will go under the repository you linked and for the mutating webhook there will be a separate repository, not yet created.

booxter:
@przemeklal great. Do we have an idea on when we may expect the split to happen? Is there anything I can help with in this regard? Please keep me posted about changes in plans related to the webhook location etc. And thanks a lot for your work.

booxter:
I know this PR is not going to be merged, but to leave another piece of feedback I have for the code: I think we should use kube-system namespace for all webhook components. Currently, manifests use the default namespace, which means that the webhook services are showing among other regular pods.

webhook/server/server.go
```diff
@@ -327,14 +327,14 @@ func mutateHandler(w http.ResponseWriter, req *http.Request) {
 		resourceRequests := make(map[string]int64)
 
 		/* unmarshal list of network selection objects */
-		networks, _ := parsePodNetworkSelections(netSelections)
+		networks, _ := parsePodNetworkSelections(netSelections, pod.ObjectMeta.Namespace)
```
booxter: I think there may be an issue here when pod doesn't have namespace explicitly set in the manifest. If that's the case, .Namespace will be an empty string here, which then results in failure to fetch network attachment.

Here is a snippet from official API description for namespace field:

```
Namespace defines the space within each name must be unique. An empty namespace is equivalent to the "default" namespace, but "default" is the canonical representation. Not all objects are required to be scoped to a namespace - the value of this field for those objects will be empty. Must be a DNS_LABEL. Cannot be updated. More info: http://kubernetes.io/docs/user-guide/namespaces
```

I think we should fall back to "default" if final namespace is still empty string after all substitutes.

booxter: And of course, this suggests we need a unit test case to cover for an empty namespace in manifest.

dougbtv:
Updated plan, just so everyone is aware, we'll take this pull request and move all of the code over to the repo in the NPWG github, and the next step from there will be to make it specific to the network attachment definition spec.

booxter:
@dougbtv thanks for the update. What does it mean to "make it specific to the network attachment definition spec"? Is there an ETA for this move to a new repo to happen? Again, thanks for the update.

booxter:
@dougbtv thanks for the update. What does it mean to "make it specific to the network attachment definition spec"? Is there an ETA for this move to a new repo to happen? Again, thanks for the update.

przemeklal:
New PR with mutating admission webhook for network resources injection has been opened here https://github.com/intel/sriov-network-device-plugin/pull/55

All comments are welcome as always.

@booxter this one includes your two latest comments - kube-system is the default namespace for deployment now and I added a fix that should solve the issue with missing default namespace in a pod spec, although not sure if that's enough as I never managed to reproduce it.

booxter:
Thanks for the update. I am on vacation and will be able to check the new code in 2019 only. Please don't hesitate to move forward without me if it's time sensitive to merge this now.

s1061123:
Close it because the code goes to NPWG repo.

Issue #139, opened 2018-08-29
add more description for whole multus.conf to clarify its requirements


soyuutaku:
@s1061123 add:
This issue is to address that current README.md improvement so that reader don't misunderstand that minimum requirements for multus.conf, especially in [https://github.com/intel/multus-cni#configuring-multus-to-use-the-kubeconfig](https://github.com/intel/multus-cni#configuring-multus-to-use-the-kubeconfig)
------------------------------ Original description --------------------
I'm using master branch to deploy multus. When i set the config file with:

```
{
    "name": "node-cni-network",
    "type": "multus",
    "kubeconfig": "/etc/kubernetes/node-kubeconfig.yaml"
}
```

I cannot start a pod with network annotation.

From logs, it said that `at least one delegate must be specified`. 

Do I have to set a default delegate to use multus?

s1061123:
Right. multus itself does not provide any network interface to container.
Pod needs network interface at least one interface and that is the default delegate.

We also plan to set default network as network attachment definition in future, but it is not yet, so 
could you please add default delegate into your config?

soyuutaku:
@s1061123, I got your point, that is: Currently we must set a default delegate, no matter pod have annotation of network or not.

What i am trying to do is use annotations to ensure every pod have one or more interface. Further more, I may need some pods have master network with flannel, and some other pods have master network with macvlan. Is that possible?

s1061123:
@soyuutaku I am not clear that the difference between default delegate and master network in your comment. I feel master network and default delegate points same network and this master network is connected to kubernetes API network, too.

Currently multus provides to add one ore more 'extra networks' with one default delegates to the pod.
Kubernetes needs to access the pod and default delegates is used for that, so we cannot remove default delegates (= master network) and Kubernetes provides only one network for that, hence we cannot do that you mentioned (provide two or more master network in a cluster).

soyuutaku:
@s1061123.  My understanding is follow steps below, I can have two pods with only one network interface, and the interfaces are in different networks. Is that right?

1. let the config file as below, and than make sure every pod have annotations of network name by crd.

```
{
    "name": "node-cni-network",
    "type": "multus",
    "kubeconfig": "/etc/kubernetes/node-kubeconfig.yaml"
}
```

2. start up a pod as:
```
apiVersion: v1
kind: Pod
metadata:
  name: samplepod1
  annotations:
    k8s.v1.cni.cncf.io/networks: macvlan
spec:
  containers:
...

```

[...]

PR #191, opened 2018-11-13
Update README.md and split into several child documents

### LINKED ISSUE #154: Break up the README to include a developer guide
Include in that guide, the suggestion that you maintain your own fork and create PRs from that.

s1061123:
Fix #154 and #139. Thank you @dougbtv for reviewing the docs!


doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
+## How to use multus-cni?
+
+### Prerequisite
+
+Kubelet must be configured to run with the CNI network plugin. Edit `/etc/kubernetes/kubelet` file and add `--network-plugin=cni` flags in `KUBELET\_OPTS `as shown below:
```
Billy99: On my system, the file `/etc/kubernetes/kubelet` does not exist. If this is the correct file, then add a comment that file may need to be created.

After I ran the 'cat ./images/{multus-daemonset.yml,flannel-daemonset.yml} | kubectl apply -f -' command, 'echo $KUBELET_OPTS' indicated the environmental variable wasn't set, so not sure it if took.

Billy99: `KUBELET\_OPTS ` should be `KUBELET_OPTS`

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+
+```
+KUBELET_OPTS="...
+--network-plugin-dir=/etc/cni/net.d
+--network-plugin=cni
+"
+```
+
+### Install multus
```
Billy99: This section has two bullets, "copy multus binary" and "use daemonset". I think it would be clearer if the text associated with each bullet were indented. Not sure if that is possible, the associated sentence may need an indented bullet '   -', but your call. Just a suggestion.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+
+- use daemonset
+
+As [Quickstart](https://github.com/intel/multus-cni/README.md#quickstart-guide), you could apply as such:
+
+```
+# Execute following command at Kubernetes master
+$ cat ./images/{multus-daemonset.yml,flannel-daemonset.yml} | kubectl apply -f -
+```
+
```
Billy99: General comment, it is hard to tell on the remaining sections which have already been done by the Quickstart "use daemonset" command (which I used).

For example:
* 70-multus.conf is created for you, just needs to be edited.
* I tried the "Create ServiceAccount, ClusterRole and its binding" step and got errors:
Error from server (AlreadyExists): error when creating "STDIN": serviceaccounts "multus" already exists
Error from server (AlreadyExists): error when creating "STDIN": clusterroles.rbac.authorization.k8s.io "multus" already exists
Error from server (AlreadyExists): error when creating "STDIN": clusterrolebindings.rbac.authorization.k8s.io "multus" already exists.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+For the detail, please take a look into [Configuration Reference](configuration.md)
+
+**You can use "clusterNetwork"/"defaultNetworks" instead of "delegates", see []() for the detail**
+
+As above config, you need to set `"kubeconfig"` in the config file for NetworkAttachmentDefinition(CRD).
+
+##### Which network will be used for "Pod IP"?
+
+In case of "delegates", the first delegates network will be used for "Pod IP". Otherwise, "clusterNetwork" will be used for "Pod IP".
```
Billy99: With regards to `In case of "delegates", the first delegates network will be used for "Pod IP".`, did this change? In the userspace CNI multus example, looks like they put flannel last. If it did change, may want to highlight or bold. If not, userspace example got it wrong and is not such a big deal.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+
+#### Create ServiceAccount, ClusterRole and its binding
```
Billy99: Looks like `Quickstart` covered this. Is there was way to display settings to determine what has been set?

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+  kind: ClusterRole
+  name: multus
+subjects:
+- kind: ServiceAccount
+  name: multus
+  namespace: kube-system
+EOF
+```
+
+#### Set up kubeconfig file
```
Billy99: Looks like `Quickstart` covered this. I see the /etc/cni/net.d/multus.d/multus.kubeconfig on master and on both my nodes.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+
+As above config, you need to set `"kubeconfig"` in the config file for NetworkAttachmentDefinition(CRD).
+
+##### Which network will be used for "Pod IP"?
+
+In case of "delegates", the first delegates network will be used for "Pod IP". Otherwise, "clusterNetwork" will be used for "Pod IP".
+
+#### Create ServiceAccount, ClusterRole and its binding
+
+Create resources for multus to access CRD objects as folloiwng command:
```
Billy99: folloiwng -> following

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+EOF
+```
+
+For the detail, please take a look into [Configuration Reference](configuration.md)
+
+**You can use "clusterNetwork"/"defaultNetworks" instead of "delegates", see []() for the detail**
+
+As above config, you need to set `"kubeconfig"` in the config file for NetworkAttachmentDefinition(CRD).
+
+##### Which network will be used for "Pod IP"?
```
Billy99: When this file is rendered, the bold statement `**You can use "clusterNetwork"/"defaultNetworks" instead of "delegates", see []() for the detail**` and section `##### Which network will be used for "Pod IP"?` look like different sections and got lost for a second. Maybe add "NOTE:" in front the bold line, or turn both into some bulleted notes about the behaviour.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            config:
+                 type: string
+EOF
+```
+
+### Create network attachment definition
```
Billy99: Maybe add an introduction paragraph such as:

The 'NetworkAttachmentDefinition' is used to setup the non-Pod IP network, ... There are two ways to configure the 'NetworkAttachmentDefinition':
* NetworkAttachmentDefinition with json CNI config
* NetworkAttachmentDefinition with CNI config file

doc/how-to-use.md
s1061123: Changed to indicate K8s document for CNI.

doc/how-to-use.md
s1061123: Add indent.

doc/how-to-use.md
s1061123: Could you please give me a pointer fo userspace CNI multus example file?

doc/how-to-use.md
s1061123: Add "(daemonset automatically does)" at the section.

doc/how-to-use.md
Billy99: https://github.com/intel/userspace-cni-network-plugin#multus-configuration-details

doc/how-to-use.md
s1061123: This is not required when you use daemonset. As I taken care of previous comment, this should be skipped for quickstart.

doc/how-to-use.md
s1061123: Thank you for your comment! I agree indeed. Add "NOTE:".

doc/how-to-use.md
s1061123: I've changed.

doc/how-to-use.md
```diff
@@ -0,0 +1,400 @@
...
+                    } ]
+                ]
+            }
+        }'
+EOF
+```
+
+#### NetworkAttachmentDefinition with CNI config file:
+
+If NetworkAttachmentDefinition has no spec, multus find a file in defaultConfDir ('/etc/cni/multus/net.d', with same name in the 'name' field of CNI config.
```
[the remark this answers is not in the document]
s1061123: As of 'type/conf.go', `/etc/cni/multus/net.d` is correct. '/etc/cni/net.d/multus' is for kubeconfig file in case of daemonset...

doc/how-to-use.md
s1061123: https://github.com/intel/multus-cni/blob/master/multus/multus.go#L270-L271

s1061123: As mentioned above, multus uses first delegates as master, so it could be valid.

doc/how-to-use.md
s1061123: Removed 'KUBELET_OPTS'.

doc/quickstart.md
```diff
@@ -0,0 +1,174 @@
...
+
+We'll first install Multus CNI, and then we'll setup some configurations so that you can see how multiple interfaces are created for pods.
+
+## Key Concepts
+
+Two things we'll refer to a number of times through this document are:
+
+* "Default network" -- This is your pod-to-pod network. This is how pods communicate among one another in your cluster, how they have connectivity. Generally speaking, this is presented as the interface named `eth0`. This interface is always attached to your pods, so that they can have connectivity among themselves. We'll add interfaces in addition to this.
+* "CRDs" -- Custom Resource Definitions. Custom Resources are a way that the Kubernetes API is extended. We use these here to store some information that Multus can ready. Primarily, we use these to store the configurations for each of the additional interfaces that are attached to your pods.
```
Billy99: Multus can ready. -> Multus can read.

doc/quickstart.md
```diff
@@ -0,0 +1,174 @@
...
+  "cniVersion": "0.3.0",
+  "type": "loopback",
+  "additional": "information"
+}
+```
+
+CNI configurations are JSON, and we have a structure here that has a few things we're interested in:
+
+1. `cniVersion`: Tells each CNI plugin which version is being used and can give the plugin information if it's using a too late (or too early) version.
+2. `type`: This tells CNI which binary to call on disk. Each CNI plugin is a binary that's called. Typically, these binaries are stored in `/opt/cni/bin` on each node, and CNI executes this binary. In this case we've specified the `loopback` binary (which create a loopback-type network interface).
```
Billy99: Just notes that might be helpful to add, but don't feel you have to:
* Does my CNI need to be in place (/opt/cni/bin/) before the quick start is run for it to be propagated to each node?
* Along the same lines, if I'm debugging and update the CNI, do I need to manually propagate it to each node?

dougbtv: Cool, made a note to check the contents of the directory to make sure your binaries are on disk. And then also made a note about how the configs are read on pod creation & deletion.

PR #195, opened 2018-11-25
Support IPRequest to specify IP address for interface

AlonaKaplan:
Currently the IP is read from the pod network annotation into the NetworkSelectionElement.IPRequest but is not passed further to DelegateNetConf.IPRequest and then to the CNI_ARGS.
This PR fixes it.
The PR also includes some test fixes for both MacRequest and IPRequest (even the MacRequest was already passed correctly it seems the multus_test didn't actually tested it).

Note: the PR assumes the format of the IP on the network annotation is ips: "ip_address" as it was previously assumed by the code read the IP from the pod network annotation into NetworkSelectionElement.IPRequest.
Although it doesn't comply with the de-facto standard specifies the format is ips: ["ip1", "ip2", ...]

### DIFF
--- a/multus/multus_test.go
+++ b/multus/multus_test.go
@@ -44,12 +44,11 @@ func TestMultus(t *testing.T) {
 }
 
 type fakePlugin struct {
-	expectedEnv     []string
-	expectedConf    string
-	expectedIfname  string
-	expectedMacAddr string
-	result          cnitypes.Result
-	err             error
+	expectedEnv    []string
+	expectedConf   string
+	expectedIfname string
+	result         cnitypes.Result
+	err            error
 }
 
 type fakeExec struct {
@@ -60,7 +59,7 @@ type fakeExec struct {
 	plugins  []*fakePlugin
 }
 
-func (f *fakeExec) addPlugin(expectedEnv []string, expectedIfname, expectedMacAddr, expectedConf string, result *types020.Result, err error) {
+func (f *fakeExec) addPlugin(expectedEnv []string, expectedIfname, expectedConf string, result *types020.Result, err error) {
 	f.plugins = append(f.plugins, &fakePlugin{
 		expectedEnv:    expectedEnv,
 		expectedConf:   expectedConf,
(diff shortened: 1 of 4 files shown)


dcbw:
LGTM.  Any chance you'd want to do a follow-up to fix up the multiple IP requests issue?

AlonaKaplan:
Sure, we do. I just don't know how multiple ips should be passed over the CNI_ARGS.

Another issue is where does the subnet should be specified? I couldn't find it in the standard. I would expect the ip in the ips list be in the format ip/prefix.

s1061123:
In travis log, `golint ./multus/... | grep -v ALL_CAPS` has some error. 
https://travis-ci.org/intel/multus-cni/builds/459436938#L484

Could you please take care of it?

s1061123:
@AlonaKaplan Thank you for your PR! 
In case of IPv4/v6 dual stack case, user may need to specify IPv4 and IPv6 with CNI_ARGS hence we may need multiple ip address. That's the reason.

multus/multus_test.go
```diff
@@ -48,6 +48,7 @@ type fakePlugin struct {
 	expectedConf    string
 	expectedIfname  string
 	expectedMacAddr string
+	expectedIpAddr  string
```
s1061123: This is golint error in travisCI, so could you please change it to expectedIPAddr?
```
$ golint ./multus/... | grep -v ALL_CAPS
multus/multus_test.go:51:2: struct field expectedIpAddr should be expectedIPAddr
```

AlonaKaplan:
> @AlonaKaplan Thank you for your PR!
> In case of IPv4/v6 dual stack case, user may need to specify IPv4 and IPv6 with CNI_ARGS hence we may need multiple ip address. That's the reason.

Thanks for the explanation. How should it be reflected in the CNI_ARGS? Currently, as far as I know the de-facto standard specifies the IP should be passed over the IP key in the CNI_ARGS. If I have both ipv4 and ipv6 how would it look like in the CNI_ARGS?

multus/multus_test.go
AlonaKaplan: Thanks, done.

dcbw:
> Another issue is where does the subnet should be specified? I couldn't find it in the standard. I would expect the ip in the ips list be in the format ip/prefix.

@AlonaKaplan Interesting point. But every network attachment that wants L3/IP must already have IPAM configured somewhere that is driven by a plugin like host-local, static, or something internal to the plugin. Otherwise there is nothing to handle the IP CNI_ARG that is passed to the plugin. And the thing that handles IPAM therefore already knows what the subnet is, and if the CNI_ARGS static IP does not fall within the range, then it is invalid and should be rejected. If it does fall within the range, then we already know the range and the user doesn't need to specify redundant information.

But that falls down somewhat for a case like Tomo's 'static' IPAM CNI plugin, which doesn't have configuration to accept a subnet via JSON config and then validate an incoming CNI_ARGS static IP against that subnet. Instead it takes the CNI_ARGS static IP as CIDR notation.

So, a question for the NPWG I think, whether we should clarify language around section `4.1.2.1.3 "ips"` to allow CIDR notation and say that it should be validated by the IPAM function of the plugin in any way the plugin desires.

s1061123:
@dcbw I agree with you that Spec allows to both, CIDR-notation and non-CIDR. From multus/meta-plugin point of view, these parameter goes to IPAM plugin for now and IPAM should take care of that.

Sometimes I feel that CNI community could come up with standard IPAM argument formats to interoperability point of view... what do you think about it?

AlonaKaplan:
@dcbw @s1061123 please re-review, I've added support for CIDR format.  Also found (and fixed) a bug that the macs and ips of all the networks were appended to the CNI_ARGS.

dcbw:
LGTM

Issue #202, opened 2018-11-30
defaultNetwork should be skipped in "kube-system" Pod


s1061123:
**What happend**:
"kube-system" Pod is trying to have defaultNetwork network attachment.

**What you expected to happen**:
"kube-system" should not have it because "kube-system" is system pod so it expects single interface at present.

**How to reproduce it (as minimally and precisely as possible)**:
Setup k8s with multus+defaultNetworks.

**Anything else we need to know?**:

**Environment**:

- Multus version 
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`):
- Primary CNI for Kubernetes cluster:
- OS (e.g. from /etc/os-release):
- File of '/etc/cni/net.d/'
- File of '/etc/cni/multus/net.d'
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)
- Target pod yaml info (with annotation, use `kubectl get pod <podname> -o yaml`)
- Other log outputs (if you use multus logging)

s1061123:
#203 fixes it.

PR #203, opened 2018-11-30
clusterNetwork/defaultNetworks and namespace spec fixed

s1061123:
This fix is to add declaration about clusterNetwork/defaultNetwork
net-attach-def is in 'default' namesspace. In addition, this code
changes to skip defaultNetwork in case of 'kube-system' namespace
as well (#202).


doc/configuration.md
```diff
@@ -53,7 +53,7 @@ User should chose following parameters combination (`clusterNetwork`+`defaultNet
 
 Multus will find network for clusterNetwork/defaultNetworks as following sequences:
 
-1. CRD object for given network name
+1. CRD object for given network name, in 'default' namespace
```
SchSeba: @s1061123 Can you please explain this the I cant understand the combination between  clusterNetwork and the default namespace.

s1061123: Do you mean that the difference between 'clusterNetwork' and 'defaultNetwork'?

clusterNetwork: this network is used for Kubernetes cluster. This network is also used by Kubernetes (e.g. Pod IP).
defaultNetwork: this network is automatically added to all Pods (other than kube-system pod).

Both value takes CRD network-attachment-definition as you mentioned. Previously we don't mention that which namespace the network-attachment-definition is belonging to, hence I explicitly added.

Is that clear?

k8sclient/k8sclient.go
```diff
@@ -621,13 +621,15 @@ func GetDefaultNetworks(k8sArgs *types.K8sArgs, conf *types.NetConf, kubeClient
 	delegate.MasterPlugin = true
 	delegates = append(delegates, delegate)
 
-	//need to revisit
-	for _, netname := range conf.DefaultNetworks {
-		delegate, err := getNetDelegate(kubeClient, netname, conf.ConfDir)
-		if err != nil {
-			return err
+	// Pod in kube-system namespace does not have default network for now.
+	if string(k8sArgs.K8S_POD_NAMESPACE) != "kube-system" {
```
bmeng: Do we want to make the special namespace list configurable instead of hardcode?

s1061123: It might be added in future feature, but for now we don't have use-case hence not a plan yet.

s1061123: Do you have any use-case for that?

bmeng: After reconsideration, I am even not sure why the 'kube-system' should be excluded from the 'defaultNetworks'

I was thinking that there are also some other namespaces may have pods running which does not need the secondary interface. But actually, both the kube-system and the namespace which contains the system services (eg, openshift-sdn) will have pod running with `hostNetwork: true`, which will not invoke the multus-cni already. 

AFAIK, we do not have any pod running in kube-system with container network, so do we really need this condition?

doc/configuration.md
SchSeba: I will prefer if this that the default will be the same namespace where the multus daemonset was deployed.

for security reasons we don't know who have access to the default namespace. a regular use can change this crd and brake all the cluster.

the default namespace for multus daemonset now is the kube-system namespace, a namespace that control all the main components of the k8s cluster. I think is the best place to put the clusterNetwork CRD. (hard coded for now works for me)

k8sclient/k8sclient.go
SchSeba: The codeDNS will be up only after you apply a cni like calico or flannel before that it will be in pending state.

Can we give the multus `hostNetwork: true` and allow him to run if there is no cni install.
```yaml
      tolerations:
      - operator: Exists
        effect: NoSchedule
```
Then you can apply multus and the clusterNetwork CRD before the real cni.

doc/configuration.md
s1061123: I agree that we may put defaultNetwork/clusterNetwork in 'kube-system'. So let me change doc/code.

BTW, multus-cni plugin itself is invoked by kubelet from conatiner host, as far as I know, we could not identify multus-daemonset's namepsace and multus-daemonset itself.

SchSeba: We can this to the daemonset
```yaml
      env:
        - name: MY_POD_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
```

k8sclient/k8sclient.go
s1061123: I'd like to keep to install multus without daemonset so I'd like to keep.

s1061123:
@SchSeba yeah, I'm looking into the issue.

s1061123:
CI works fine.

PR #205, opened 2018-12-05
[dockerfile] Adds Dockerfile.rhel for OpenShift build

dougbtv:
Creates a new Dockerfile specific to OpenShift builds


dougbtv:
Great suggestion! I went ahead and added the comment, also a comment to the original Dockerfile (and removed the OpenShift specific labels from it). I'll merge after we have a passing Travis build. Thanks!

PR #210, opened 2018-12-11
Pod network extra args

AlonaKaplan:
Support extra keys in pod network annotation

Acorrding to the de-facto standard documentation-
"All key names that do not include a period character are reserved to ensure this specification
may be extended in the future. Implementations that write keys other than those defined in this
specification must use reverse domain name notation (eg "​ org.foo.bar.key-name​ ") to
name the non-standard keys."

This PR extends the NetworkSelectionElement with a key-value map (string->string)
containg the extra, non-stadnard keys passes in the pod network annotation.
(Note - only keys wih a period character are considered as extra keys).

The extra keys and their values are passed to the relevant cni plugin via the CNI_ARGS
environment variable using the following format -
$CNI_ARGS=$CNI_ARGS;key.1=value1;key.2=value2;

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -172,7 +172,7 @@ func delegateAdd(exec invoke.Exec, ifName string, delegate *types.DelegateNetCon
 		return nil, logging.Errorf("cannot set %q ifname to %q: %v", delegate.Conf.Type, ifName, err)
 	}
 
-	if delegate.MacRequest != "" || delegate.IPRequest != "" {
+	if delegate.MacRequest != "" || delegate.IPRequest != "" || delegate.ExtraFields != nil {
 		if cniArgs != "" {
 			cniArgs = fmt.Sprintf("%s;IgnoreUnknown=true", cniArgs)
 		} else {
@@ -203,8 +203,23 @@ func delegateAdd(exec invoke.Exec, ifName string, delegate *types.DelegateNetCon
 			cniArgs = fmt.Sprintf("%s;IP=%s", cniArgs, delegate.IPRequest)
 			logging.Debugf("Set IP address %q to %q", delegate.IPRequest, ifName)
 		}
+
+		if delegate.ExtraFields != nil {
+			invalidChars := "; | ="
+			for key, value := range delegate.ExtraFields {
+				if strings.ContainsAny(key, invalidChars) {
+					return nil, logging.Errorf("Key %q contains %q character/s which is invalid", key, invalidChars)
+				}
+				if strings.ContainsAny(value, invalidChars) {
+					return nil, logging.Errorf("Value %q contains %q character/s which is invalid", value, invalidChars)
+				}
+				cniArgs = fmt.Sprintf("%s;%s=%s", cniArgs, key, value)
+				logging.Debugf("Set %q %q to %q", key, value, ifName)
+			}
+		}
+
 		if os.Setenv("CNI_ARGS", cniArgs) != nil {
-			return nil, logging.Errorf("cannot set %q mac to %q and ip to %q", delegate.Conf.Type, delegate.MacRequest, delegate.IPRequest)
+			return nil, logging.Errorf("cannot set %q CNI_ARGS", delegate.Conf.Type)
 		}
 	}
 
--- a/types/types.go
+++ b/types/types.go
@@ -74,11 +74,12 @@ type DelegateNetConf struct {
 	ConfList      types.NetConfList
 	IfnameRequest string `json:"ifnameRequest,omitempty"`
 	MacRequest    string `json:"macRequest,omitempty"`
-	IPRequest string `json:"ipRequest,omitempty"`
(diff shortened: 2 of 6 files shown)


types/types.go
```diff
@@ -127,6 +128,7 @@ type NetworkSelectionElement struct {
 	// InterfaceRequest contains an optional requested name for the
 	// network interface this attachment will create in the container
 	InterfaceRequest string `json:"interface,omitempty"`
+	ExtraFields      map[string]string
```
SchSeba: Can you please a comment here to explain this field?

multus/multus.go
```diff
@@ -202,9 +202,23 @@ func delegateAdd(exec invoke.Exec, ifName string, delegate *types.DelegateNetCon
...
+
+			if delegate.ExtraFields != nil {
+				for key, value := range delegate.ExtraFields {
+					if strings.Contains(key, ";") {
+						return nil, logging.Errorf("Key %q contains ';' character which is invalid", key)
+					}
+					if strings.Contains(value, ";") {
+						return nil, logging.Errorf("Value %q contains ';' character which is invalid", value)
+					}
+					cniArgs = fmt.Sprintf("%s;%s=%s", cniArgs, key, value)
```
SchSeba: I think you need also to check if the key or value contains `=` this can also break the code.

multus/multus.go
```diff
@@ -202,9 +202,23 @@ func delegateAdd(exec invoke.Exec, ifName string, delegate *types.DelegateNetCon
 
 			cniArgs = fmt.Sprintf("%s;IP=%s", cniArgs, delegate.IPRequest)
 			logging.Debugf("Set IP address %q to %q", delegate.IPRequest, ifName)
+
+			if delegate.ExtraFields != nil {
+				for key, value := range delegate.ExtraFields {
+					if strings.Contains(key, ";") {
```
SchSeba: Can you add a test for this validation?

multus/multus.go
```diff
@@ -202,9 +202,23 @@ func delegateAdd(exec invoke.Exec, ifName string, delegate *types.DelegateNetCon
...
+						return nil, logging.Errorf("Key %q contains ';' character which is invalid", key)
+					}
+					if strings.Contains(value, ";") {
```
SchSeba: Can you add a test for this validation?

SchSeba:
I change the extra name on the test.
related to https://github.com/containernetworking/cni/issues/608

SchSeba:
Hi @dcbw can you please take a look on this PR?

dcbw:
@SchSeba @AlonaKaplan the spec doesn't say anything about restrictions on the content of the extra args fields. But obviously you can't pass that through environment variables like CNI_ARGS, so this PR doesn't allow things that the spec does allow.

Also, CNI_ARGS is deprecated upstream. You could fix both issues by using the CNI 'args' key in the JSON that gets passed to each plugin via stdin, which is the replacement for CNI_ARGS. That has no restrictions on content because JSON allows values to be escaped.

SchSeba:
@dcbw sure I can work on that just to be sure how the args key should look

```golang
type DelegateNetConf struct {
	Conf          types.NetConf
	ConfList      types.NetConfList
	...
	// Raw JSON
	Bytes []byte
        Args map[string]string
}
```

something like that?

SchSeba:
@dcbw ping?

rkamudhan:
@dcbw Some thing like this ?

```
type NetArgs struct {
	types.CommonArgs
}
```

so the CNI looks like
```
{
  "cniVersion": "0.4.0",
  "name": "pci",
  "type": "ovs",
  // type (plugin) specific
  "bridge": "ovs0",
  "vxlanID": 42,
  "ipam": {
    "type": "dhcp",
    "routes": [ { "dst": "10.3.0.0/16" }, { "dst": "10.4.0.0/16" } ]
  }
  // args may be ignored by plugins
  "args": {
    "labels" : {
        "appVersion" : "1.0"
    }
  }
}
```
Example: https://github.com/containernetworking/plugins/blob/fb7a24405e1dfc33bd10be4df5a16609d648db7e/plugins/meta/tuning/tuning.go#L47

SchSeba:
I am not sure how to continue here.

@dcbw @s1061123 can you describe to me how you prefer to implement the extra args field/struct so I can update this PR?

s1061123:
Currently we stop to use environment variable for CNI_ARGS and changed to json cni-args.

https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config

And with 'cni-args` option is standardized in spec and multus.
https://github.com/intel/multus-cni/blob/master/types/types.go#L142

https://github.com/k8snetworkplumbingwg/multi-net-spec/blob/master/v1.1/%5Bv1.1%5D%20Kubernetes%20Network%20Custom%20Resource%20Definition%20De-facto%20Standard.pdf

Hence I close the issue.

Issue #211, opened 2018-12-12
Error in setting  networks status with delegate plugin that returns empty Result


ahalim-intel:
**What happend**:
The sriov cni plugin with DPDK mode has no IPAM config. Therefore returns empty Result. Multus throwing in error as it tries to build network status from this empty Result.

**What you expected to happen**:
Multus should skip delegate results with no IPConfig when aggregating delegate results.

**How to reproduce it (as minimally and precisely as possible)**:
Attache any network CRDs without any IPAM config.

**Anything else we need to know?**:

**Environment**:

- Multus version 
Latest version.
Works fine until 707007abba91b305cb8f946058610891d19b98b5

  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`):
v1.14.0-alpha.0
- Primary CNI for Kubernetes cluster:

- OS (e.g. from /etc/os-release):
CentOS 7
Kernel 4.13.7-1.el7.elrepo.x86_64
- File of '/etc/cni/net.d/'
```
# cat /etc/cni/net.d/00-multus.conf

{
   "name":"defaultnetwork",
   "type":"multus",
   "LogLevel": "debug",
   "LogFile": "/var/log/multus.log",
   "kubeconfig": "/etc/kubernetes/node-kubeconfig.yaml",
   "delegates": [{"cniVersion": "0.3.0", "name":"defaultnetwork", "type": "flannel", "isDefaultGateway": true}],
   "confDir": "/etc/cni/multus/net.d"
}
```
- File of '/etc/cni/multus/net.d'
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)
```
# kubectl get net-attach-def -o yaml
apiVersion: v1
items:
- apiVersion: k8s.cni.cncf.io/v1
  kind: NetworkAttachmentDefinition
  metadata:
    creationTimestamp: "2018-12-11T20:31:16Z"
    generation: 1
    name: sriov-dpdk
    namespace: default
    resourceVersion: "602527"
    selfLink: /apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions/sriov-dpdk
    uid: b5a50e62-fd83-11e8-a46e-408d5c537d23
  spec:
    config: '{ "type": "sriov", "name": "sriov-dpdk-net", "master": "enp2s0f0", "dpdk":
[...]

PR #212, opened 2018-12-13
create empty NetworkStatus for empty Result struct

ahalim-intel:
This fixes the issue described in #211 where LoadNetworkStatus is
throwing in errors if a delegate plugin returns empty Result that
contains no IPAM information.

This change will allow ignoring the errors propagated from parsing
an empty Result and continue with next one.

Change-Id: Ife4b6103de044256233d581fa74759423ed94ff5

### DIFF
--- a/types/conf.go
+++ b/types/conf.go
@@ -122,15 +122,16 @@ func LoadCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, rc 
 func LoadNetworkStatus(r types.Result, netName string, defaultNet bool) (*NetworkStatus, error) {
 	logging.Debugf("LoadNetworkStatus: %v, %s, %t", r, netName, defaultNet)
 
-	// Convert whatever the IPAM result was into the current Result type
-	result, err := current.NewResultFromResult(r)
-	if err != nil {
-		return nil, logging.Errorf("error convert the type.Result to current.Result: %v", err)
-	}
-
 	netstatus := &NetworkStatus{}
 	netstatus.Name = netName
 	netstatus.Default = defaultNet
+
+	// Convert whatever the IPAM result was into the current Result type
+	result, err := current.NewResultFromResult(r)
+	if err != nil {
+		logging.Errorf("error convert the type.Result to current.Result: %v", err)
+		return netstatus, nil
+	}
 
 	for _, ifs := range result.Interfaces {
 		//Only pod interfaces can have sandbox information


booxter:
I've validated the fix for ```vfio``` use case (we use it in ```kubevirt``` to pass devices into managed VMs). The fix is clean.

rkamudhan:
@ahalim-intel @booxter the fix is giving me the network status nil. Basically give entire structure
```
type NetworkStatus struct {
	Name      string    `json:"name"`
	Interface string    `json:"interface,omitempty"`
	IPs       []string  `json:"ips,omitempty"`
	Mac       string    `json:"mac,omitempty"`
	Default   bool      `json:"default,omitempty"`
	DNS       types.DNS `json:"dns,omitempty"`
}
```

I will get the network status without a name and Interface. The ideal case is to get only IPs, MAC and DNS as nil in my network status. Multus is using the current package as all other core CNIs in the community https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L383. So we should have fix in the SRIOV, if it use the current cni package in the l2 mode or vfio mode. we should just make IPAM information, correctly populated.

booxter:
@rkamudhan you are probably right. In case of ```vfio```, what would ```interface``` mean though? If a device is registered with ```vfio```, it means it was unbound from its network driver, so there is no ```netlink``` device when SR-IOV CNI is called. (One may argue that a CNI is not needed in this case at all, but ```multus``` needs something to call to, even if it's doing nothing.)

ahalim-intel:
> @rkamudhan you are probably right. In case of `vfio`, what would `interface` mean though? If a device is registered with `vfio`, it means it was unbound from its network driver, so there is no `netlink` device when SR-IOV CNI is called. (One may argue that a CNI is not needed in this case at all, but `multus` needs something to call to, even if it's doing nothing.)

I second this.

ahalim-intel:
@rkamudhan 
> I will get the network status without a name and Interface. The ideal case is to get only IPs, MAC and DNS as nil in my network status. 

So populating an empty Result struct does not solve this.

If IP is nil we still get the following error:
```
 4279 2018-12-12T10:05:19Z [error] Multus: Err in setting  networks status: error convert the type.Result to current.Result: cannot convert: no valid IP addresses
```
I don't see problem with empty network status where there is no network interface to populate the network status for.  Are we suggesting that even if there's no IPAM config in the sriov-cni config it should return an results struct with dummy IPAM? How this network status will be useful anyway? Besides that, IPAM requires a valid interface in the Container, with a VF with DPDK driver there is no network interface name for that VF. 

IMHO, Multus should respect other plugins that do not invoke IPAM if they do not need to and throwing errors.

rkamudhan:
@ahalim-intel I need the proper network status in the pod spec, not the nil value or skipping it.. it should be populated with what network crd name is plumbed to the pod, In this case, we missing the results. It is not having IP, DNS, & Routes. So change should be in the network status api in Multus, not skipping it.

rkamudhan:
Will create PR on the permanent solution on the empty struct for the spec 0.1.0,0.2.0,0.3.0 and 0.3.1. Thanks @ahalim-intel for the PR and other for the review

PR #214, opened 2018-12-15
Optimize docker image

micw:
Hello,

I did some optimations of the build process and the docker image:
- ensure that the binary is statically linked (it was dynamically linked against glibc due to the "net" dependency - fortunately there's an easy fix)
- Switch to official golang build image rather than build on rhel or centos
- Use a multi-stage docker build process, so that the final image does not contain any build/compile stuff
- Switch to alpine for the delivered image to further reduce the size
- a small update to entrypoint.sh (use a sleep loop rather than "infinity" because it's not implemented in all sleep variantes; trap container shutdown signals for faster shutdown)

Here are some test builds with this image: https://hub.docker.com/r/micwy/test-multus-portmapping/tags

I was able to reduce the image size from 150MB to 15MB.

If you are fine with that change and default to this Dockerfile, we could reduce the size even more by switching from bash to sh for the entrypoint anduse a bussybox base image.

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -178,5 +178,7 @@ fi
 
 echo "Entering sleep... (success)"
 
-# Sleep forever.
-sleep infinity
+# Sleep until container is stopped.
+trap : TERM INT
+(while true; do sleep 3600; done) &
+wait
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,0 +1,20 @@
+# This Dockerfile is used to build the image available on DockerHub
+FROM golang:1.11-alpine AS build
+
+RUN apk add --update bash alpine-sdk
+RUN apk add --update linux-headers
+
+ADD . /usr/src/multus-cni
+
+WORKDIR /usr/src/multus-cni
+
+RUN ./build && file bin/multus
+
+FROM alpine:3.8
+
+RUN apk add --update --no-cache bash
+
+COPY --from=build /usr/src/multus-cni/bin/ /usr/src/multus-cni/bin/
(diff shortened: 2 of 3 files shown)


dougbtv:
Mike -- thanks for taking a stab at this one! I have a number of considerations here, I'm going to hold off on this one until I can make the reciprocal part of this (what's necessary downstream)

The big consideration here for me is that -- I wanted to be able to use the Dockerfile for upstream and downstream distributions, so I couldn't (in the past) do multi-stage builds, nor could I use a base-image that wasn't CentOS/RHEL based.

micw:
@dougbtv What's the issue behind
a) not being able to use multi-stage builds and
b) being forced to use RHEL?

If I understand it better, we can maybe find a better way to fit all needs.

dougbtv:
I think having a dockerfile.alpine and a dockerfile.rhel will probably be the path forward to having all worlds taken care of! 

Re: B) To save the backstory -- the `Dockefile.rhel` separate from the rest will work in teh future.

images/entrypoint.sh
```diff
@@ -178,5 +178,7 @@ fi
 
 echo "Entering sleep... (success)"
 
-# Sleep forever.
-sleep infinity
+# Sleep until container is stopped.
```
intlabs: in a follow up it would be great to make this "sleep until the heat death of the universe" approach optional, so the container can be used as an init container more easily.

dougbtv:
Is this still valid? Dupe of #259

micw:
No, sorry did not saw this.

images/entrypoint.sh
micw: Please look at the new Version at https://github.com/intel/multus-cni/pull/259 - I have added this as an option.

PR #215, opened 2018-12-15
Flannel daemonset remove unused config

micw:
The cni configmap implies that flanell does something with port forwarding. I commented it out to make clear that it has no effect.


micw:
Wrong branch

PR #216, opened 2018-12-15
Remove unused config in flannel-daemonset.yaml

micw:
The cni configmap implies that flannel does something with port forwarding. I commented it out to make clear that it has no effect.

### DIFF
--- a/images/flannel-daemonset.yml
+++ b/images/flannel-daemonset.yml
@@ -55,25 +55,26 @@ metadata:
     tier: node
     app: flannel
 data:
-  cni-conf.json: |
-    {
-      "name": "cbr0",
-      "plugins": [
-        {
-          "type": "flannel",
-          "delegate": {
-            "hairpinMode": true,
-            "isDefaultGateway": true
-          }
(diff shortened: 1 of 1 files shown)


dougbtv:
Mike, I'm OK with this one. The one issue is that this file will probably periodically be updated from the upstream Flannel manifest @ https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml, so, there's the potential for the humans to forget to update this again (the main thing that we remove from the upstream is this copying that config into place and we let Multus do its thing, instead)

Secondarily, and a follow-up to this is this example of how I had manipulated the manifests to make it specific to another method of distribution -- for example here's how I modified it to work with previous versions of OpenShift (likely untested in the last couple months, fwiw), in this case... I have the config as a config map, and then I pass in a param to tell the entrypoint to use that alternate config --> 
https://github.com/dougbtv/openshift-ansible/blob/multus-developer-preview/playbooks/openshift-multinetwork/files/multus-daemonset.yml#L135-L136

PR #220, opened 2018-12-17
 add MULTUS_CONF_FILE to the arr for checking

shahar-klein:
Seems like a typo in the original.


dougbtv:
Looks like a typo to me, thanks @shahar-klein 

Hey @Kusanagi9999 -- does that change look OK to you? (last I knew you were using this for kube-spray installation, iirc?)

PR #231, opened 2019-01-08
add kubelet client for Pod resource info

ahalim-intel:
This change introduces kubelet client to get allocated device
information of a Pod from newly added Kubelet grpc service.
For more information please see:
https://github.com/kubernetes/kubernetes/pull/70508/

With this kubelet service we no longer need to rely on kubelet
checkpoint file to get device information.

Change-Id: I3ba098cfea9a0e798f065a4d9b39d0085e125407
Signed-off-by: Abdul Halim <abdul.halim@intel.com>


rkamudhan:
@dougbtv @s1061123 Targeting for master branch.

dcbw:
This brings in a *ton* of vendored deps, +2million lines. Are we really sure we need all those deps?

ahalim-intel:
> This brings in a _ton_ of vendored deps, +2million lines. Are we really sure we need all those deps?
This PR has following dependencies:
```
	"k8s.io/kubernetes/pkg/kubelet/apis/podresources"
	"k8s.io/kubernetes/pkg/kubelet/apis/podresources/v1alpha1"
	"k8s.io/kubernetes/pkg/kubelet/util"
```
So, I've added only these sub-packages in glide.yaml with correct version then run `glide up -v`.  Looks like even though all I needed is the those sub-packages only, glide added pretty much everything from Kubernetes! That's why the vendor became gigantic! Any suggestions/alternatives how to overcome this?

s1061123:
Hi @ahalim-intel ,

How about to use '--skip-test' option in glide? It may reduce dependent packages.

ahalim-intel:
@s1061123 
> How about to use '--skip-test' option in glide? It may reduce dependent packages.

Actually  glide '--skip-test' is counter-intuitive. It does the opposite. From 'glide up --help':
```
   --skip-test              Resolve dependencies in test files.
```
I've tried it anyway and it didn't change the vendor/ at all. 
I've also tried `glide up -v --no-recursive`. This reduces vendor/ size dramatically but build fails.

ahalim-intel:
Folks, what is the status of this PR?

s1061123:
@ahalim-intel 

- as Dan's comment, your code removes checkpoint code and you remove checkpoint feature. This change removes previous feature and don't have compatibility. How do you support checkpoint?
- your diff is slightly old so could you please rebase it to latest?

ahalim-intel:
> @ahalim-intel
> 
> * as Dan's comment, your code removes checkpoint code and you remove checkpoint feature. This change removes previous feature and don't have compatibility. How do you support checkpoint?
> * your diff is slightly old so could you please rebase it to latest?

@s1061123 
I will be working on this in coming days. I am thinking of introducing an option field in Multus netconf something like `resourceInfoProvider`: with values `checkpoint`(default) OR `kubelet-api`. What do you think?

zshi-redhat:
> @ahalim-intel
> 
> * as Dan's comment, your code removes checkpoint code and you remove checkpoint feature. This change removes previous feature and don't have compatibility. How do you support checkpoint?
> * your diff is slightly old so could you please rebase it to latest?

Tomo, just want to mention that checkpoint file is a temporary solution for getting device information from kubelet and it will be deprecated in future kube releases. we'd like to switch to kubelet endpoint with kube-1.13 and 1.13+. This feature is mainly used for passing device info to SR-IOV CNI, I dont' know if any other CNI use this feature.  I understand that Multus needs maintain backward compatibility, do you know how we can deprecate a feature like checkpoint file in Multus? Like how many kube releases Multus need to maintain backward compat with?

ahalim-intel:
@s1061123 @dougbtv This PR is rebased now. Also addressed the backward compatibility. If Kubelet resource API is available Multus will use that OR else fallback in Kubelet checkpoint.

rkamudhan:
@zshi-redhat Could you please test this PR ?

ahalim-intel:
@s1061123 

I'm bit confused with the travis test output there. 

Output from travisci:
```
# github.com/intel/multus-cni/kubeletclient
kubeletclient/kubeletclient_test.go:106:3: client declared but not used
vet: typecheck failures
FAIL	github.com/intel/multus-cni/kubeletclient [build failed]
```

If you see the closure variable `client` on line 106, it is being used within other context block. It's not 'not used'. What do you suggest?


And running test.sh locally with same file(kubeletclient/kubeletclient_test.go):
```
=== RUN   TestKubeletclient                                                             
Running Suite: Kubeletclient Suite                                                      
==================================                                                      
Random Seed: 1553184260                                                                 
Will run 5 of 5 specs                                                                   

•••••
Ran 5 of 5 Specs in 0.018 seconds
SUCCESS! -- 5 Passed | 0 Failed | 0 Pending | 0 Skipped --- PASS: TestKubeletclient (0.02s)
PASS                                                                                       
coverage: 77.5% of statements                                                              
ok      github.com/intel/multus-cni/kubeletclient       0.023s  coverage: 77.5% of statements

```

ahalim-intel:
Ok, nevermind it. I've move the variable from there and updated the PR.

zshi-redhat:
> @zshi-redhat Could you please test this PR ?

I ran the following tests with this PR applied to multus master branch, all passed:
1) Enable KubeletPodResources feature gate with kube-1.13 env, Multus can detect that kubelet podResource endpoint is enabled and pass the correct device ID that allocated by SR-IOV Device Plugin to SR-IOV CNI.
2) Disable KubeletPodResources (default config) with kube-1.13 env, Multus can fall back to checkpoint file solution to get Device ID information and pass it to SR-IOV CNI.

kubeletclient/kubeletclient.go
```diff
@@ -0,0 +1,113 @@
...
+		return getKubeletClient()
+	} else {
+		logging.Printf(logging.VerboseLevel, "GetResourceClient(): using Kubelet device plugin checkpoint")
+		return checkpoint.GetCheckpoint()
+	}
+}
+
+func getKubeletClient() (types.ResourceClient, error) {
+	newClient := &kubeletClient{}
+	if kubeletSocket == "" {
```
zshi-redhat: This will always be false as ``kubeletSocket`` is set in ``hasKubeletAPIEndpoint``.

kubeletclient/kubeletclient.go
```diff
@@ -0,0 +1,113 @@
...
+				}
+			}
+		}
+	}
+	return resourceMap, nil
+}
+
+func hasKubeletAPIEndpoint() bool {
+	// Check for kubelet resource API socket file
+	kubeletSocket = filepath.Join(defaultPodResourcesPath, defaultKubeletSocketFile)
```
zshi-redhat: why not use the same as ``util.LocalEndpoint(defaultPodResourcesPath, podresources.Socket)``?

kubeletclient/kubeletclient.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	defaultKubeletSocketFile   = "kubelet.sock"
+	defaultPodResourcesPath    = "/var/lib/kubelet/pod-resources"
+	defaultPodResourcesMaxSize = 1024 * 1024 * 16 // 16 Mb
+)
+
+var (
+	kubeletSocket string
```
zshi-redhat: why this is a global var?

ahalim-intel: Need this global here for unit-tests to override it.

kubeletclient/kubeletclient.go
ahalim-intel: > why not use the same as `util.LocalEndpoint(defaultPodResourcesPath, podresources.Socket)`?

`util.LocalEndpoint(defaultPodResourcesPath, podresources.Socket)` returns path with unix:// URL scheme prefix. All we need here simple path string.

kubeletclient/kubeletclient.go
ahalim-intel: > This will always be false as `kubeletSocket` is set in `hasKubeletAPIEndpoint`.

Good catch! I've corrected the variable in `hasKubeletAPIEndpoint` which should be only local.

kubeletclient/kubeletclient.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+	return resourceMap, nil
+}
+
+func hasKubeletAPIEndpoint() bool {
+	// Check for kubelet resource API socket file
+	kubeletSocket = filepath.Join(defaultPodResourcesPath, defaultKubeletSocketFile)
+	if _, err := os.Stat(kubeletSocket); err != nil {
+		logging.Errorf("hasKubeletAPIEndpoint(): error looking up kubelet resource api socket file: %q", err)
```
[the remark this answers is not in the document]
ahalim-intel: > Might be good to have a warning message instead of error.

Agree! I will change this to Printf() as the logging has no Warningf()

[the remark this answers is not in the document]
zshi-redhat: +1, Printf() shall be fine.

kubeletclient/kubeletclient.go
zshi-redhat: gotcha, thanks!

ahalim-intel:
@s1061123 @zshi-redhat @dougbtv 

I've addressed pretty much all the comments and feedback. Also fixed the travis-ci tests. All checks now pass. It's ready to be merged!

Thanks everyone for help reviewing/improving this PR... much appreciated!

PR #238, opened 2019-01-15
Update to entrypoint.sh -- IPv6 environment Kubeconfig

kisanth88:
Updated server in cluster to parse IPv6 IP address in URL properly.

If using IPv6 only Internal network, PODs using Multus fail to start be
cause the URL parses IPv6 segment colon separator as port separator. Ne
ed open and close brackets around URL to allow proper URL parsing.

Error when fail:
Warning  FailedCreatePodSandBox  3m28s (x682 over 116m)  kubelet, k8swo
rker  (combined from similar events): Failed create pod sandbox: rpc er
ror: code = Unknown desc = [failed to set up sandbox container "84716ac
76af98da22b71ba6e6ada42c699cb04ffb42ac75c71b618a93e98e798" network for 
pod "coredns-576cbf47c7-kkdmw": NetworkPlugin cni failed to set up pod 
"coredns-576cbf47c7-kkdmw_kube-system" network: Multus: Err in loading 
K8s Delegates k8s args: Multus: Err in getting k8s network from pod: ge
tPodNetworkAnnotation: failed to query the pod coredns-576cbf47c7-kkdmw
in out of cluster comm: Get https://fd00:1234::1:443/api/v1/namespaces/
kube-system/pods/coredns-576cbf47c7-kkdmw: invalid URL port "1234::1:44
3", failed to clean up sandbox container "84716ac76af98da22b71ba6e6ada4
2c699cb04ffb42ac75c71b618a93e98e798" network for pod "coredns-576cbf47c
7-kkdmw": NetworkPlugin cni failed to teardown pod "coredns-576cbf47c7-
kkdmw_kube-system" network: Multus: Err in getting k8s network from pod
: getPodNetworkAnnotation: failed to query the pod coredns-576cbf47c7-k
kdmw in out of cluster comm: Get https://fd00:1234::1:443/api/v1/namesp
aces/kube-system/pods/coredns-576cbf47c7-kkdmw: invalid URL port "1234:
:1:443"]

Signed-off-by: John Bittenbender <kisanth88@gmail.com>

Please be gentle to the FNG ;-)


dougbtv:
Hey @kisanth88 -- did you check that this still works with IPv4 as well?

Also, can you remove the changes to the daemonset, we'll want these to continue to point at our image builds. Thanks

s1061123:
@kisanth88 , looking https://tools.ietf.org/html/rfc3986, square brackets, '[' and ']', are only used for "IPv6 or later IP protocol literal address", so your fix makes multus only for IPv6 literal address.

So could you please consider to support IPv4/DNS as well as IPv6?

images/multus-daemonset.yml
```diff
@@ -119,7 +119,7 @@ spec:
       serviceAccountName: multus
       containers:
       - name: kube-multus
-        image: nfvpe/multus:latest
+        image: kisanth88/kmultusv6test:latest
```
dcbw: Also, I don't think this change to the image path is intended?

kisanth88: Image change is a mistake, sorry, had updated my fork for testing and the image change accidentally got added to pull request.

We do have other K8S addons working with this, however I did not explicitly retest it with IPv4 due to other priorities. I will make an attempt to retest this as soon as I can. Our lab has been unstable due to some switching upgrades that have been going on. So our business priorities have taken precedence over this request. I am happy to withdraw the pull request for now until I can do some detailed testing if you guys would like; however this is an issue with the entrypoint.sh and multus working with IPv6 only K8S environment that needs to be addressed.

kisanth88: Replying with an idea that I think makes more sense than my original change.

I will try to take a whack at this if I can find some time, but it may be more forward looking and flexible to do a check in the entrypoint.sh script for if $KUBERNETES_SERVICES_HOST is IPv6 and then have two variants of the section with comment starting with:

# Write a kubeconfig file for the CNI plugin.

dougbtv:
This functionality came in on another PR, seems. Thanks regardless!

PR #246, opened 2019-01-25
Caches all pod delegates json for pods deletion without k8s info

### LINKED ISSUE #164: Cache CRD data at the CmdAdd() for graceful pod network deletion
When CRD data is changed before CmdDel(), multus may call wrong CNI plugin (or arguments) at CmdDel(), hence it may be necessary.

### LINKED ISSUE #243: tryLoadK8sPodDefaultNetwork() may send unnecessary error message
`"Error deleting network: tryLoadK8sDelegates: Err in loading K8s cluster default network from pod annotation: tryLoadK8sPodDefaultNetwork: failed to get pod annotation: getPodDefaultNetworkAnnotation: failed to query the pod podest-1 in out of cluster comm: pods "podtest-1" not found"`

Description:
In case of `terminationGracePeriodSeconds: 0`, Kubernetes may invoke cmdDel() after pod removal in Kubernetes API because of `terminationGracePeriodSeconds: 0` and above messages are shown.

s1061123:
This fixes #243. This fix changes to 

* enable delegate cache for all pods
* use the caches if multus cannot get pod information by Kubernetes API

This change also optimizes to call GetPod() to Kubernetes for efficiency.

Fixes #164


dougbtv:
Tomo, just an FYI that I setup an environment to give it a test -- still experiencing the same error condition when I delete.

I realize it's a work in progress, but, just wanted to let you know. 

Thanks regardless for getting this work underway!

k8sclient/k8sclient.go
```diff
@@ -511,24 +504,31 @@ func GetPodNetwork(k8sclient KubeClient, k8sArgs *types.K8sArgs, confdir string,
...
+	logging.Debugf("GetNetworkDelegates: %v, %v, %v, %v, %v", k8sclient, pod, networks, confdir, confnamespaceIsolation)
 	// resourceMap holds Pod device allocation information; only initizized if CRD contains 'resourceName' annotation.
 	// This will only be initialized once and all delegate objects can reference this to look up device info.
 	var resourceMap map[string]*types.ResourceInfo
 
 	// Read all network objects referenced by 'networks'
 	var delegates []*types.DelegateNetConf
+	defaultNamespace := pod.ObjectMeta.Namespace
+
+	podID := pod.UID
```
dcbw: Instead of using the pod.UID to track caching, I would use the sandbox ID that is passed via the CNI args as the "container ID". The problem with using the Pod UID is that it will be the same for all sandboxes for the pod, and you can have overlapping sandboxes if one fails. Kubelet garbage-collects dead sandboxes asynchronously.

So for example, you start a pod and there's a transient error creating the sandbox network. Kubelet will schedule that one for garbage collection but not call CNI DEL yet. Then after a short period of time it will create another sandbox and call CNI ADD and perhaps this one works. Now the original one gets GC-ed but the second is still alive.

So if you use the Pod UID as a key, you might delete data you had cached that still applies to the second sandbox. But each sandbox will have a unique container ID (CNI's CNI_CONTAINERID) that you can use to ensure stuff like this doesn't overlap.

multus/multus.go
```diff
@@ -312,16 +323,14 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
...
 	if err != nil {
 		return nil, logging.Errorf("Multus: Err in loading K8s Delegates k8s args: %v", err)
 	}
 
-	if numK8sDelegates == 0 {
-		// cache the multus config if we have only Multus delegates
-		if err := saveDelegates(args.ContainerID, n.CNIDir, n.Delegates); err != nil {
-			return nil, logging.Errorf("Multus: Err in saving the delegates: %v", err)
-		}
+	// cache the multus config if we have only Multus delegates
```
dcbw: this comment is no longer true right? we're now caching *all* delegates.

multus/multus.go
```diff
@@ -417,23 +426,30 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
...
 	numK8sDelegates, kc, err := k8s.TryLoadPodDelegates(k8sArgs, in, kubeClient)
 	if err != nil {
-		return err
+		// in error case, need to proceed without return for bailout
+		_ = logging.Errorf("Multus: Err in getting k8s network from pod: %v", err)
 	}
 
 	if numK8sDelegates == 0 {
+		logging.Debugf("cmdDel: reading delegates from cache")
 		// re-read the scratch multus config if we have only Multus delegates
```
dcbw: Hmm. Shouldn't we also cover the case of the pod annotation changing before DEL? Unless we admission-control it, the user can delete a network from the selection annotation, and then multus won't clean that network up on DEL.

Perhaps we should do something like:

1) cache all network configs (even cluster-wide default network) on ADD
2) on DEL, tear down the cluster-wide default network. Use the on-disk config for this (because it too could have changed), but if for some reason that doesn't exist, fall back to /etc/cni/net.d or Net Attach object config.
3) Do a readdir() in the sandbox container ID's cache directory to find all other networks to delete from the cache info, and then delete them.
4) only if we cannot find the cached info for the sandbox container ID, ask Kubernetes what the annotations are and use that to determine what networks to delete.

Basically, after a pod ADD we cannot trust the kube API not to change.

multus/multus.go
s1061123: Right, fix soon and update PR.

multus/multus.go
s1061123: Agree, fix it and update PR diff.

s1061123: @dcbw Considering the case, I come up with the idea to 'just use cache in deletion' because not only pod annotation, also defaultNetwork/clusterNetwork may be changed during the pod running. In case of dynamic change of network interface, multus re-cache to track the changes then we can make cache consistent. I will change it and update PRs.

s1061123: But we still need to keep to use k8sclients in cmdDel() to update networkstatus...

multus/multus.go
```diff
@@ -417,23 +426,30 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
...
 				return nil
 			}
-			return logging.Errorf("Multus: Err in  reading the delegates: %v", err)
+			return logging.Errorf("Multus: Err in reading the delegates: %v", err)
 		}
 
 		if err := json.Unmarshal(netconfBytes, &in.Delegates); err != nil {
 			return logging.Errorf("Multus: failed to load netconf: %v", err)
 		}
+	} else {
```
dcbw: @s1061123 this will delete the cache before actually calling the plugins, which leaves the possibility of errors being returned after the cache files are gone. I think we should move the cache deletion to the end of the function, after all other calls have run.

s1061123: Fixed.

multus/multus.go
```diff
@@ -415,25 +424,24 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
...
-				// Per spec should ignore error if resources are missing / already removed
-				return nil
-			}
-			return logging.Errorf("Multus: Err in  reading the delegates: %v", err)
+	// Read the cache to get delegates json for the pod
+	netconfBytes, path, err := consumeScratchNetConf(args.ContainerID, in.CNIDir)
+	defer os.Remove(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			// Per spec should ignore error if resources are missing / already removed
```
dcbw: If we're going to always use the cached delegates we need to make sure that the cache dir is persistent on-disk, eg that the DaemonSet uses a volume mount to persist the cache on the host.

But now we have an upgrade problem. When somebody updates from old multus to this one while the cluster is running (which is very common in OpenShift 4 for example), there won't necessarily be cache files. Then when kubelet deletes the pod, the delete will fail because the cache is empty.

Per https://github.com/intel/multus-cni/pull/246#pullrequestreview-201630484 I think we still do need to fall back to the K8s delegates if the cache is empty. While not ideal, that will fix both the upgrade case and possibly user errors.

So perhaps move the TryLoadPodDelegates() after the consumeScratchNetConf(), and if if os.IsNotExist(err) is true then we use the delegates from TryLoadPodDelegates()?

dcbw: @s1061123 ^

s1061123: @dcbw , thank you for your comment! Now I'm modifying and will update the diff soon.

s1061123: Update the diffs with the fix.

dcbw:
@s1061123 I think the logic is OK now. But can you add a testcase that does an ADD, then explicitly deletes the scratch netconf, then calls DEL and provides a valid fake KubeClient? That may make the coverage tests happier.

s1061123:
agree. will take care of it, soon.

s1061123:
@dcbw could you please check the latest diff?

PR #250, opened 2019-01-28
Feature: default network for specific namespaces

pawansingh126:
Add new map type config field 'namespaceNetwork' for
adding namespace specific default network.


pawansingh126:
@dougbtv Please have a look at committed changes. Added a new config parameter "namespaceNetwork" which lets user specify default network for a namespace. Default network for different namespaces can be provided in this config parameter and in case of no mapping present in the config for a namespace, clusterNetwork is used as default network.

Issue #244, opened 2019-01-24
Bin directory is different when using CRI-O


dougbtv:
When using CRI-O we have a different default binary directory (`/usr/libexec/cni`). I propose that we have an extra daemonset that's used for CRI-O based installations.

Such as: https://github.com/intel/multus-cni/blob/master/images/multus-crio-daemonset.yml

We'd then go and set the bin directory in the mounted dirs: https://github.com/intel/multus-cni/blob/master/images/multus-daemonset.yml#L135-L136

And then we'd add parameters to the entrypoint script so we set these variables: https://github.com/intel/multus-cni/blob/master/images/entrypoint.sh#L7-L8

dougbtv:
Add to [container section](https://github.com/intel/multus-cni/blob/master/images/multus-daemonset.yml#L120-L129) of the daemonset:

```
        command: ["/entrypoint.sh"]
        args: ["--cni-bin-dir=/usr/libexec/cni"]
```

s1061123:
I use `000-multus.conf` for config and it works in case of cri-o.

aneeshkp:
@s1061123 In that case should rename images/70-multus.conf to 000-Multus.conf in repo itself?
Or may be add instructions?

dougbtv:
Hrmmm, maybe we want to name it something like.... `05-multus.conf` or something... higher than 0. The only reason being is that it's nice for debugging if you have to create a `00-mytest.conf`. 

But, I like the idea of reducing it to less than 10, I think that's a good idea.

dougbtv:
WE now have a CRIO deployment yaml

PR #255, opened 2019-01-29
Fix Bin directory is different when using CRI-O

aneeshkp:
Fix for Issue #244: Bin directory is different when using CRI-O 
Added multus-crio-daemonset.yml  and updated readme for crio runtime

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,13 @@ We apply these files as such:
 $ cat ./images/{multus-daemonset.yml,flannel-daemonset.yml} | kubectl apply -f -
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -32,6 +32,13 @@ We apply these files as such:
 $ cat ./images/{multus-daemonset.yml,flannel-daemonset.yml} | kubectl apply -f -
 ```
 
+Note: For crio runtime use multus-crio-daemonset.yml (crio uses /usr/libexec/cni as default path for plugin directory). Before deploying daemonsets,delete all default network plugin configuration files under /etc/cni/net.d
```
dougbtv: Aneesh -- can we move this out of the `README.md` and into the `./docs/quickstart.md`? Just looking to keep README.md mimimal and maybe more options in the quickstart.

PR #258, opened 2019-02-04
multus: simplify teardown on add error and clarify error message

dcbw:
@s1061123 @dougbtv @rkamudhan @danwinship


dcbw:
Now with testcases.

PR #222, opened 2018-12-17
Minimize dockerfile

micw:

PR #259, opened 2019-02-04
Dockerfile improvements

micw:
Follow-up to #222

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,12 @@
 FROM busybox:latest
 
+ENV \
+  CNI_CONF_DIR="/host/etc/cni/net.d" \
+  CNI_BIN_DIR="/host/opt/cni/bin"
+
+
 ADD bin/multus /opt/multus-cni/
 ADD ./images/entrypoint.sh /opt/multus-cni/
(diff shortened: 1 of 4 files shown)


dougbtv:
Thanks @micw -- still a fan of where you're going with this. I'd also like to verify that the entrypoint still works with the `/dockerfile.openshift` build too, that might also require tweaks.

micw:
@dougbtv @intlabs finally got some time to finish it (and rebase to current master).
A docker image built with this change can soon be found at https://hub.docker.com/r/micwy/multus-cni

micw:
@dougbtv Oh, seen that you already commented. Can you use 2-stage builds on openshift?

micw:
@dougbtv could you please check if the modified openshift Dockerfile works for you?

micw:
@dougbtv Can we make any progress with this? Meanwhile people are comitting to the old entrypoint, so I need to modify this PR regulary...

micw:
Hi. I have seen that multus-cni is still being shipped with that large centos based image. Are you still interested in this PR? If so, I'd rebase it to the current master.

PR #264, opened 2019-02-19
Avoid loading network status if IPAM is not configured for delegate

krsacme:
In case of DPDK, where IPAM config is not required, loading the
network status will throug an error of "no valid IP addresses".
Avoid network status update for delegates which does not have
IPAM config.

Signed-off-by: Saravanan KR <skramaja@redhat.com>

Error log with DPDK driver bound a VF:

```
Warning  FailedCreatePodSandBox  9s               
kubelet, dell-r730-044.dsal.lab.eng.rdu2.redhat.com  Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "92d15b26bffd2229f520714cfc156e4b73327c59a83d746eeebc8705cec4ec22" 
network for pod "testpod2": NetworkPlugin cni failed to set up pod "testpod2_default" network: Multus:
Err in setting  networks status: error convert the type.Result to current.Result: cannot convert: no valid IP addresses
```

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -349,7 +349,11 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) (cn
 
 		//create the network status, only in case Multus as kubeconfig
 		if n.Kubeconfig != "" && kc != nil {
-			if !types.CheckSystemNamespaces(kc.Podnamespace, n.SystemNamespaces) {
+			// For non-master plugin, IPAM should be configured in the delegate, for cases where IPAM is
+			// not required, like DPDK, skip the network status
+			if !delegate.MasterPlugin && delegate.Conf.IPAM.Type == "" {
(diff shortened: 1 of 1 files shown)


dcbw:
@krsacme I don't think this PR is correct; we still need to post network status to the Kube API regardless of whether IPAM information exists or not. In the DPDK case, what does the Result acutally contain?

krsacme:
Thanks for your comments. In the [sriov-cni](https://github.com/intel/sriov-cni/blob/master/sriov/main.go#L80) case, the result is empty. Recently, I got to know the DPDKMode in sriov-cni is legacy support. So we can discard this PR.

> If it contains anything useful (interfaces in the container, MAC adresses, DNS info, etc) then the result should get posted

The result processing in [cni](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L183) is expecting the IP address, else it throws error. This has to be enhanced to handle the available result items as you mentioned.

PR #268, opened 2019-02-21
Added the option for default network parameters

SchSeba:
Added the option for default network parameters.

Example:
```yaml
apiVersion: v1
kind: Pod
metadata:
name: pod-example
annotations:
 v1.multus-cni.io/default-network: '[{"mac":"aa:bb:cc:dd:ee:ff"}]'
```

This will merge the configuration with the master plugins and pass the mac data to it.

This PR also allow to configure 1 addition network

Example:
```yaml
apiVersion: v1
kind: Pod
metadata:
name: pod-example
annotations:
 v1.multus-cni.io/networks: '{"mac":"aa:bb:cc:dd:ee:ff"}'
```

without the slice [].

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -274,3 +274,20 @@ annotations:
  v1.multus-cni.io/default-network: calico-conf
 ...
 ```
+
+### Pass additional parameters to the default cluster network
+
+User may want to pass additional parameters for the default network like mac, ip address, etc...
+
+You are able to do it by passing a map in the default network annotation.
+
+Example:
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+name: pod-example
+annotations:
+ v1.multus-cni.io/default-network: '{"mac":"aa:bb:cc:dd:ee:ff"}'
+...
+```
--- a/k8sclient/k8sclient.go
+++ b/k8sclient/k8sclient.go
@@ -204,7 +204,12 @@ func parsePodNetworkAnnotation(podNetworks, defaultNamespace string) ([]*types.N
 
 	if strings.IndexAny(podNetworks, "[{\"") >= 0 {
 		if err := json.Unmarshal([]byte(podNetworks), &networks); err != nil {
-			return nil, logging.Errorf("parsePodNetworkAnnotation: failed to parse pod Network Attachment Selection Annotation JSON format: %v", err)
+			// Try to unmarshal as one element
+			var network *types.NetworkSelectionElement
+			if err := json.Unmarshal([]byte(podNetworks), &network); err != nil {
+				return nil, logging.Errorf("parsePodNetworkAnnotation: failed to parse pod Network Attachment Selection Annotation JSON format: %v", err)
+			}
+			networks = append(networks, network)
 		}
 	} else {
 		// Comma-delimited list of network attachment object names
@@ -445,6 +450,7 @@ func TryLoadPodDelegates(k8sArgs *types.K8sArgs, conf *types.NetConf, kubeClient
 	if delegate != nil {
 		logging.Debugf("tryLoadK8sDelegates: Overwrite the cluster default network with %v from pod annotations", delegate)
 
+		delegate.MasterPlugin = true
 		conf.Delegates[0] = delegate
 	}
 
@@ -602,7 +608,7 @@ func getNetDelegate(client KubeClient, netname, confdir, namespace string) (*typ
 		return delegate, nil
 	}
 
-	// option3) search directry
+	// option3) search directory
 	fInfo, err := os.Stat(netname)
 	if err == nil {
 		if fInfo.IsDir() {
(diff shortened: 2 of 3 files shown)


doc/configuration.md
```diff
@@ -274,3 +274,20 @@ annotations:
...
+You are able to do it by passing a map in the default network annotation.
+
+Example:
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+name: pod-example
+annotations:
+ v1.multus-cni.io/default-network: "{"mac":"aa:bb:cc:dd:ee:ff"}"
```
booxter: shouldn't we escape internal double-quotes here?

k8sclient/k8sclient.go
```diff
@@ -686,6 +692,16 @@ func tryLoadK8sPodDefaultNetwork(k8sArgs *types.K8sArgs, conf *types.NetConf, ku
 		return nil, logging.Errorf("tryLoadK8sPodDefaultNetwork: more than one default network is specified: %s", netAnnot)
 	}
 
+	if networks[0].Name == "" {
```
booxter: I am not sure I follow. Why do we require an empty name?

doc/configuration.md
booxter: shouldn't we also have name of the network, and probably its namespace, in the dict? otherwise how is multus supposed to determine which network to use?

k8sclient/k8sclient.go
booxter: Maybe I miss something. I am not sure I understand what you say.

Let's get to basics.

What is the goal of the `defaultNetwork` annotation? My understanding is that it tells Multus which network should be used instead of default one configured as the primary delegate in its configuration file. This is achieved by listing the corresponding network name in `defaultNetwork` annotation. This annotation was added in: https://github.com/intel/multus-cni/pull/174/

The current limitation of this approach that you try to address is that right now `defaultNetwork` annotation value format is a single network name which user may be interested in passing additional arguments / attributes to the CNI plugin.

The goal of your PR is to make `defaultNetwork` annotation support JSON format that would include the name of the default network but also additional attributes. (This behavior reflects handling of `networks` attribute which may be of two formats: simple comma separated list as well as a JSON list.)

In the example I see added by this PR in `doc/configuration.md`, you only list additional attributes (`mac`) without core attributes (`name`, `namespace`). In this case, it's not possible for Multus to determine which network should be used as default for the pod.

Perhaps your PR adds ability to pass additional attributes to the default cluster network. But by doing so, it also replaces ability to choose a different network to use as default for a pod. Which is a problem.

Is there something I miss in my explanation?

SchSeba: @booxter thanks for the comment I will try to answer

> Perhaps your PR adds ability to pass additional attributes to the default cluster network. But by doing so, it also replaces ability to choose a different network to use as default for a pod. Which is a problem.

this PR adds ability to pass additional attributes in 3 cases.
1. configure only extra parameter (use the multus.conf delegate)
https://github.com/intel/multus-cni/pull/268/files#diff-bdea1124c0b7de33d6bff38302fad2aaR570
2. configure only extra parameter (use the multus.conf clusterNetwork)
https://github.com/intel/multus-cni/pull/268/files#diff-bdea1124c0b7de33d6bff38302fad2aaR605
3. configure default network and extra parameter
https://github.com/intel/multus-cni/pull/268/files#diff-bdea1124c0b7de33d6bff38302fad2aaR641

continue to support the network name in default network annotation
https://github.com/intel/multus-cni/blob/e64db39f44939774d8d082edc4fd5782ba4c187e/k8sclient/k8sclient_test.go#L495

please tell me if this comment answer you question

k8sclient/k8sclient.go
```diff
@@ -204,7 +204,12 @@ func parsePodNetworkAnnotation(podNetworks, defaultNamespace string) ([]*types.N
 
 	if strings.IndexAny(podNetworks, "[{\"") >= 0 {
 		if err := json.Unmarshal([]byte(podNetworks), &networks); err != nil {
-			return nil, logging.Errorf("parsePodNetworkAnnotation: failed to parse pod Network Attachment Selection Annotation JSON format: %v", err)
+			// Try to unmarshal as one element
```
s1061123: parsePodNetworkAnnotation() is mainly used for "k8s.v1.cni.cncf.io/networks" annotation and
previously "v1.multus-cni.io/default-network" is (a kind of) compatible hence we use it to
parse network annotation.

But if you introduce to add NetworkSelectionElement into "v1.multus-cni.io/default-network",
we should not use parsePodNetworkAnnotation() because the requirement is different.

- "k8s.v1.cni.cncf.io/networks" could have one or more network but "v1.multus-cni.io/default-network" must have one network
- "k8s.v1.cni.cncf.io/networks" should have network attachment to identify the network, but "v1.multus-cni.io/default-network" may skip to add network-attachment-network (in your example)

So you may need to new another function, not adding to code into parsePodNetworkAnnotation().

k8sclient/k8sclient_test.go
```diff
@@ -567,6 +567,113 @@ var _ = Describe("k8sclient operations", func() {
 		Expect(netConf.Delegates[0].Conf.Type).To(Equal("mynet1"))
 	})
 
+	It("adds extra data to the default network", func() {
```
s1061123: Could you please add following tests to cover them?

- "v1.multus-cni.io/default-network" has json net-attach-def

PR #269, opened 2019-02-21
Allows cmdDel to finish if netns doesn't exist

### LINKED ISSUE #267: Incomplete deletion when pod is failing
**What happend**: If a pod is in a failed state, when you go to remove it, the deletion may be incomplete. That is, the DEL call to Multus doesn't properly make it all the way to delegate CNI plugin. The logging is also somewhat incomplete.

**What you expected to happen**: Deletes on failed pods always call the delegate CNI plugin.

Label a node so you can reliably put the pod on a node with a node selector:

```
kubectl label nodes {your-worker-node-1} debugger=true
```

Create a pod like so, that intentionally fails, in this case the container waits 5 seconds and then exits non-zero: 

```
cat <<EOF | kubectl create -f -
apiVersion: v1
kind: Pod
metadata:
  name: testpod
spec:
  containers:
  - name: testpod
    command: ["/bin/bash", "-c", "sleep 5; false"]
    image: centos/tools
  nodeSelector:
    debugger: "true"
EOF
```

Watch it and wait until it goes into a failed state. like so,

```
[centos@kube-nonetwork-master ~]$ kubectl get pods 
NAME      READY   STATUS   RESTARTS   AGE
testpod   0/1     Error    1          20s
```

**NOTE** This part is *VERY* important. If you don't time this perfectly while it's in the error state, you won't get the error from Multus.

Delete the pod:

```
[centos@kube-nonetwork-master ~]$ date && kubectl delete pod testpod && date
Tue Feb 19 05:37:33 UTC 2019
pod "testpod" deleted
Tue Feb 19 05:37:35 UTC 2019
```

Watch the logs to see that there is very little information during the delete:

```
2019-02-19T05:37:33Z [debug] LoadDelegateNetConf: {"cniVersion":"0.3.0","name":"weave","plugins":[{"hairpinMode":true,"name":"weave","type":"weave-net"},{"capabilities":{"portMappings":true},"snat":true,"type":"portmap"}]}, <nil>, 
2019-02-19T05:37:33Z [debug] LoadDelegateNetConfList: {"cniVersion":"0.3.0","name":"weave","plugins":[{"hairpinMode":true,"name":"weave","type":"weave-net"},{"capabilities":{"portMappings":true},"snat":true,"type":"portmap"}]}, &{{0.3.0 weave  map[] {} {[]  [] []}} {  []}    false false []}
```

Also happening in OpenShift, [bugzilla reported here](https://bugzilla.redhat.com/show_bug.cgi?id=1679036).

**Environment**:

- Multus version: Snapshot.
- Kubernetes version (use `kubectl version`): v1.13.3 or OKD 4.0
- Primary CNI for Kubernetes cluster: Weave (also happens with OpenShift-SDN)
- OS (e.g. from /etc/os-release): `CentOS Linux release 7.4.1708 (Core) `
- File of '/etc/cni/net.d/': `{ "name": "multus-cni-network", "logFile": "/var/log/multus.log", "logLevel": "debug", "type": "multus", "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig", "delegates": [ { "cniVersion": "0.3.0", "name": "weave", "plugins": [ { "name": "weave", "type": "weave-net", "hairpinMode": true }, { "type": "portmap", "capabilities": {"portMappings": true}, "snat": true } ] } ] }`
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`): (none used)

dougbtv:
More detail in the referenced issue. Generally the gist is that while a pod is in a failed state, the netns that Multus receives doesn't exist and the `cmdDel` function returns prematurely. With the premature return, we don't call the delegates during the delete -- in the case that the delegated CNI plugin needs to perform cleanup functions, this can leave it in a dirty state.

This fix lets the cmdDel continue.

Fixes #267

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -402,11 +402,15 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 		// https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444
 		_, ok := err.(ns.NSPathNotExistErr)
 		if ok {
-			return nil
+			logging.Debugf("cmdDel: WARNING netns may not exist, netns: %s, err: %s", netns, err)
+		} else {
+			return fmt.Errorf("failed to open netns %q: %v", netns, err)
 		}
-		return fmt.Errorf("failed to open netns %q: %v", netns, err)
 	}
-	defer netns.Close()
+
+	if netns != nil {
+		defer netns.Close()
+	}
 
 	k8sArgs, err := k8s.GetK8sArgs(args)
 	if err != nil {


multus/multus.go
```diff
@@ -402,11 +402,15 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient k8s.KubeClient) err
 		// https://github.com/kubernetes/kubernetes/issues/43014#issuecomment-287164444
 		_, ok := err.(ns.NSPathNotExistErr)
 		if ok {
-			return nil
+			logging.Debugf("cmdDel: WARNING netns may not exist, netns: %s, err: %s", netns, err)
```
rkamudhan: if there is no namespace, it should delegate the issue to the plugins ? Example Bridge plugin: https://github.com/containernetworking/plugins/blob/master/plugins/main/bridge/bridge.go#L508
Macvlan plugin: https://github.com/containernetworking/plugins/blob/master/plugins/main/macvlan/macvlan.go#L251

So the conclusion is the delegate plugins should take care of the issues, Multus should delegate it ?

dougbtv: I think so, consider the case without Multus that the plugin is being called directly -- the plugin would have to handle this case just the same.

And in the case where the plugin has cleanup to do that's irrespective of the namespace being there or not,  it can still perform that cleanup.

rkamudhan: That is cool.. Agreed

Issue #265, opened 2019-02-20
Generate the multus config file regardless if MULTUS_CONF_FILE is set to 'auto'


pliurh:
**What would you like to be added**:

Currently, the Multus daemon will only generate file 00-multus.conf when it doesn't exist. But, I propose to generate that file regardless. It means if MULTUS_CONF_FILE is 'auto',  multus daemon shall always generate the file 00-multus.conf, and if that file exists just overwrite it. 

**Why is this needed**:

This change could be beneficial in following scenarios:

1. If Multus is used by Kubernetes default network migration, with this mechanism, we switch the default network in Multus config file by just triggering Multus pod restart.

2. Assuming there might some configuration format change of Multus itself in future release, this mechanism can also facilitate Multus update.

PR #273, opened 2019-02-27
Generate Multus config file regardless

pliurh:
During deployment, the daemonSet will alway generate the multus config file no matter if file '00-multus.conf' exists or not.

Fixes: #265

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -157,12 +157,10 @@ fi
 
 if [ "$MULTUS_CONF_FILE" == "auto" ]; then
   echo "Generating Multus configuration file ..."
-  MASTER_PLUGIN="$(ls $CNI_CONF_DIR | grep -E '\.conf(list)?$' | head -1)"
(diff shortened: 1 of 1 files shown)


images/entrypoint.sh
```diff
@@ -157,12 +157,10 @@ fi
 
 if [ "$MULTUS_CONF_FILE" == "auto" ]; then
   echo "Generating Multus configuration file ..."
-  MASTER_PLUGIN="$(ls $CNI_CONF_DIR | grep -E '\.conf(list)?$' | head -1)"
```
booxter: This is not directly related to your patch, but I think there is a bug here, because this line makes an assumption that the order in which ```ls``` lists directory entries is predetermined (and perhaps alphanumeric). I don't think it's a valid assumption, not on every file system type at least.

pliurh: At least on Linux platform, this assumption is true.
In man of 'ls', it says "List information about the FILEs (the current directory by default).  Sort entries alphabetically if none of -cftuvSUX nor --sort is specified."

PR #280, opened 2019-03-07
Add options to specify logfile & loglevel in entrypoint

dougbtv:
Adds options to specify log level & log file in entrypoint when using config file autogeneration.

Additionally, remove exit on unknown argument for future compatibility in the, ahem, future.


dougbtv:
Just an FYI regarding the Travis tests, note that the "PR" one passes. The "push" one fails, however, this is only due to Travis apparently pulling a branch that does not exist.

PR #284, opened 2019-03-14
Fix issue we the SetNetworkStatus function the kubeclient is null

SchSeba:
when we pass the kubeclient(interface type) and run the GetK8sClient(conf.Kubeconfig, kubeClient)
this is not reflected in the kubeclient object after we finnish the related function

the solution is to run also run the GetK8sClient also inside the SetNetworkStatus function

Related to issue https://github.com/intel/multus-cni/issues/281

I think the issue was introduce here https://github.com/intel/multus-cni/commit/5dc774a5471acfa8a6c43deb4750ff4730ce1fc0#diff-1a409ce08104667a3288dace284c289a

### DIFF
--- a/k8sclient/k8sclient.go
+++ b/k8sclient/k8sclient.go
@@ -81,12 +81,18 @@ func setKubeClientInfo(c *clientInfo, client KubeClient, k8sArgs *types.K8sArgs)
 	c.Podname = string(k8sArgs.K8S_POD_NAME)
 }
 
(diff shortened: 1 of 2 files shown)


SchSeba:
@s1061123 @dougbtv @dcbw can you please take a look on this PR?

SchSeba:
Duplicate by https://github.com/intel/multus-cni/pull/283

PR #286, opened 2019-03-21
Updates image tagging / adds version tagged images to Daemonsets

### LINKED ISSUE #260: Proposal for improvements to releases and image tagging
Here's my proposal for the changes to releases / image tagging.

## Release schedule

We should tag a release soon (as v3.2), and come up with a schedule for having a regularly tagged release (I propose every 3 months), we haven't had a tagged release in 6 months.

In concert with this, we'll also update the `images/multus-daemonset.yaml` to point at the newly tagged release, e.g.

```
image: nfvpe/multus:v3.2
```

## Improved image tagging

I propose that we tag a `:latest` image with each merge into master, this will make for a more intuitive use of the `:latest` tag. Previously, we only tagged `:latest` with each tagged release, and we would tag `:snapshot` with each merge into master (e.g. the `:snapshot` tagged image was equivalent to tip of master)

On the tagging of each release, we'll tag an image with the version, e.g. `vX.Y` (as we currently do) but, we'll also tag this image as `:stable`.

dougbtv:
This PR updates our image tagging scheme, now the `:latest` tagged image will be equivalent to `:snapshot` (we keep snapshot for historical reasons, but, both `:latest` and `:snapshot` will now be the builds from the tip of master)

In order to retain stability for users -- this also updates the daemonsets to point at a tagged release image -- in this case, it anticipates the `:v3.2` tag.

In preparation for this, and before the release v3.2 is tagged -- I have tagged a v3.2 release image so we can merge this now before we officially tag. So that users aren't stuck with a daemonset that can't pull an image. That image tag will be overwritten when we make the tag. (image tags viewable here: https://hub.docker.com/r/nfvpe/multus/tags)

Fixes #260

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,10 @@ deploy:
     script: >
       bash -c '
       docker tag nfvpe/multus nfvpe/multus:$TRAVIS_TAG;
+      docker tag nfvpe/multus nfvpe/multus:stable;
       docker login -u "$REGISTRY_USER" -p "$REGISTRY_PASS"; 
-      docker push nfvpe/multus; 
+      docker push nfvpe/multus;
+      docker push nfvpe/multus:stable;
       docker push nfvpe/multus:$TRAVIS_TAG;
       echo foo'
     on:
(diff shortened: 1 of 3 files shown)


rkamudhan:
@dougbtv Please rebase and merge with master.

PR #287, opened 2019-03-22
Enable Rolling Update on defined DaemonSets

francares:
With RollingUpdate update strategy, after you update a DaemonSet template, old DaemonSet pods will be killed, and new DaemonSet pods will be created automatically, in a controlled fashion.

I added this configuration in order to deal with updates automatically at the moment of updating multus/etc docker image version in running DaemonSets

### DIFF
--- a/examples/multus-with-flannel.yml
+++ b/examples/multus-with-flannel.yml
@@ -91,6 +91,8 @@ metadata:
     tier: node
     app: multus
 spec:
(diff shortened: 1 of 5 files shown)


dougbtv:
@francares -- awesome touch up on this one, good eye and good idea! LGTM. I'll let someone else give a thumbs up and then we can merge it in, appreciate the contribution.

Issue #289, opened 2019-03-25
SR-IOV doesn't work with Multus using conflist


zshi-redhat:
**What happend**:

SR-IOV CNI doesn't recognize the ``deviceID`` passed from Multus in the case of ``conflist`` (chained plugin config format) is used.

**What you expected to happen**:

Multus passes the correct conflist format to SR-IOV CNI.

**How to reproduce it (as minimally and precisely as possible)**:

Follow the instructions in SR-IOV Network Device Plugin [README](https://github.com/intel/sriov-network-device-plugin) to create pods with SR-IOV device attached to it.
replace sriov-crd.yaml net-attach-def Custom Resource with:
```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: sriov-net1
  annotations:
    k8s.v1.cni.cncf.io/resourceName: intel.com/sriov
spec:
  config: '{
        "name": "sriov-network",
        "plugins": [
                {
                        "type": "sriov",
	                "ipam": {
		                "type": "host-local",
		                "subnet": "10.56.217.0/24",
		                "routes": [{
			                "dst": "0.0.0.0/0"
		                }],
		                "gateway": "10.56.217.1"
                        }
	        }
        ]
}'
```
Create pods requesting SR-IOV resource again.

**Anything else we need to know?**:

**Environment**:

- Multus version 
  image path and image ID (from 'docker images')

Multus master branch

- Kubernetes version (use `kubectl version`):

any kube version, for example, 1.12, 1.13

- Primary CNI for Kubernetes cluster:

Multus

- OS (e.g. from /etc/os-release):

CentOS 7.6

- File of '/etc/cni/net.d/'
```
{
	"name": "multus-cni-network",
	"type": "multus",
	"logLevel": "debug",
	"logFile": "/var/log/multus.log",
	"delegates": [{
		"type": "flannel",
		"delegate": {
			"isDefaultGateway": true
		}
	}],
	"kubeconfig": "/etc/kubernetes/admin.conf"
}
```

- File of '/etc/cni/multus/net.d'

N/A

- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)
- Target pod yaml info (with annotation, use `kubectl get pod <podname> -o yaml`)
- Other log outputs (if you use multus logging)

```
[...]

PR #290, opened 2019-03-26
fix missing deviceID in NetConfList

ahalim-intel:
This patch fixes the issue described in #289 where deviceID for
delegate plugin was not adding properly if the plugin conf inside
NetConfList.

Change-Id: I1d221f6b0e60a5b888b8e823611dfe12635e6897
Signed-off-by: Abdul Halim <abdul.halim@intel.com>

### DIFF
--- a/types/conf.go
+++ b/types/conf.go
@@ -31,7 +31,7 @@ const (
 	defaultConfDir                = "/etc/cni/multus/net.d"
 	defaultBinDir                 = "/opt/cni/bin"
 	defaultReadinessIndicatorFile = ""
-	defaultMultusNamespace         = "kube-system"
+	defaultMultusNamespace        = "kube-system"
 )
 
 func LoadDelegateNetConfList(bytes []byte, delegateConf *DelegateNetConf) error {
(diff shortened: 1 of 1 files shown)


types/conf.go
```diff
@@ -276,6 +280,41 @@ func delegateAddDeviceID(inBytes []byte, deviceID string) ([]byte, error) {
...
+		return nil, logging.Errorf("addDeviceIDInConfList(): unable to typecast pMap")
+	}
+	// Inject deviceID
+	firstPlugin["deviceID"] = deviceID
+
+	configBytes, err := json.Marshal(rawConfig)
+	if err != nil {
+		return nil, logging.Errorf("addDeviceIDInConfList(): failed to re-marshal: %v", err)
+	}
+	logging.Verbosef("addDeviceIDInConfList(): updated configBytes %s", string(configBytes))
```
s1061123: Could you move it to Debugf()?

zshi-redhat:
Verified this patch works with SR-IOV CNI config wrapped in ``NetConfList plugins``.

types/conf.go
ahalim-intel: Sure

PR #300, opened 2019-04-03
Remove Flannel from default installation method in favor of auto-config

dougbtv:
* Removes Flannel instructions from quick-start method
* Updates documentation to reflect auto-configuration method
* Removes outdated files.


README.md
```diff
@@ -29,7 +29,7 @@ Firstly, clone this GitHub repository. We'll apply files to `kubectl` from this
 We apply these files as such:
```
s1061123: We also need to add some description that we want users to install CNI plugin for k8s cluster network because previous version we expect that user don't install any CNI plugin, as well as you mentioned that `doc/quickstart.md`

dougbtv: Thanks for the eye on this one, totally missed that paragraph in the main README, appreciate it.

PR #307, opened 2019-04-25
assign device id to pciBusID in delegated config

zshi-redhat:
This allows host-device plugin to recognize
Device PCI address passed from Multus.
It is related to the change in host-device which
enables use of device pci address as a config option:
https://github.com/containernetworking/plugins/pull/300


zshi-redhat:
@s1061123 PTAL

zshi-redhat:
> Looks good to me. More than happy if you add test code ;)

Thanks Tomo!  Added test cases for device ID assignment.

zshi-redhat:
@rkamudhan @ahalim-intel PTAL

PR #309, opened 2019-04-30
Adds --multus-autoconfig-dir and entrypoint documentation

dougbtv:
Basically the gist behind this is that when using the autoconfiguration option, this gives an option to store those configurations in an alternate directory. 

This is useful for cases where you need finer grained control of when you denote that the node is ready or not (as advertised by the Kubelet when a CNI configuration is present)

The initial impetus for this is that with openshift-sdn, we need to both wait that openshift-sdn is fully initialized, and we do not want to launch any pods on a particular before our Multus configuration is in place (but only after openshift-sdn is fully initalized).

### DIFF
--- a/doc/how-to-use.md
+++ b/doc/how-to-use.md
@@ -1,4 +1,4 @@
-## How to use multus-cni?
+## Multus CNI usage guide
 
 ### Prerequisites
 
@@ -7,7 +7,7 @@
 
 Your Kubelet(s) must be configured to run with the CNI network plugin. Please see [Kubernetes document for CNI](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#cni) for more details.
 
-### Install multus
+### Install Multus
 
 Generally we recommend two options: Manually place a Multus binary in your `/opt/cni/bin`, or use our [quick-start method](quickstart.md) -- which creates a daemonset that has an opinionated way of how to install & configure Multus CNI (recommended).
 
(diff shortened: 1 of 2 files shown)


dcbw:
@dougbtv the thought being that a Multus daemonset would set --multus-autoconfig-dir=XYZ, and then cluster-wide default network daemonsets would map their inside-container-/etc/cni/net.d to XYZ on the host? Would secondary networks also go into XYZ or still /etc/cni/net.d?

dougbtv:
@dcbw :

> Multus daemonset would set --multus-autoconfig-dir=XYZ, and then cluster-wide default network daemonsets would map their inside-container-/etc/cni/net.d to XYZ on the host?

That's right the default network daemonsets would place their configurations in XYZ, and Multus during initialization would wait for the configurations to appear on XYZ, and then could write the Multus configuration in the official /etc/cni/net.d (or whatever your Kubernetes CNI location is, which can also be set)

> Would secondary networks also go into XYZ or still /etc/cni/net.d?

Multus still would need to have flat file configurations stored in /etc/cni/net.d, which may be a limitation here.

doc/how-to-use.md
```diff
@@ -468,3 +468,67 @@ $ kubectl exec -it pod-case-06 -- ip -d address
...
+
+    --cni-conf-dir=/host/etc/cni/net.d
+
+This is the configuration directory where Multus will write its configuration file.
+
+    --cni-bin-dir=/host/opt/cni/bin
+
+This the directory in which the Multus binary will be installed.
+
+    --namespace-isolation=false
```
rkamudhan: false is the default option, right?

PR #312, opened 2019-05-07
Allow system:nodes to get network-attachment-definitions

nirmoy:
multus fails retrieve delegates without kubernetes user binding:
```
kubectl get --raw=/apis/k8s.cni.cncf.io/v1/namespaces/default/network-attachment-definitions/macvlan-conf --kubeconfig=/etc/kubernetes/kubelet.conf
Error from server (Forbidden): network-attachment-definitions.k8s.cni.cncf.io "macvlan-conf" is forbidden: User "system:node:chapman" cannot get resource "network-attachment-definitions" in API group "k8s.cni.cncf.io" in the namespace "default"
```

### DIFF
--- a/examples/clusterrole.yml
+++ b/examples/clusterrole.yml
@@ -17,3 +17,16 @@ rules:
     verbs:
       - get
       - update
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: system:kube-nodes
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: multus-crd-overpowered
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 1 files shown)


s1061123:
In https://github.com/intel/multus-cni/blob/master/examples/README.md#rbac-configuration, we mentioned as following, so we do not need to add clusterrolebinding in yaml file.

---- from README.md#rbac-configuration
You will then create a clusterrolebinding for each hostname in the Kubernetes cluster. Replace HOSTNAME below with the host name of a node, and then repeat for all hostnames in the cluster.

kubectl create clusterrolebinding multus-node-HOSTNAME \
    --clusterrole=multus-crd-overpowered \
    --user=system:node:HOSTNAME
----

In addition, we are going to recommend to use 'daemonset-style' install and this also makes service account for multus with clusterrole/clusterrolebinding, so it does not met the issue.

Issue #227, opened 2018-12-20
Calico examples for daemonset-style installation


dougbtv:
**What would you like to be added**: I talked to Vivek at Kubecon, and we got talking about Multus, and he's been using Calico in his lab environment for the default network. We noted that it'd be beneficial to have this available as an option.

**Why is this needed**: Having other default network examples outside of Flannel can give users other options when experimenting with Multus initially and provide more flexibility in general.

intlabs:
@dougbtv this would be really nice to have, I've been attempting to set up multus with calico for some time with limited success - so pointers in the right direction would be very appreciated.

lefranco6910:
Hi,

This is my recipe to use Multus over Calico (05-multus.conf).

This file must be dropped in the /etc/cni/net.d folder. Make sure it has the lowest index of all the files in there ("05" in the example). Restart Kubelet. You should be all set. Make sure "nodename" is adjusted to the hostname of the K8s host. Also, the certificates must be in the required locations.

[05-multus.txt](https://github.com/intel/multus-cni/files/2773545/05-multus.txt)

dougbtv:
@lefranco6910 -- thank you a *ton* for posting this configuration! I'll give this technique a try (Vivek had some nice feedback about how well it worked for him) and I'll use it as a basis to incorporate into our documentation as an example of how to use Calico with Multus, thanks!!!

rkamudhan:
@dougbtv we should add both flannel and calico multus daemonset in the https://github.com/kubernetes/website/blob/master/content/en/docs/setup/independent/create-cluster-kubeadm.md as PR

s1061123:
https://github.com/K8sNetworkPlumbingWG/reference-deployment/tree/master/multus-calico contains calico yaml for multus, so it should be closed.

ragkota:
When I use this method, I run into permission issue with calico-kubeconfig. I see that this method already create a calico-kubeconfig under /etc/cni/multus/calico/net.d/ directory, but calico.conflist still points to /etc/cni/net.d/calico-kubeconfig. Is this an issue that is causing permission problem in my case? Thanks 

[...]

PR #314, opened 2019-05-08
Update YAML files to support Calico

kairen:
The commit has updated YAML files in the `images` dir to support Calico network. #227.


kairen:
@dougbtv Thank you for your review, sorry that I missed #300 changes. On account of the above suggestion, I will create a PR to https://github.com/K8sNetworkPlumbingWG/reference-deployment in this weekend, as Multus + Calico reference of deployment.

PR #316, opened 2019-05-10
Change '--multus-conf-file=auto' to multus default install method

### LINKED ISSUE #313: Make daemonset installer as default

Currently [images/entrypoint.sh](https://github.com/intel/multus-cni/blob/master/images/entrypoint.sh) seems to have enough feature to install multus with various CNI plugins, hence we could make it as default install method.

This change requires followings:

- Change daemonset installer generic (`--multus-conf-file=auto`)
- Remove flannel daemonset (and move it to K8sNetworkPlumbingWG/reference-deployment)
- Change quickstart doc
- Change README.md

**Why is this needed**:

We could provide a way to install multus not only with flannel, also with other CNI plugins as cluster network.

s1061123:
This change fixes #313, change 'multus-conf-file=auto' as multus
default install method. Flannel deployment yaml goes to NPWG's
reference-deployment repo as following.

https://github.com/K8sNetworkPlumbingWG/reference-deployment


s1061123:
closed due to dupped.

PR #318, opened 2019-05-15
Update libcni to 0.7.0 (Spec ver 0.4.0)

s1061123:


s1061123:
Need to resubmit due to nested vendoring.

PR #328, opened 2019-06-05
Add compatibility to Calico/Canal manual config.

AxeyGabriel:
Hi, I updated this file to add compatibility with Calico/Canal CNI
There is a entry in those CNI's configuration where we need to put the node name in one of the entries
Here is a bit of my configuration for calico to work:

  cni-conf.json: 
    .......
      "delegates": [
        {
          "cniVersion": "0.3.1",
          "name": "default-cni-network",
          "plugins": [
            {
              "name": "k8s-pod-network",
              "cniVersion": "0.3.0",
              "type": "calico",
              "log_level": "info",
              "datastore_type": "kubernetes",
              "nodename": "__KUBERNETES_NODE_NAME__", # << HERE
              "ipam": {
                "type": "host-local",
                "subnet": "usePodCidr"
              },
              "policy": {
                "type": "k8s"
              },
      .......

You can see the entry "nodename": "__KUBERNETES_NODE_NAME__"
My pull request will find the __KUBERNETES_NODE_NAME__ string and replace it by the hostname retrieved by kubernetes spec.nodeName, then save it in an emptydir and link it to the multus container
This way calico can work without a problem, otherwise there will be an error by calico refusing the __KUBERNETES_NODE_NAME__ string

### DIFF
--- a/images/multus-daemonset.yml
+++ b/images/multus-daemonset.yml
@@ -73,9 +73,10 @@ data:
   # change the "args" line below from
   # - "--multus-conf-file=auto"
   # to:
-  # "--multus-conf-file=/tmp/multus-conf/70-multus.conf"
+  # "--multus-conf-file=/tmp/expanded-conf/00-multus.conf"
   # Additionally -- you should ensure that the name "70-multus.conf" is the alphabetically first name in the
   # /etc/cni/net.d/ directory on each node, otherwise, it will not be used by the Kubelet.
+  # NOTE: The string __KUBERNETES_NODE_NAME__ will be replaced by the node name.
   cni-conf.json: |
     {
       "name": "multus-cni-network",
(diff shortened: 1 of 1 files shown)


s1061123:
Hi @AxeyGabriel , thank you for your PR!

I looked your PR and it seems for Calico/Canal specifically. Currently multus has another repo for multus daemonset that is for specific deployments in K8s NPWG (Network Plumbing Working Group).
https://github.com/K8sNetworkPlumbingWG/reference-deployment

`images/multus-daemonset.yml` is for easy-to-setup multus for general CNI plugin (as cluster network) and we understand that it may not fit to some CNI plugin, hence we create reference-deployment repo for the reason.

If you don't mind, could you please submit PR to https://github.com/K8sNetworkPlumbingWG/reference-deployment?

AxeyGabriel:
Oh sorry for the response time
I wasn't aware of this another repository
I was looking there now and found that they already did what i did, but in a much better way haha
It won't be necessary to submit this PR there, and i'm also closing it here
Thanks

PR #332, opened 2019-06-07
[minor][docs] changes sleep in docs to example that uses a trap INT

dougbtv:
changes sleep in docs to example that uses a trap INT


dougbtv:
@SchSeba -- it might be redundant after all! I wound up iterating it a few times and this is what wound up working... I probably should've tried to remove that `& wait` and see if it still worked, missed that it was redundant.

PR #333, opened 2019-06-09
Update quickstart.md

kempy007:
updated prerequisites to include that interface needs to be in promiscuous mode

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -29,6 +29,8 @@ kubectl get nodes
 
 In the case that your default network is ready
 
+Your network interface will need to be in promiscuous mode.
+
 ## Installation
 
 Our recommended quickstart method to deploy Multus is to deploy using a Daemonset (a method of running pods on each nodes in your cluster), this spins up pods which install a Multus binary and configure Multus for usage.


s1061123:
Hi @kempy007 

I don't know why promiscuous mode is required. In my case, promiscuous mode is not required with flannel + multus.
Could you please clarify why?

kempy007:
It would not work on 'Centos Atomic Host' with Openshift 3.11, it was not until I ran tcpdump and realised it only worked whilst the tool was running, which drove me to the conclusion that it would only work with promisc mode on. Can you check if your host already has it set with ``` ip a show [Interface] | grep -i promisc```

kempy007:
You also find more details on the general slack channel https://intel-corp-team.slack.com/

s1061123:
If you talked about macvlan, it is expected behavior. macvlan interface does not allow to communicate with its master interface in the host due to linux network design.
Please see the following doc (of docker):

https://docs.docker.com/v17.09/engine/userguide/networking/get-started-macvlan/#macvlan-bridge-mode-example-usage
> When using macvlan, you cannot ping or communicate with the default namespace IP address. For example, if you create a container and try to ping the Docker host’s eth0, it will not work. That traffic is explicitly filtered by the kernel modules themselves to offer additional provider isolation and security.

Macvlan 'mode bridge' allows to communicate among macvlan interface in the host, but it does not to communicate to container host (master interface). You can test it with 'ip link' command (without OpenShift,k8s and multus).

Of course, if you use bridge CNI plugin, it is not happen, hence it is not Multus issue.

PR #351, opened 2019-07-26
Support pre v3.2 "interfaceRequest"

giovanism:
What I've done:
- Update docs to use new `interface` property
- Implement Unmarshaler that honors `interfaceRequest` alternatif to `interface`

Fix https://github.com/intel/multus-cni/issues/294

### DIFF
--- a/types/conf.go
+++ b/types/conf.go
@@ -330,3 +330,42 @@ func CheckSystemNamespaces(namespace string, systemNamespaces []string) bool {
 	}
 	return false
 }
+
+// UnmarshalJSON unmarshal Network Selection Annotation. This also supports
+// the deprecated "interfaceRequest" property.
+func (net *NetworkSelectionElement) UnmarshalJSON(bytes []byte) error {
+	var networkSelectionMap map[string]string
+
+	if err := json.Unmarshal(bytes, &networkSelectionMap); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


types/conf.go
```diff
@@ -330,3 +330,42 @@ func CheckSystemNamespaces(namespace string, systemNamespaces []string) bool {
 	}
 	return false
 }
+
+// UnmarshalJSON unmarshal Network Selection Annotation. This also supports
+// the deprecated "interfaceRequest" property.
+func (net *NetworkSelectionElement) UnmarshalJSON(bytes []byte) error {
```
SchSeba: where you call this function?

s1061123: UnmarshalJSON() will be invoked in Unmarshal(NetworkSelectionElement).

See followings:
- https://golang.org/pkg/encoding/json/#RawMessage.UnmarshalJSON
- https://mariadesouza.com/2017/09/07/custom-unmarshal-json-in-golang/

giovanism: Thanks for your explanation @s1061123 . Though I'm not really sure if types/conf.go is the right place to write this function.

nicklesimba: I'll look into unit tests for this.

s1061123: @giovanism we may change the function to somewhere in future, but for now, it is okey.

PR #354, opened 2019-08-01
Unit tests and update to quickstart guide

nicklesimba:
Added additional unit tests and added a stub section for Network Status Annotations in quickstart.md


kubeletclient/kubeletclient_test.go
```diff
@@ -115,6 +115,15 @@ var _ = Describe("Kubelet resource endpoint data read operations", func() {
 			_, err := GetResourceClient()
 			Expect(err).NotTo(HaveOccurred())
 		})
+
+		It("should fail with missing file", func() {
+			// os.Mkdir("tempdir", 0644)
```
s1061123: Could you remove commented out code (if it is not necessary)?

kubeletclient/kubeletclient_test.go
```diff
@@ -115,6 +115,15 @@ var _ = Describe("Kubelet resource endpoint data read operations", func() {
...
+			kubeletSocket = "Andrew Bays"
```
s1061123: Good example to us but not so universal :laughing:  We may use simple string for that, like "sampleSocketString".

k8sclient/k8sclient_test.go
```diff
@@ -611,4 +613,187 @@ var _ = Describe("k8sclient operations", func() {
...
+	        "type": "weave-net"
+	    }],
+	  "runtimeConfig": {
+	      "portMappings": [
+	        {"hostPort": 8080, "containerPort": 80, "protocol": "tcp"}
+	      ]
+	    }
+	}`
+
+		delegate, err := types.LoadDelegateNetConf([]byte(conf), nil, "0000:00:00.0")
```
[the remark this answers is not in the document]
nicklesimba: I changed this in my most recent commit, but I'm not sure what "non-sriov case" means. Can you explain so I will know where/why such changes should be made for future? Thank you

PR #380, opened 2019-09-19
Removes duplicate NetworkAttachment CRD ref

trozet:
The CRD is now defined in the NetworkPlumbingWG org, along with
generated libs for client, listers, informers, and deep copy functions.
Therefore remove the duplicate definition here, and use the standard
definition.

Signed-off-by: Tim Rozet <trozet@redhat.com>


dougbtv:
As Tomo mentions, let's wait until next version of NPWG spec passes and any pending changes that address new spec version before we merge, in case we want to bump version of vendored lib. 

Also TY for the contribution!!

s1061123:
@trozet now network-attachment-definition-client changed to support K8s 1.16, so could you please update vendor with go mod command?

https://github.com/K8sNetworkPlumbingWG/network-attachment-definition-client

trozet:
@s1061123 can you please also review?

k8sclient/k8sclient.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/intel/multus-cni/kubeletclient"
 	"github.com/intel/multus-cni/logging"
 	"github.com/intel/multus-cni/types"
+	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
```
s1061123: Can we use 'github.com/K8sNetworkPlumbingWG/network-attachment-definition-client' instead of 'github.com/k8snetworkplumbingwg/network-attachment-definition-client/'?

trozet: will fix, thanks

PR #399, opened 2019-11-05
multus: print pod namespace/name in cmdAdd/cmdDel error messages

dcbw:
Helps when debugging messages like:

```
Nov 05 17:43:29 ip-10-0-136-174 crio[1881]: 2019-11-05T17:43:29Z [error] Multus: error adding pod to network "ovn-kubernetes": delegateAdd: error invoking DelegateAdd - "ovn-k8s-cni-overlay": error in getting result from AddNetwork: CNI request failed with status 400: 'Nil response to CNI request
```

Signed-off-by: Dan Williams <dcbw@redhat.com>

@dougbtv @s1061123

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -359,16 +359,24 @@ func delPlugins(exec invoke.Exec, argIfname string, delegates []*types.DelegateN
 	return nil
 }
 
+func cmdErr(k8sArgs *types.K8sArgs, format string, args ...interface{}) error {
+	prefix := "Multus: "
+	if k8sArgs != nil {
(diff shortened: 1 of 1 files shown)


multus/multus.go
```diff
@@ -359,6 +359,11 @@ func delPlugins(exec invoke.Exec, argIfname string, delegates []*types.DelegateN
 	return nil
 }
 
+func cmdErr(k8sArgs *types.K8sArgs, format string, args ...interface{}) error {
+	msg := fmt.Sprintf(format, args...)
+	return logging.Errorf("Multus: [%s/%s]: %s", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, msg)
```
s1061123: Could you please add alternative return in case of k8sArgs==nil for future?

dcbw: Done!

PR #408, opened 2019-11-13
Entrypoint updates

micw:
- allow vars to be overridden by env vars
- don't use "sleep infinity" since it's not implemented by all sleep commands (e.g. not in busybox)
- allow the sleep to be interrupted to allow graceful container termination

I pulled this out from https://github.com/intel/multus-cni/pull/259 to make the PR more granular.


dougbtv:
@micw thanks a bunch I like this pull request a lot! Let me give it a shot in my environment, and I'll merge it in.

PR #409, opened 2019-11-13
Multus kubeconfig is not correct for IPv6 Installs

svallala:
Issue:
/etc/cni/net.d/multus.d/multus.kubeconfig is not properly setup for IPv6
installs

Change Summary:

If the API server host is IPv6 address then the server URL should
include square braces to separate the IP address and port correctly

Signed-off-by: shravan@robin.io


svallala:
Minor change to include square braces no matter what. This is how other CNIs install their kubeconfig file on the node

s1061123:
In this case, IPv4 address kubeconfig will be failed, correct?
Could you please support IPv4 case as well because many user still uses IPv4.

svallala:
@s1061123 using with braces will work both for IPv4 and IPv6.

svallala:
@s1061123 do you need any other changes on my end for this to be committed.

s1061123:
merged. Thank you for your PR, @svallala !

PR #422, opened 2019-12-12
reduce binary size

oilbeater:


dougbtv:
Thanks @oilbeater for the contribution, we'd prefer to have the function names available during debugging, so we'll skip this optimization for now. If you'd like to join our maintainers meeting to discuss, we'd love to open the conversation again. Thanks again!

PR #423, opened 2019-12-16
Add Dockerfile for building image on arm64 platform

TrevorTaoARM:
Add Dockerfile.arm64 for building Multus image on arm64 platform.

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,0 +1,20 @@
+# This Dockerfile is used to build the image available on DockerHub
+FROM golang:1.13.4 as build
+
+# Add everything
+ADD . /usr/src/multus-cni
+
+#ENV GOARCH "arm64"
+#ENV GOOS "linux"
+
+RUN  cd /usr/src/multus-cni && \
+     ./build
+
+# build arm64 container
+FROM centos:7
+COPY --from=build /usr/src/multus-cni /usr/src/multus-cni
+
+WORKDIR /
+ADD ./images/entrypoint.sh /
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -4,8 +4,8 @@ FROM golang:1.13.4 as build
 # Add everything
 ADD . /usr/src/multus-cni
 
-#ENV GOARCH "arm64"
-#ENV GOOS "linux"
+ENV GOARCH "arm64"
+ENV GOOS "linux"
 
 RUN  cd /usr/src/multus-cni && \
      ./build


s1061123:
Hi Trevor, thank you so much for the PR!

I've tested with your dockerfile, the multus-cni binary is built successfully, however, this base image uses 'centos:7' hence x86_64 binary still contained in the container image as following.

```
[tohayash@tohayash-srv multus-cni-amd64]$ docker run --rm --privileged --entrypoint=bash -it multus-cni:arm64
[root@deb33a316cfc /]# yum install file
Loaded plugins: fastestmirror, ovl
(snip)
[root@deb33a316cfc /]# file /bin/bash
/bin/bash: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=7e644dee920bc3ba797c38e05383286563712b49, stripped
[root@deb33a316cfc /]# file ./usr/src/multus-cni/bin/multus 
./usr/src/multus-cni/bin/multus: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, not stripped
```

We expects that user use this container with [multus-daemonset.yml](https://github.com/intel/multus-cni/blob/master/images/multus-daemonset.yml), hence not only multus binary also other binary should be for same architecture because multus container image uses bash in its install script.

You can change base image to `arm64v8/centos:7` as following and this makes container images for arm64.
Could you please change it?

```
[tohayash@tohayash-srv multus-cni-amd64]$ git diff
diff --git a/Dockerfile.arm64 b/Dockerfile.arm64
index 50126ef8..2105f2ad 100644
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -11,7 +11,7 @@ RUN  cd /usr/src/multus-cni && \
      ./build
 
 # build arm64 container
-FROM centos:7
+FROM arm64v8/centos:7
 
 WORKDIR /
```

```
[tohayash@tohayash-srv multus-cni-amd64]$ docker run --rm --privileged --entrypoint=bash -it multus-cni:arm64
[root@59d232392886 /]# yum install file
Loaded plugins: fastestmirror, ovl
(snip)
[root@59d232392886 /]# file /bin/bash
/bin/bash: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 3.7.0, BuildID[sha1]=8221316a32fc963e8c014e2b4f37175621415b3b, stripped
[root@59d232392886 /]# file /usr/src/multus-cni/bin/multus 
/usr/src/multus-cni/bin/multus: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, not stripped
```

TrevorTaoARM:
@s1061123 Thank you for your reply.
For the issue you mentioned, I would like to add that the base image centos:7 is multi-arched, so when it is built on arm64 platform, it would just pull the arm64 version instead of amd64 version. So this Dockerfile is correct when it's built both on arm64 and amd64. Actually it is a mult-arch capable version which could be used to work for both amd64 and arm64.
If you still think it's necessary to show the difference in the base image of the Dockerfile for arm64 from amd64, I would change it accordingly as you said. Nice to get your idea here.

TrevorTaoARM:
@s1061123 Any further comments? Thank you!

TrevorTaoARM:
@s1061123 I abandoned this PR. 
Create a new PR about it at: https://github.com/intel/multus-cni/pull/457

PR #433, opened 2020-01-27
Support guid

Mmduh-483:


types/conf.go
```diff
@@ -181,6 +187,9 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 		if len(rc.Mac) != 0 {
 			capabilityArgs["mac"] = rc.Mac
 		}
+		if len(rc.GUID) != 0 {
+			capabilityArgs["guid"] = rc.GUID
```
adrianchiris, line 191: So, according to [1] the name is infiniband_guid

https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#well-known-capabilities

PR #436, opened 2020-02-06
[WIP] Simple e2e test with kubernetes-in-docker (KIND)

dougbtv:
Work in progress! Don't merge until ready.

Attempts to create a basic end-to-end test using kubernetes-in-docker.

Current status: Have KIND installed, trying to create wait loop to properly wait for Multus daemonset to come up before continuing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,31 @@ before_install:
 
 install:
   - go get -u golang.org/x/lint/golint
+  # Download kind & deps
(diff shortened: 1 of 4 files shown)


s1061123:
@dougbtv, thank you for your work!

I also plays with kind as following log. We can create eth1 in worker node by koko hence
we can create centos w/ macvlan and check ping test!

Please see the following log:
https://gist.github.com/s1061123/c0b857ec1a399c1e174531c0b826a81c

PR #439, opened 2020-02-10
[bugfix] Fixes unnecessary wait when readinessindicatorfile is not present

dougbtv:
Would wait every single time without the file being present.


s1061123:
Let's wait travis and then merged.

PR #440, opened 2020-02-12
[bugfix] Fixes unnecessary wait when readinessindicatorfile is not present

dougbtv:
Would wait every single time without the file being present.


dougbtv:
Replaces #439 (in order to run github actions)

s1061123:
@dougbtv (new) CI is passed, so let's merge it!

Issue #413, opened 2019-11-15
Multus pod fails to come up and "Topology Affinity Error" status is shown


pperiyasamy:
**What happend**:

We have two k8s worker nodes and topology manager policies enabled in each worker node.

```
worker - 1 = best-effort
worker - 2 = single-numa-node
```

When multus daemon set is installed with this [command](https://github.com/intel/multus-cni#quickstart-installation-guide), then multus pod doesn't come up on the worker - 2 node and finally fails with "Topology Affinity Error".

```
root@dl380-006-ECCD-SUT:~/cnis/multus-cni# kubectl get pods -n kube-system | grep multus
kube-multus-ds-amd64-gzj6x                   0/1     Topology Affinity Error   0          2m42s
kube-multus-ds-amd64-jp45m                   1/1     Running                   0          19m
```

This happens even though node meets [cpu and memory criteria](https://github.com/intel/multus-cni/blob/master/images/multus-daemonset.yml#L152) on a single numa socket as provided in 
 the spec.

Here are the kubelet logs from workers.

_worker - 1_ 

```
Nov 15 10:34:28 dl380-006-ECCD-SUT kubelet[51160]: I1115 10:34:28.144422   51160 topology_manager.go:308] [topologymanager] Topology Admit Handler
Nov 15 10:34:28 dl380-006-ECCD-SUT kubelet[51160]: I1115 10:34:28.144444   51160 topology_manager.go:317] [topologymanager] Pod QoS Level: Guaranteed
Nov 15 10:34:28 dl380-006-ECCD-SUT kubelet[51160]: I1115 10:34:28.144487   51160 topology_manager.go:194] [topologymanager] Hint Provider has no preference for NUMA affinity with any resource
Nov 15 10:34:28 dl380-006-ECCD-SUT kubelet[51160]: I1115 10:34:28.144499   51160 topology_manager.go:194] [topologymanager] Hint Provider has no preference for NUMA affinity with any resource
Nov 15 10:34:28 dl380-006-ECCD-SUT kubelet[51160]: I1115 10:34:28.144506   51160 topology_manager.go:285] [topologymanager] ContainerTopologyHint: {0000000000000000000000000000000000000000000000000000000000000011 true}
[...]

PR #449, opened 2020-03-06
Multus daemonset yaml for k8s 1.16

pperiyasamy:
The multus daemonset configuration uses Guaranteed QoS class
for the container, but this won't work if worker node is
configured with 'single-numa-node' topology management policy
when k8s version is 1.16. It is a k8s topology manager issue
and got solved in k8s 1.17.

This fix adds k8s 1.16 daemonset yaml configuration without
having Guaranteed QoS class for the container.

Issue #413

Signed-off-by: Periyasamy Palanisamy <periyasamy.palanisamy@est.tech>

### DIFF
--- a/images/multus-daemonset-1.16.yml
+++ b/images/multus-daemonset-1.16.yml
@@ -1,0 +1,245 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: network-attachment-definitions.k8s.cni.cncf.io
+spec:
+  group: k8s.cni.cncf.io
+  scope: Namespaced
+  names:
+    plural: network-attachment-definitions
+    singular: network-attachment-definition
+    kind: NetworkAttachmentDefinition
+    shortNames:
+    - net-attach-def
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          type: object
+          properties:
+            spec:
+              type: object
+              properties:
+                config:
+                  type: string
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: multus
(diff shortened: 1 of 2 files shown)


pperiyasamy:
can we merge this ?

dougbtv:
@pperiyasamy could you make the changes to the existing daemonset files? Changes to the limits should be OK -- just use the files named like "pre-1.16". Thank you!

pperiyasamy:
> could you make the changes to the existing daemonset files?

@dougbtv I don't think so. The `apiVersion `is changed from `v1beta1 `to `v1 `in k8s 1.16, so i just made it with new files.

s1061123:
For 1.16, we already have following yaml file for install.
https://github.com/intel/multus-cni/blob/master/images/multus-daemonset.yml

Our daemonset is for 'usual kubernetes deployment', not for all kubernetes deployment (e.g. change container runtime to something else). If you need to customize it, you can change it in your yaml (as the other project does) in your repo. 
As far as I know, topology manager in k8s is alpha at 1.16, so it should be kinds of 'customized install', not for usual deployment.

In addition, Kubernetes 1.16 will be no longer supported in community at next release, hence I don't know why we need to support 1.16. Could you please explain why we need to support 1.16 for a while even though Kubernetes is going to stop the support?

pperiyasamy:
> Current multus-daemonset.yaml could be used as following for your cases:
> `$ curl https://raw.githubusercontent.com/intel/multus-cni/master/images/multus-daemonset.yml | sed -e '/cpu: "100m"/d' | kubectl create -f -`
> 
> In addition, Kubernetes 1.16 will be no longer supported in community at next release, hence I don't know why we need to support 1.16. Could you please explain why we need to support 1.16 for a while even though Kubernetes is going to stop the support?

Okay. Closing this PR as it's no more valid.

PR #467, opened 2020-03-26
Use latest image version 3.4.1 in daemonset file

Mmduh-483:


Mmduh-483:
@s1061123 @dougbtv Can you please review?

PR #511, opened 2020-05-28
Change the error handling for kubernetes client

### LINKED ISSUE #492: Multus bringing up a Pod with incorrect number of interfaces
Had a Pod with following annotations:
    k8s.v1.cni.cncf.io/networks: '[ { "interface": "eth1", "namespace": "t001-u000003",
      "name": "test1-pktgen-21-net1" } ]'
    v1.multus-cni.io/default-network: '[ { "interface": "eth0", "namespace": "kube-system",
      "name": "calico" } ]'

Pod came up with only one interface using Calico CNI plugin. There are no annotations to the Pod containing the "network-status".

Looked at the logs and the source code noticed the following error. Multus does not return a failure, instead, I think it assigns interface from the default network. This does not seem right, it brings up a Pod with incorrect configuration. The best thing would be to fail the CNI.

Apr 26 07:50:42 qct-03 kubelet: 2020-04-26T07:50:42-07:00 [debug] TryLoadPodDelegates: Err in loading K8s cluster default network from pod annotation: rpc error:      code = Unavailable desc = etcdserver: leader changed, use cached delegates

 Ignoring this error can be hard to track when 100s of Pods are deployed.

s1061123:
This change adds retry to wait a bit if kubernetes client
returns service unavailable.

This fixes #492

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -37,13 +37,13 @@ import (
 	cnicurrent "github.com/containernetworking/cni/pkg/types/current"
 	cniversion "github.com/containernetworking/cni/pkg/version"
 	"github.com/containernetworking/plugins/pkg/ns"
+	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
+	"github.com/vishvananda/netlink"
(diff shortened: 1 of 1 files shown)


s1061123:
Not actually correct. Without the fix, multus ignored error only if we cannot get the pod information. Other error case captured by multus.

In #492 case, multus cannot get pod information due to Kubernetes server error, so it is kinds of corner case.

PR #522, opened 2020-06-11
Fix/del error fix

### LINKED ISSUE #519: fix:511 is not compete,  how about the cmdDel part?
**What happend**:
pod could be nil in cmdDel..

https://github.com/intel/multus-cni/commit/fff8e490b964daf686de5cca9d421562e475af19

same reason with fix in cmdAdd part,  in cmdDel,  if getpod not working due to kclient acces,
pod is nil. then multus.go will panic, right?
I think just add the code which #511 introduced to cmdDel()
**Environment**:

- Multus version 
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`):
- Primary CNI for Kubernetes cluster:
- OS (e.g. from /etc/os-release):
- File of '/etc/cni/net.d/'

s1061123:
fix #519

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -701,11 +701,33 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 		}
 	}
 
+	kubeClient, err = k8s.GetK8sClient(in.Kubeconfig, kubeClient)
+	if err != nil {
+		return cmdErr(nil, "error getting k8s client: %v", err)
+	}
+
 	pod := (*v1.Pod)(nil)
 	if kubeClient != nil {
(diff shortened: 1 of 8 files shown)


multus/multus.go
```diff
@@ -705,7 +705,24 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 	if kubeClient != nil {
```
amorenoz, line 705: Hi @s1061123 , I'm also hitting this issue.
Note that probably kubeClient  == nil
There is no call to k8s.GetK8sClient(in.Kubeconfig, kubeClient) in cmdDel. Maybe it should be added?

s1061123: Thank you for your comment! Yeah, I need to get kubeClient in cmdDel as well. I changed it.

PR #524, opened 2020-06-11
Test/error on fail

nicklesimba:
Added unit test for error-on-fail commit

### DIFF
--- a/multus/multus_test.go
+++ b/multus/multus_test.go
@@ -1459,7 +1459,7 @@ var _ = Describe("multus operations cniVersion 0.2.0 config", func() {
 
 		os.Setenv("CNI_COMMAND", "ADD")
 		os.Setenv("CNI_IFNAME", "eth0")
-		result, err := cmdAdd(args, fExec, clientInfo)
+		result, err := cmdAdd(args, fExec, *clientInfo)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(fExec.addIndex).To(Equal(len(fExec.plugins)))
 		r := result.(*current.Result)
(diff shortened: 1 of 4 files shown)


multus/multus_test.go
```diff
@@ -68,6 +74,56 @@ type fakeExec struct {
...
+
+// AddPod adds pod into kubernetes
+func (c *FakeKubeClient) AddPod(pod *v1.Pod) (*v1.Pod, error) {
+	return c.Client.Core().Pods(pod.ObjectMeta.Namespace).Create(pod)
+}
+
+// GetPod gets pod from kubernetes
+func (c *FakeKubeClient) GetPod(namespace, name string) (*v1.Pod, error) {
+	pod, _ := c.Client.Core().Pods(namespace).Get(name, metav1.GetOptions{})
+	return pod, errors.NewServiceUnavailable("Testing purposes")
```
s1061123, line 96: This code always return error. How about to have a function which sets error code to be returned, such like that:
`func (c *FakeKubeClient) SetError(err *StatusError)`

Of corse, if we set `SetError(nil)` then this client returns no error (i.e. succeed to get pod).

multus/multus_test.go
```diff
@@ -1392,16 +1450,16 @@ var _ = Describe("multus operations cniVersion 0.2.0 config", func() {
 		}, nil)
 
 		clientInfo := NewFakeClientInfo()
-		_, err := clientInfo.Client.Core().Pods(fakePod.ObjectMeta.Namespace).Create(fakePod)
+		_, err := (*clientInfo).GetKubeClient().Core().Pods(fakePod.ObjectMeta.Namespace).Create(fakePod)
```
s1061123, line 1453: (*clientInfo) could be clientInfo here and followings.

PR #527, opened 2020-06-16
Add support for log rotation

shahar-klein:
When using debug level the log files tends to fill up quickly.
Add support for log rotation using lumberjack, simple and easy to use
log rotation module.

Signed-off-by: Shahar Klein <sklein@nvidia.com>


s1061123:
First, thank you so much for your pull request!

1.  Currently multus keeps `vendor` directory to build multus without go modules. Could you do `go mod vendor` to update vendor package?
2. Could you please change `loggingFp` in `logging/logging_test.go` file?

shahar-klein:
> First, thank you so much for your pull request!
> Almost OK to me, but let me add minor comment as following:
> 
Thank you for your comments - I fixed and pushed

logging/logging.go
```diff
@@ -139,16 +141,22 @@ func SetLogFile(filename string) {
...
-		loggingFp = nil
+	w := &lumberjack.Logger{
+                        Filename:   filename,
+                        MaxSize:    100, // megabytes
+                        MaxBackups: 5,
+                        MaxAge:     5, // days
+                        Compress:   true,
+                    }
+	if w == nil {
 		fmt.Fprintf(os.Stderr, "multus logging: cannot open %s", filename)
```
girishmg, line 152: @shahar-klein why would `w` be nil here?

shahar-klein: Thanks for the comment - fixed.

s1061123:
LGTM, too! thank you for your help! Merged.

PR #534, opened 2020-07-08
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This `date` call only works with GNU date (probably not a problem since kubernes is pretty Linux-specific).

Also use UTC to be independent of timezone.

### DIFF
--- a/build
+++ b/build
@@ -19,7 +19,7 @@ if [ -z "$VERSION" ]; then
 	fi
 	set -e
 fi
-DATE=$(date --iso-8601=seconds)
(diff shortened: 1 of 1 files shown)


dougbtv:
Thanks for the contribution and the links to describe your intention, appreciate it!

Issue #544, opened 2020-08-05
Capabilities from underlying CNI plugins absent from 00-multus.conf


clementnuss:
**What happend**:
The `portMapping` [capability](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#dynamic-plugin-specific-fields-capabilities--runtime-configuration), present in an underlying CNI plugin, was not in the `00-multus.conf` file (which is automatically generated by the [entrypoint.sh](https://github.com/intel/multus-cni/blob/master/images/entrypoint.sh) script.
As a consequence, `runtimeConfig` related to the portMappings capability was not sent from the runtime to the Multus (and underlying) CNI, so that underlying plugin would not receive this info and `portMapping` would not happen.

**What you expected to happen**:
I expected either the container runtime or Multus to look for nested capabilities in the `delegates` config of `00-multus.conf`, so that the capabilities-based `runtimeConfig` information would be sent to Multus. However the runtime, (at least `dockershim and Kubernetes) only considers the first valid CNI plugin from `/etc/cni/net.d/`, and it does not look at the `delegates` key from this file, thereby ignoring any nested plugin capability.

**How to reproduce it (as minimally and precisely as possible)**:
Install Multus with underlying plugins Calico and [`portmap`](https://github.com/containernetworking/plugins/tree/master/plugins/meta/portmap), declare the `portMapping` capability in the CNI plugin configuration, and start Multus in `conf-file=auto` mode: the generated `00-multus.conf` file won't have any capability in its top-level CNI plugin information.

Here is what a capability declaration looks like, and what should be present in the top-level CNI config file:
```
{
  "type": "portmap",
  "capabilities": {"portMappings": true},
}
```

**Anything else we need to know?**:
There are many more capabilities that can be used: although `portMapping` is [not recommended by Kubernetes team](https://kubernetes.io/docs/concepts/configuration/overview/#services), it can be useful, and many more capabilities-based plugin exist that would benefit from pushing nested capabilities in the toplevel CNI configuration file.

---

**Environment**:
- Multus version: **hub.docker.com/nfvpe/multus:v3.4.2**
- Kubernetes version (use `kubectl version`): **1.18.3**
- Primary CNI for Kubernetes cluster: **Calico**
- OS (e.g. from /etc/os-release): **CentOS Linux, Version 8**
- File of '/etc/cni/net.d/': 
[00-multus.conf.txt](https://github.com/intel/multus-cni/files/5026101/00-multus.conf.txt)
[...]

PR #545, opened 2020-08-05
Pull underlying capabilities up to 00-multus.conf

clementnuss:
As discussed in issue #544 the nested plugin `capabilities` are not integrated in the `00-multus.conf` file generated by the `images/entrypoint.sh` file.
I implemented the parsing of the underlying master plugin json configuration file with python, to retrieve the nested capabilities and to integrate them in the top-level section of the `00-multus.conf` file (because it is not sufficient to have capabilities declared in the `delegates` section of the CNI config file.)

The python script for analyzing underlying plugins capabilities is compatible with Python 2 and Python 3, it could be a bit shorter if I used only Python 3 syntax, but I noticed that the Dockerfile is based on a CentOS version that uses Phyton 2.
When no capabilities in the underlying plugins, then nothing gets added to the `00-multus.conf` file.

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -302,6 +302,24 @@ if [ "$MULTUS_CONF_FILE" == "auto" ]; then
         MASTER_PLUGIN_NET_NAME="multus-cni-network"
       fi
 
+      capabilities_python_filter_tmpfile=$(mktemp)
+      cat << EOF > $capabilities_python_filter_tmpfile
+import json,sys
+conf = json.load(sys.stdin)
+capabilities = {}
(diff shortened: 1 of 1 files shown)


clementnuss:
Sorry for the multiple `push`: I introduced a modification that improves the parsing of the capabilities, and first commited a `vim` file instead of my modification, and when I force pushed a new commit with the correct file I forgot to put a title for the commit 🤦‍♂️😅

Everything should be OK now.

s1061123:
Thank you for your PR, @clementnuss !

It looks good to me. Let's wait a few days for other comments and merge it.

clementnuss:
Thanks for merging!
Any why the build failed ? the same commit passed all tests successfully yesterday 🤔

PR #547, opened 2020-08-09
📝 Docs: ReadME Updated(badges and header aligned)

pawarashish564:
Badges and header are now aligned with the Logo image in the center.
> If you have any thoughts on this please let me know.


dougbtv:
We'd prefer to not have the HTML in the markdown, and go with the github default alignment. Thanks regardless for your contribution.

PR #548, opened 2020-08-09
Use latest image version 3.6 in daemonset files

Mmduh-483:


Mmduh-483:
@s1061123 @dougbtv Can you please review?

Mmduh-483:
@s1061123 Any comments here?

PR #556, opened 2020-09-14
Add e2e tests for default-route annotation

s1061123:


dougbtv (approved):
LGTM! ...This was even better than I was thinking, thanks for the new e2e test!

PR #571, opened 2020-10-23
Bump kubernetes dependencies to v0.18.3

Billy99:


Billy99:
@dougbtv PTAL

dougbtv:
Thanks for the PR Billy, appreciate it! Looks good.

Just an FYI for anyone else dropping by, this is necessary in general to catch up with the k8s libs, but, specifically for upcoming improvements for the device id enhancement to the NPWG spec.

PR #574, opened 2020-10-29
Allows specifying "global namespaces" when using namespace isolation

### LINKED ISSUE #567: Namespace Isolation: Allowing more than just the default namespace
When [namespace isolation](https://github.com/intel/multus-cni/blob/master/doc/configuration.md#namespace-isolation) is enabled, it prevents you from referring to net-attach-defs in namespaces outside of the current namespace of a pod. This is enabled by default in OpenShift.

e.g., If you have a pod in namespace "foo" -- you can only refer to net-attach-defs that are created in foo, so you can't refer to a net-attach-def in namespace `bar`, with an annotation `bar/some-net-attach-def`.

The exception is the default namespace, you can refer to a net-attach-def in `default` namespace from any namespace, e.g. `default/macvlan-conf`.

However, this scenario doesn't exactly work perfectly for operators, where they'd like to specifically refer to net-attach-defs in namespaces other than `default` for organizational purposes.

How can we enable further namespaces in addition to default? Possibly add a configuration option `--namespace-isolation-global-namespaces=default,openshift-multus,foo,bar,quux`?

dougbtv:
When using `namespaceIsolation` -- we have only allowed for referring to the `default` namespace from any given namespace. This enhancement adds a configuration option `globalNamespaces` which allows one to specify any additional namespaces which they would like to refer to "globally" -- that is, from any given namespace (as opposed to "locally" -- from the same namespace in which a pod resides, which is the core functionality of `namespaceIsolation`).

The `globalNamespaces` parameter works only when `namespaceIsolation` is set to true. When `globalNamespaces` is unset, it is implied that `default` will still be allowed globally.

cc: @rohan47

Fixes #567

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -119,6 +119,8 @@ The functionality provided by the `namespaceIsolation` configuration option enab
 
 **NOTE**: The default namespace is special in this scenario. Even with namespace isolation enabled, any pod, in any namespace is allowed to refer to `NetworkAttachmentDefinitions` in the default namespace. This allows you to create commonly used unprivileged `NetworkAttachmentDefinitions` without having to put them in all namespaces. For example, if you had a `NetworkAttachmentDefinition` named `foo` the default namespace, you may reference it in an annotation with: `default/foo`.
 
+**NOTE**: You can also add additional namespaces which can be referred to globally using the `global-namespaces` option (see next section).
+
 For example, if a pod is created in the namespace called `development`, Multus will not allow networks to be attached when defined by custom resources created in a different namespace, say in the `default` network.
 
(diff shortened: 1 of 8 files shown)


rohan47:
@dougbtv If I understand correct, `globalNamespaces` needs to be specified in the multus config and can't be specified through `net-attach-defs` right?

dougbtv:
@matrohon that's correct -- so in the case of OpenShift, we'll specify the `globalNamespaces` using the cluster-network-operator. 

I want to allow some flexibility of configuration to allow using some namespaces globally, but, if we allowed it in net-attach-defs then it opens the door for unprivileged users to be able to specify it (the namespace isolation feature is intended to disallow some class of user from accessing networks that are considered "more privileged").

matrohon:
you probably meant @rohan47 ?

dougbtv:
(sorry for the bad autocomplete Mathieu!) Yes, @rohan47 above ^^^ :)

s1061123 (approved):
LGTM, thanks!

PR #577, opened 2020-11-04
Add DeviceInfo Data to NetworkStatus Annotation

Billy99:
Update the NetworkStatus with DeviceInfo per the approved device-info-spec, which can be found:
https://github.com/k8snetworkplumbingwg/device-info-spec

This allows DevicePlugins and CNIs to add additional data to the NetworkStatus annotation, which can be used by a container to add in deployment of the container workload.

This PR is based on [npwg_poc2](https://github.com/amorenoz/multus-cni/commits/npwg_poc2) by Adrian Moreno


Billy99:
/hold

This PR requires https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/pull/29 to be merged first. Once that is merged, go.mod will be updated to point to those changes,

multus/multus.go
```diff
@@ -558,9 +585,16 @@ func cmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 	cniArgs := os.Getenv("CNI_ARGS")
 	for idx, delegate := range n.Delegates {
 		ifName := getIfname(delegate, args.IfName, idx)
+		rt, cniDeviceInfoPath := types.CreateCNIRuntimeConf(args, k8sArgs, ifName, n.RuntimeConfig, delegate)
+		if cniDeviceInfoPath != "" {
+			err = nadutils.CopyDeviceInfoForCNIFromDP(cniDeviceInfoPath, delegate.ResourceName, delegate.DeviceID)
```
Billy99, line 590: @amorenoz There are several places in the PR where I ignore the errors coming back from utility calls for the DP and CNI files. Is there anyway to know for sure if a file should exist and take extra action? Or leave as is and handle file if it exists and move on otherwise.

types/conf.go
```diff
@@ -122,23 +126,30 @@ func LoadDelegateNetConf(bytes []byte, net *NetworkSelectionElement, deviceID st
...
 	delegateConf.Bytes = bytes
 
 	return delegateConf, nil
 }
 
-// MergeCNIRuntimeConfig creates CNI runtimeconfig from delegate
-func MergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetConf) *RuntimeConfig {
-	logging.Debugf("MergeCNIRuntimeConfig: %v %v", runtimeConfig, delegate)
+// mergeCNIRuntimeConfig creates CNI runtimeconfig from delegate
+func mergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetConf) *RuntimeConfig {
```
dougbtv, line 144: Just making sure I understand this, but, you're making this a private function in favor of the `types.CreateCNIRuntimeConf` where it's called from multus.go?

Billy99: Yes making private. All cases in multus.go that `mergeCNIRuntimeConfig` was called, it was followed by a call to `CreateCNIRuntimeConf`. So `mergeCNIRuntimeConfig` is now called from `CreateCNIRuntimeConf` and no longer called from multus.go.

multus/multus.go
amorenoz: The thing is, everything is optional for now right?
From Multul's point of view:
- If DP file does not exist, we may want to log, but that 's not an error
- If CNI does not create a file, we may want to log, but is not an error either.

If DP does exist, we do the copy and then it is magically removed after the delegation-chain call (i.e: when we write the NetworkStatus Annotation), then yes, that's an error.

types/conf.go
```diff
@@ -164,18 +168,26 @@ func mergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetCo
 		if delegate.DeviceID != "" {
 			runtimeConfig.DeviceID = delegate.DeviceID
 		}
+		logging.Debugf("mergeCNIRuntimeConfig: add runtimeConfig for net-attach-def: %v", runtimeConfig)
 	}
 
 	return runtimeConfig
 }
 
 // CreateCNIRuntimeConf create CNI RuntimeConf for a delegate
```
amorenoz, line 177: This function has changed, maybe update the comment?

types/types.go
```diff
@@ -111,6 +112,8 @@ type DelegateNetConf struct {
 	ConfListPlugin bool `json:"-"`
 	// DeviceID is only used internal housekeeping
 	DeviceID string `json:"deviceID,omitempty"`
+	// ResourceName is only used internal housekeeping
+	ResourceName string `json:"resourceName,omitempty"`
```
amorenoz, line 106: Why do we need this for?
It may be there from my old poc code in which I used the ResourceName to derive the CNIDeviceInfoFile.
But now we're using the RuntimeConfig to store the entire CNIDeviceInfoFile so this might not be needed

k8sclient/k8sclient.go
```diff
@@ -285,7 +285,7 @@ func getKubernetesDelegate(client *ClientInfo, net *types.NetworkSelectionElemen
 		return nil, resourceMap, err
 	}
 
-	delegate, err := types.LoadDelegateNetConf(configBytes, net, deviceID)
+	delegate, err := types.LoadDelegateNetConf(configBytes, net, deviceID, resourceName)
```
dougbtv, line 288: I must be missing something fundamental, but... Where is this set in advance? Wouldn't this be unassigned?

dougbtv: I'm reading this wrong. I looked at the whole file, ignore.

Billy99: Was just getting ready to comment that it is set above in this function:
`resourceName, ok := customResource.GetAnnotations()[resourceNameAnnot]`

types/types.go
dougbtv: It looks like Billy needs to pass this between two places and he's stashed it here in the DelegateNetConf. 

I'm personally OK with this. However, if you have a recommendation for a way to extract it otherwise, I'm definitely all ears.

Billy99: ResourceName is used in multus.go in `cmdAdd()` to call `nadutils.CopyDeviceInfoForCNIFromDP()`, used to determine the DP file name. There may be a better way to get the data from k8sclient.go in `getKubernetesDelegate()` where it is pulled from the annotation from NetworkAttachmentDefinition, but I haven't found it.

multus/multus.go
```diff
@@ -100,6 +100,22 @@ func getIfname(delegate *types.DelegateNetConf, argif string, idx int) string {
...
+	// If the DPDeviceInfoFile was not created, CNI might have created it. So
+	// either way, load CNIDeviceInfoFile.
+	if info, ok := runtimeConf.CapabilityArgs["CNIDeviceInfoFile"]; ok {
+		logging.Errorf("getDelegateDeviceInfo(): CapArgs - info=%v ok=%v", info, ok)
+		if infostr, ok := info.(string); ok {
+			logging.Errorf("getDelegateDeviceInfo(): infostr=%s", infostr)
+			return nadutils.LoadDeviceInfoFromCNI(infostr)
+		}
+	} else {
+		logging.Errorf("getDelegateDeviceInfo(): No CapArgs - info=%v ok=%v", info, ok)
```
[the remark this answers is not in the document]
Billy99: There are several log calls that should be there. Left over from some debugging. Next commit will clean them up.

types/conf.go
Billy99: Tweaked the comment. Let me know if you think there should be more.

types/conf.go
```diff
@@ -202,6 +212,9 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 		if len(delegateRc.InfinibandGUID) != 0 {
 			capabilityArgs["infinibandGUID"] = delegateRc.InfinibandGUID
 		}
+		if delegateRc.DeviceID != "" {
```
zshi-redhat, line 227: Does this mean we need to add `DeviceID` capability explicitly when creating net-attach-def?

types/conf.go
```diff
@@ -122,6 +122,13 @@ func LoadDelegateNetConf(bytes []byte, net *NetworkSelectionElement, deviceID st
 		if net.InfinibandGUIDRequest != "" {
 			delegateConf.InfinibandGUIDRequest = net.InfinibandGUIDRequest
 		}
+		if net.DeviceID != "" {
+			if deviceID != "" {
+				logging.Debugf("Warning: Both RuntimeConfig and ResourceMap provide deviceID. Ignoring RuntimeConfig")
```
zshi-redhat, line 131: Do you see RuntimeConfig.DeviceID is to replace ResourceMap.DeviceID as default in the future?

dougbtv (approved):
LGTM! Thanks Billy -- I realize this was a long road. Thanks for all the attention between both the spec, as well as this PR, and all the due dilligence.

types/conf.go
amorenoz: @zshi-redhat AFAIK, I don't think so. RuntimeConfig.DeviceID is useful for cases in which you don't have Device Plugin (hence, no ResourceMap.DeviceID)

types/conf.go
amorenoz: Only if the CNI supports deviceIDs not coming from DevicePlugin, e.g: host-device

PR #586, opened 2020-11-24
Multus should exit zero on DEL when cache file is missing and pod cannot be found.

dougbtv:
Otherwise, this can cause a pod to not be fully deleted, its sandbox may remain as the DEL continues to be retried due to exiting non-zero every time.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1900835

### DIFF
--- a/multus/multus.go
+++ b/multus/multus.go
@@ -758,7 +758,10 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 				logging.Errorf("Multus: failed to get delegates: %v, but continue to delete clusterNetwork", err)
 			}
 		} else {
-			return cmdErr(k8sArgs, "error reading the delegates: %v", err)
+			// The options to continue with a delete have been exhausted (cachefile + API query didn't work)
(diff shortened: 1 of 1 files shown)


multus/multus.go
```diff
@@ -758,7 +758,9 @@ func cmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 				logging.Errorf("Multus: failed to get delegates: %v, but continue to delete clusterNetwork", err)
 			}
 		} else {
-			return cmdErr(k8sArgs, "error reading the delegates: %v", err)
+			// The options to continue with a delete have been exhausted (cachefile + API query didn't work)
+			// We cannot exit with an error as this may cause a sandbox to never get deleted.
+			logging.Errorf("Multus: failed to get the cached delegates file: %v, cannot properly delete", err)
```
s1061123, line 763: How about just return after Errorf()? otherwise `return delPlugins(...)` is invoked and it may also return error...

dougbtv: +1, thanks for that, agreed. Have an update :+1:

s1061123 (approved):
LGTM, thanks!

PR #591, opened 2020-12-11
Restructured for Go standard project layout

dougbtv:
I've got Nikhil's work rebased and refactored to account for changes on master.

Currently, it looks like it's failing the e2e-kind test. I'm working on replicating the issue in my own environment to evaluate.


dougbtv:
Looks like the binary it's producing is a problem...

```
./bin/multus 
./bin/multus: line 1: syntax error near unexpected token `newline'
./bin/multus: line 1: `!<arch>'
```

dougbtv:
Looks like the build was referencing building something that was not `main` so I cleared that up, I believe.

However, it's failing during a `docker build` for the `Dockerfile.openshift`, with a weird error on copy. Nikhil had reported this, and I can't replicate locally. Needs some more digging.

dougbtv:
For example, local run looks like: https://pasteall.org/Nnok

hack/build-go.sh
```diff
@@ -40,7 +40,7 @@ if [ "$GO111MODULE" == "off" ]; then
 	export GO15VENDOREXPERIMENT=1
 	export GOBIN=${PWD}/bin
 	export GOPATH=${PWD}/gopath
-	go install -tags no_openssl -ldflags "${LDFLAGS}" "$@" ${REPO_PATH}/multus
+	go build -o ${PWD}/bin/multus -tags no_openssl -ldflags "${LDFLAGS}" "$@" ${REPO_PATH}/cmd
```
dougbtv, line 43: Here's what I needed to change that was causing the GitHub actions build of this to fail. It's kind of redundant with the `export GOBIN` above, but should wind up with the binary in the correct place regardless.

s1061123 (approved):
LGTM! Thanks!

PR #593, opened 2020-12-14
Cleanup multus config file 00-multus.conf when deleting daemonset

### LINKED ISSUE #592: Clean Up auto generated config
When multus daemonset is deployed with --multus-conf-file=auto, it generates an auto cnofig file in CNI_DIR, usually it is 00-multus. After removing the multus daemonset and roles, this file will stay which cause to call multus cni that doesn't have the permissions to access k8s api, that cause the multus to fail everytime.

Clean up the auto generated config file

**Why is this needed**:
Leaving config file for mulus without permission causing the cluster to not be able to schedule pods

Mmduh-483:
When multus daemonset is deleted the auto generated config file
00-multus.conf is left, when new pod created and uses that conf
cni file, it will fail because multus roles are removed

This patch cleanup multus conf file 00-multus to allow using previous
primary cni config file

Fixes #592

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -380,6 +380,11 @@ generateMultusConf
 
 # ---------------------- end Generate "00-multus.conf".
 
+# remove "00-multus.conf" on close.
+if [ "$MULTUS_CONF_FILE" == "auto" ]; then
+  trap "rm -f $CNI_CONF_DIR/00-multus.conf" EXIT
+fi
+
 # Enter either sleep loop, or watch loop...
 if [ "$MULTUS_CLEANUP_CONFIG_ON_EXIT" == true ]; then
   log "Entering watch loop..."


images/entrypoint.sh
```diff
@@ -380,6 +380,9 @@ generateMultusConf
 
 # ---------------------- end Generate "00-multus.conf".
 
+# remove "00-multus.conf" on close.
+trap "rm -f $CNI_CONF_DIR/00-multus.conf" EXIT
```
adrianchiris, line 384: IMO This should be delete only if it was created by the entrypoint.

i.e: `"$MULTUS_CONF_FILE" == "auto"`

also i wonder if it should depend on `MULTUS_CLEANUP_CONFIG_ON_EXIT`

dougbtv:
I've got to think on this one, but... I think it might need to be parameterized.

images/entrypoint.sh
```diff
@@ -380,6 +380,11 @@ generateMultusConf
 
 # ---------------------- end Generate "00-multus.conf".
 
+# remove "00-multus.conf" on close.
+if [ "$MULTUS_CONF_FILE" == "auto" ]; then
+  trap "rm -f $CNI_CONF_DIR/00-multus.conf" EXIT
+fi
+
 # Enter either sleep loop, or watch loop...
 if [ "$MULTUS_CLEANUP_CONFIG_ON_EXIT" == true ]; then
```
Mmduh-483, line 389: @dougbtv "--cleanup-config-on-exit" is used to set "MULTUS_CLEANUP_CONFIG_ON_EXIT" which keep tracking that the file exists, which is confusing, do u want to change this param to something else, then use it to clean config file 00-multus.conf

dougbtv:
Mmduh, currently we think that it might be a better route for these customizations to go into a custom build and/or a modified yaml file.

PR #595, opened 2020-12-14
Fix entrypoint fail to read capabilities in non chain plugin config

### LINKED ISSUE #594: Entrypoint fails in non chain plugin config
**What happend**:
Building and deploying image from master fails in the entrypoint

Traceback (most recent call last):
  File "/tmp/tmp.4lja25dJNn", line 4, in <module>
    for capa in [p['capabilities'] for p in conf['plugins'] if 'capabilities' in p]:
KeyError: 'plugins'

Generating multus config file fails in the python script to read capabilities

Build image from master then deploy it arg `--multus-conf-file=auto`

**Extra information**:
Primary config file;
```
{
  "name": "cbr0",
  "cniVersion": "0.3.1",
      "type": "flannel",
      "delegate": {
        "hairpinMode": true,
        "isDefaultGateway": true
      }
}

```

Mmduh-483:
entrypoint script fails with error when try reading capabilities in
non chain plugin config file when using "--multus-conf-file=auto"

Fixes #594

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -317,8 +317,11 @@ if [ "$MULTUS_CONF_FILE" == "auto" ]; then
 import json,sys
 conf = json.load(sys.stdin)
 capabilities = {}
-for capa in [p['capabilities'] for p in conf['plugins'] if 'capabilities' in p]:
-    capabilities.update({capability:enabled for (capability,enabled) in capa.items() if enabled})
+if 'plugins' in conf:
+    for capa in [p['capabilities'] for p in conf['plugins'] if 'capabilities' in p]:
+        capabilities.update({capability:enabled for (capability,enabled) in capa.items() if enabled})
+elif 'capabilities' in conf:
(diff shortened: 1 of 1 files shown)


images/entrypoint.sh
```diff
@@ -317,7 +317,7 @@ if [ "$MULTUS_CONF_FILE" == "auto" ]; then
 import json,sys
 conf = json.load(sys.stdin)
 capabilities = {}
-for capa in [p['capabilities'] for p in conf['plugins'] if 'capabilities' in p]:
+for capa in [p['capabilities'] for p in conf.get('plugins', []) if 'capabilities' in p]:
```
adrianchiris, line 320: So, to my understanding this bit of code is used to get the list of capabilities the master plugin supports.

Shouldn't we support the case where the master plugin is not specified as a configuration list ?

Mmduh-483: Updated to cover master plugin

adrianchiris (approved):

dougbtv:
@adrianchiris :100: Thanks for picking this one up!!

dougbtv (approved):

PR #600, opened 2021-01-07
Pods in daemonset should have grace period reduced.

dougbtv:
Otherwise, it can take a long time for the daemonset to stop, and there's no real reason it needs to hang out for a long time.

### DIFF
--- a/images/multus-daemonset-crio.yml
+++ b/images/multus-daemonset-crio.yml
@@ -199,6 +199,7 @@ spec:
           mountPath: /host/usr/libexec/cni
         - name: multus-cfg
           mountPath: /tmp/multus-conf
+      terminationGracePeriodSeconds: 10
       volumes:
         - name: run
           hostPath:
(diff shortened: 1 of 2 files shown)


s1061123:
Looks good to me!

dougbtv:
I think you're right Tomo, trapping SIGTERM doesn't buy a whole lot here.

However, there is an improvement with the graceperiodseconds set lower, before adding grace period seconds...

```
[centos@kube-singlehost-master multus-cni]$ time kubectl delete pod kube-multus-ds-amd64-hx582 -n kube-system
pod "kube-multus-ds-amd64-hx582" deleted

real	0m36.297s
```

After adding it...

```
[centos@kube-singlehost-master multus-cni]$ time kubectl delete pod kube-multus-ds-amd64-g5hzq -n kube-system
pod "kube-multus-ds-amd64-g5hzq" deleted

real	0m17.937s
```

s1061123:
Sorry for late comment but how about to change not only amd64 also ppc/arm?

dougbtv:
Thanks for the review and spotting the missing sections! Added to the missing sections.

s1061123 (approved):
LGTM!

PR #604, opened 2021-01-31
Add validateK8sApi flag

Mmduh-483:
When this flag is set it validates that multus can communicate with Kubernetes API and if failed to communicate it fails the cni. If set to false then skip failure when multus can't get pod data

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -22,6 +22,7 @@ RESTART_CRIO=false
 CRIO_RESTARTED_ONCE=false
 RENAME_SOURCE_CONFIG_FILE=false
 SKIP_BINARY_COPY=false
+VALIDATE_K8S_API=false
 
 # Give help text for parameters.
 function usage()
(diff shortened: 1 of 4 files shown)


dougbtv:
We decided we'll move forward, but with making this behavior configurable.

* types, and conf.go -- have the structures and parsing for config options
* docs/configuration.md -- which should be updated to describe the flag

Mmduh-483:
@dougbtv

images/entrypoint.sh
```diff
@@ -22,6 +22,7 @@ RESTART_CRIO=false
 CRIO_RESTARTED_ONCE=false
 RENAME_SOURCE_CONFIG_FILE=false
 SKIP_BINARY_COPY=false
+VALIDATE_K8S_API=false
```
s1061123, line 25: Why default is false?

s1061123:
Thank you so much for update PR, first.

[multi-net-spec of NPWG](https://github.com/k8snetworkplumbingwg/multi-net-spec)  defines that the plugin need to communicate to Kubernetes API and if cannot make it, plguin should cancel to create pod as error (see section 7.2 for details). So if we change 'skip api' is false as default, multus-cni does not following the [multi-net-spec of NPWG](https://github.com/k8snetworkplumbingwg/multi-net-spec).

I hope that multus-cni should following spec because multus-cni is the reference implementation of [multi-net-spec of NPWG](https://github.com/k8snetworkplumbingwg/multi-net-spec).

So the option should be opt-in configuration as following:

- option name could be `ignoreK8sAPI`(in json) / `--ignore-k8s-api` (in shell script) or something like that
- default behavior keeps as it is, 'fail (i.e. cancel to create pod) if multus cannot communicate k8s api' as spec mentioned

dougbtv:
Yeah, let's flip that logic so the default matches the spec. And I also like the name `ignoreK8sAPI`.

If this is a problem, we could also bring it back up at the NPWG.

Mmduh-483:
@s1061123 @dougbtv Thanks for your comments, after internal discussion we decided to go with another approach to handle the network operator use case

PR #606, opened 2021-02-05
Bump protobuf to v1.3.2

dougbtv:
Update gogo/protobuf to v1.3.2 to fix https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-3121. The related k8s fix is here: kubernetes/kubernetes#98477

~Also bumps libcni to v0.8.1~


s1061123:
Why the test failed???
https://github.com/intel/multus-cni/pull/606/checks?check_run_id=1841681814#step:10:78

go.mod
```diff
@@ -4,17 +4,17 @@ go 1.12
 
 require (
 	github.com/Microsoft/go-winio v0.4.14 // indirect
-	github.com/containernetworking/cni v0.7.1
+	github.com/containernetworking/cni v0.8.1
 	github.com/containernetworking/plugins v0.8.2
+	github.com/gogo/protobuf v1.3.2 // indirect
```
fedepaol, line 9: I think with this only we won't cover indirect deps.
If you look at go.sum, there are still older version used. To always use 1.3.2 you need to use the replace directive

s1061123: Yeah, right. As far as I checked k8s repo, master or v1.21.0-alpha.3 only fixes the protobuf isssue.

https://github.com/kubernetes/kubernetes/commit/4d3ed163a0ed45e8790801d74503fe300d05daa0

That is what I know of, for now.

- Currently multus uses kubernetes v1.13.0 library (which contains kubelet/api/podresources/v1alpha1)
- To update protobuf we need to update v1.21.0-alpha.3 or later (or wait for bugfix version...)
- v1.21.0-alpha.3 does not have kubelet/api/podresources/v1alpha1, so multus build is failed

s1061123: So we may need to stop support old sr-iov feature (kubelet)...

s1061123:
Looks good to me, thanks! Please merge it!

PR #607, opened 2021-02-09
conf: do not modify global runtimeconfig when merging

amorenoz:
When we call mergeRuntimeConfig, the global RuntimeConfig gets
overwritten with the result of the merging, thus affecting the
subsequent delegates.

Do not modify the global RuntimeConfig and instead make a copy
when merging it

Also, if a value has been provided for CNIDeviceInfoFile in the
delegate's runtimeconfig, overwrite it to avoid possible name
colissions.

Signed-off-by: Adrian Moreno <amorenoz@redhat.com>
Signed-off-by: Billy McFall <bmcfall@redhat.com>

### DIFF
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -145,35 +145,38 @@ func LoadDelegateNetConf(bytes []byte, net *NetworkSelectionElement, deviceID st
 // mergeCNIRuntimeConfig creates CNI runtimeconfig from delegate
 func mergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetConf) *RuntimeConfig {
 	logging.Debugf("mergeCNIRuntimeConfig: %v %v", runtimeConfig, delegate)
+	var mergedRuntimeConfig RuntimeConfig
+
 	if runtimeConfig == nil {
-		runtimeConfig = &RuntimeConfig{}
+		mergedRuntimeConfig = RuntimeConfig{}
+	} else {
+		mergedRuntimeConfig = *runtimeConfig
 	}
 
 	// multus inject RuntimeConfig only in case of non MasterPlugin.
 	if delegate.MasterPlugin != true {
-		logging.Debugf("mergeCNIRuntimeConfig: add runtimeConfig for net-attach-def: %v", runtimeConfig)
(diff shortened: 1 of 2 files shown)


amorenoz:
@Billy99 PTAL

Billy99 (approved):
/lgtm

pkg/types/conf.go
```diff
@@ -145,52 +145,58 @@ func LoadDelegateNetConf(bytes []byte, net *NetworkSelectionElement, deviceID st
...
-		runtimeConfig = &RuntimeConfig{}
+		mergedRuntimeConfig = RuntimeConfig{}
+	} else {
+		mergedRuntimeConfig = *runtimeConfig
 	}
 
 	// multus inject RuntimeConfig only in case of non MasterPlugin.
 	if delegate.MasterPlugin != true {
-		logging.Debugf("mergeCNIRuntimeConfig: add runtimeConfig for net-attach-def: %v", runtimeConfig)
+		logging.Debugf("mergeCNIRuntimeConfig: add mergedRuntimeConfig for net-attach-def: %v", mergedRuntimeConfig)
```
s1061123, line 158: I suppose we could keep debug message "mergeCNIRuntimeConfig: add runtimeConfig for net-attach-def: %v". Please keep variable name as mergedRuntimeConfig.

s1061123 (changes requested):
Thank you so much for the PR! I understand that we need the change.

BTW, could you please add unit-test in `pkg/types/conf_test.go` as well, to verify that mergeCNIRuntimeConfig does not overwrite runtimeConfig?

pkg/types/conf.go
amorenoz: Yep, thanks! My replacing magic went a bit too far :)

amorenoz: Done

amorenoz:
> Thank you so much for the PR! I understand that we need the change.
> 
> BTW, could you please add unit-test in `pkg/types/conf_test.go` as well, to verify that mergeCNIRuntimeConfig does not overwrite runtimeConfig?

@s1061123 Please check out the extra check in the unit-test.

s1061123 (approved):
LGTM, thank you for the changes!

PR #608, opened 2021-02-09
Adds a trap for SIGTERM in entrypoint script

dougbtv:
In my tests this improved the time to `kubectl delete pod $multus-daemonset-pod -n kube-system` by 5 seconds.

@s1061123 -- I had to use `wait` instead of `read` -- if I use `sleep $x & read` it waits for input before continuing, there is one spot where we sleep infinity, however, in a number of other spots we sleep inside a loop.

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -3,6 +3,13 @@
 # Always exit on errors.
 set -e
 
+# Trap sigterm
+function exitonsigterm() {
+  echo "Trapped sigterm, exiting."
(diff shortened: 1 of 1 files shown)


s1061123:
@dougbtv in the case of `read`, you just use `read`, without `sleep`. So we could use `read` at the last `sleep infinite`.

Other cases we still use `sleep <n> & wait `

s1061123:
LGTM, thanks!

PR #609, opened 2021-02-09
Change to sleep infinity & read and omit & wait on sleep n

dougbtv:


images/entrypoint.sh
```diff
@@ -397,22 +397,22 @@ if [ "$MULTUS_CLEANUP_CONFIG_ON_EXIT" == true ]; then
...
       generateMultusConf
       log "Continuing watch loop after configuration regeneration..."
     fi
-    sleep 1 & wait
+    sleep 1
   done
 else
   log "Entering sleep (success)..."
-  sleep infinity & wait
+  sleep infinity & read
```
s1061123, line 417: It shuld be `read` only, no sleep, no wait. `read` waits forever in this case.

s1061123 (changes requested):

s1061123 (approved):
LGTM, thanks!

PR #611, opened 2021-02-18
Add s390x support

yoheiueda:
This PR updates the GitHub Actions workflows to build multus
container images for the IBM Z architecture (s390x), and also
adds daemonsets for s390x nodes.

entrypoint.sh uses python3 when it is available, since centos images
are not available for s390x, and python2 is not installed by default
on debian-based images.

Signed-off-by: Yohei Ueda <yohei@jp.ibm.com>


dougbtv (approved):
LGTM! Thanks for the new architecture build and associated changes, appreciate it!

PR #616, opened 2021-02-24
Provide the option to not to log via STDERR

cathy-zhou:
Today, Multus will always log via STDERR, and these logs will then
logged by the Kubelet. If we also choose to have Multus log to a file by
setting the LogFile option in the CNI configuration, the same logs will
be logged twice.

This commit provide the option to disable logging to STDERR.

Signed-off-by: Yun Zhou <yunz@nvidia.com>

### DIFF
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -257,7 +257,8 @@ func GetGatewayFromResult(result *current.Result) []net.IP {
 
 // LoadNetConf converts inputs (i.e. stdin) to NetConf
 func LoadNetConf(bytes []byte) (*NetConf, error) {
-	netconf := &NetConf{}
+	// LogToStderr's default value set to true
+	netconf := &NetConf{LogToStderr: true}
 
 	logging.Debugf("LoadNetConf: %s", string(bytes))
 	if err := json.Unmarshal(bytes, netconf); err != nil {
@@ -265,6 +266,7 @@ func LoadNetConf(bytes []byte) (*NetConf, error) {
 	}
 
 	// Logging
+	logging.SetLogStderr(netconf.LogToStderr)
 	if netconf.LogFile != "" {
 		logging.SetLogFile(netconf.LogFile)
 	}
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -43,6 +43,7 @@ type NetConf struct {
 	DefaultNetworks []string                 `json:"defaultNetworks"`
 	LogFile         string                   `json:"logFile"`
 	LogLevel        string                   `json:"logLevel"`
+	LogToStderr     bool                     `json:"logToStderr,omitempty"`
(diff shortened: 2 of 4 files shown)


girishmg:
@dougbtv @zshi-redhat can you guys PTAL? with `debug` set, our `/va/log` directory is getting full since `multus` is very chatty and we log twice. Thanks.

pkg/types/types.go
```diff
@@ -43,6 +43,7 @@ type NetConf struct {
 	DefaultNetworks []string                 `json:"defaultNetworks"`
 	LogFile         string                   `json:"logFile"`
 	LogLevel        string                   `json:"logLevel"`
+	LogToStderr     bool                     `json:"logToStderr,omitempty"`
```
s1061123, line 46: Can we change `LogToStderr` type to `*bool` to keep the default behavior?

Following is the detailed reason. 

If we use 'bool' as a type of `LogToStderr` and no `LogToStderr` in CNI config, then `netconf.LogToStderr` become `false` because golang `omitempty` fills `false` as default and multus cannot identify whether the `false` is user specified or automatically filled as `omitempty`. As a result, multus behavior is changed and it is not expected.

If we use `*bool` then `netconf.LogToStderr` is nul if no LogToStderr in CNI config. Then we could handle it (i.e. set `true` if `nil`) and the behavior is kept.

pkg/types/conf.go
```diff
@@ -257,14 +257,16 @@ func GetGatewayFromResult(result *current.Result) []net.IP {
...
+	// LogToStderr's default value set to true
+	netconf := &NetConf{LogToStderr: true}
 
 	logging.Debugf("LoadNetConf: %s", string(bytes))
 	if err := json.Unmarshal(bytes, netconf); err != nil {
 		return nil, logging.Errorf("LoadNetConf: failed to load netconf: %v", err)
 	}
 
 	// Logging
+	logging.SetLogStderr(netconf.LogToStderr)
```
s1061123, line 269: Can we handle default value here (see 'types.go' review comment)?

s1061123 (changes requested):
Thank you for the PR, @cathy-zhou ! It seems to be reasonable feature and almost looks good to me.
I have one comment to keep the default behavior, so please take a look into it?

pkg/types/types.go
cathy-zhou: @s1061123 Thanks for your comments. 

The existing code already assures to keep the exisiting default behavior (that LogToStderr's default value is true if there is no logToStderr in CNI config). See code in https://github.com/intel/multus-cni/blob/f97f01f380788b89497515536dec42410d0e1124/pkg/types/conf.go#L261, I've also tested it out.

pkg/types/conf.go
cathy-zhou: See my replies above.

PR #617, opened 2021-02-24
Conditional check for tty/non-tty in case of wait (i.e. read)

### LINKED ISSUE #615: `read` command in entrypoint may not work without a TTY
I think having `read` in the entrypoint is causing the daemonset pods to crashloop, since there's not a tty:

Note how this will exit immediately:
```
docker run --entrypoint="" centos/tools /bin/bash -c 'read'
```

But if you give it a tty it waits:

```
docker run -t --entrypoint="" centos/tools /bin/bash -c 'read'
```

Compared to a control like so:

```
docker run --entrypoint="" centos/tools /bin/bash -c 'sleep 100000000000'
```

s1061123:
Fix #615.

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -418,5 +418,9 @@ if [ "$MULTUS_CLEANUP_CONFIG_ON_EXIT" == true ]; then
   done
 else
   log "Entering sleep (success)..."
-  read
+  if tty -s; then
+	  read
+  else
+	  sleep infinity
+  fi
 fi


s1061123:
Here's the sample output of `tty -s`

```
$ docker run --entrypoint="" centos/tools /bin/bash -c 'if tty -s; then echo "tty"; read; else echo "not TTY"; sleep 3; fi'
not TTY
$ docker run -t --entrypoint="" centos/tools /bin/bash -c 'if tty -s; then echo "tty"; read; else echo "not TTY"; sleep 3; fi'
tty
```

images/entrypoint.sh
```diff
@@ -418,5 +418,9 @@ if [ "$MULTUS_CLEANUP_CONFIG_ON_EXIT" == true ]; then
   done
 else
   log "Entering sleep (success)..."
-  read
+  if tty -s; then
+	  read
+  else
+	  sleep
```
cathy-zhou, line 424: should this be "sleep infinity"?

s1061123: right. will be changed. thanks!

dougbtv (approved):
LGTM!

PR #625, opened 2021-03-11
Fix link to an example pod with SRIOV network

e0ne:


dougbtv (approved):
LGTM! Thanks for the fix up :+1:

PR #633, opened 2021-03-22
Updates Readinessindicator error message to be actionable by users

dougbtv:


pkg/multus/multus.go
```diff
@@ -529,7 +529,7 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			return err == nil, nil
 		})
 		if err != nil {
-			return nil, cmdErr(k8sArgs, "PollImmediate error waiting for ReadinessIndicatorFile: %v", err)
+			return nil, cmdErr(k8sArgs, "Have you checked that your default network is ready? We're still waiting for the ReadinessIndicatorFile. PollImmediate error: %v", err)
```
s1061123, line 532: How about to add filename of indicater file in error message?

dougbtv: Thanks Tomo! Good call. I added that, I also changed the case to be all lower case.

s1061123:
Looks good to me! Let's merge that!

PR #636, opened 2021-03-25
Return error on annotation parsing error

### LINKED ISSUE #635: Multus does not raise a error on network annotation syntax error
**What happend**:
When the Pod network annotation is wrong (e.g. json syntax error), multus-cni does not raise error and proceed to build a pod without additional network interface

In above case, multus-cni should raise a error and cancel to build a pod.

Create a pod with wrong json network annotation such as:

```
---
apiVersion: v1
kind: Pod
metadata:
  name: centos-runtimeconfig
  annotations:
    k8s.v1.cni.cncf.io/networks: '[
            { "name": "centos-runtimeconfig-def",
              "ips": [ "10.1.1.104/24" ] 
    ]'
spec:
  containers:
  - name: centos-runtimeconfig
    image: docker.io/centos/tools:latest
    command:
    - /sbin/init
    securityContext:
      privileged: true
```

N/A

**Environment**:
snip because I'm working on fix.

s1061123:
Fix #635

### DIFF
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -367,7 +367,7 @@ func TryLoadPodDelegates(pod *v1.Pod, conf *types.NetConf, clientInfo *ClientInf
 		return len(delegates), clientInfo, nil
 	}
 
-	return 0, clientInfo, nil
+	return 0, clientInfo, err
 }
 
 // GetK8sClient gets client info from kubeconfig


dougbtv (approved):
LGTM!! Can't believe this was the case.

mumoshu:
@s1061123 @dougbtv FYI, I think I spotted a possible ssieu caused by this ☝️

PR #638, opened 2021-03-26
Change libcni's cache directory from default for multus delegates

s1061123:
This changes CNI cache directory to non default places to avoid conflicts among CNI runtimes.


dougbtv (approved):
LGTM!

PR #644, opened 2021-03-31
Skip to call device-info related function if not required

s1061123:
The case withot DeviceInfo/ResourceName, we could skip functions
related to device info. This change skips to call these functions
if delegates does not have DeviceInfo/ResourceName.


dougbtv (approved):
LGTM

PR #655, opened 2021-04-19
Correcting Dockerfile path for manual e2e tests

bjosv:
Setting up for manual e2e-tests fails due to that the multus container image is not built. This works fine in CI since this part is done separately within the github action.

This is a minor change to use the new location of the Dockerfile when running the e2e-tests manually.


s1061123:
LGTM! Thank you for your quick fix!

Issue #656, opened 2021-04-19
network name is empty under network-status annotation


amshankaran:
What happened:

Calico is configured as primary CNI. Multus uses this as delegate. Created a pod (via deployment). while describing the pod, under "networks-status:" annotation, network name is empty string "name": ""
master-0-ejlcluster> kubectl describe pod nginx-shr-6584fcbb55-lpndj
Name: nginx-shr-6584fcbb55-lpndj
Namespace: default
Priority: 0
Node: worker-pool1-w7m9sczq-ejlcluster/10.0.10.97
Start Time: Sat, 17 Apr 2021 09:30:08 +0000
Labels: app=nginx-shr
pod-template-hash=6584fcbb55
Annotations: k8s.v1.cni.cncf.io/network-status:
[{
"name": "",
"ips": [
"192.168.185.80"
],
"default": true,
"dns": {}
}]
k8s.v1.cni.cncf.io/networks-status:
[{
"name": "",
"ips": [
"192.168.185.80"
],
"default": true,
"dns": {}
}]

What you expected to happen:

under "networks-status:" annotation "name": "k8s-pod-network" is expected.

director-0> kubectl describe pod nginx-shr-6584fcbb55-5gqnf
Name: nginx-shr-6584fcbb55-5gqnf
Namespace: default
Priority: 0
Node: worker-p1-ywj6c6j5/10.0.10.17
Start Time: Fri, 16 Apr 2021 11:01:42 +0000
Labels: app=nginx-shr
pod-template-hash=6584fcbb55
Annotations: k8s.v1.cni.cncf.io/network-status:
[{
"name": "k8s-pod-network",
"ips": [
"192.168.248.74"
],
"default": true,
"dns": {}
}]
k8s.v1.cni.cncf.io/networks-status:
[{
"name": "k8s-pod-network",
"ips": [
"192.168.248.74"
],
"default": true,
"dns": {}
}]

How to reproduce it (as minimally and precisely as possible):
just deploy a nginx pod in latest kubernets cluster
kubectl create deployment nginx-shr --image=nginx

Anything else we need to know?:

master-0-ejlcluster> sudo vim 00-multus.conf

[...]

PR #657, opened 2021-04-19
Add cluster network name in pod network annotation

s1061123:
This addresses #656


e0ne (approved):
I'm not familiar good enough with a codebase but tests describes how it works and it fixes the issue

dougbtv (approved):
LGTM! Thanks Tomo

PR #672, opened 2021-05-21
Update vendor package

s1061123:


dougbtv (approved):
Thanks Tomo, LGTM. 

@aojea has also recommended these changes when he proposed this vendor bump. I've documented them in this issue @ https://github.com/k8snetworkplumbingwg/multus-cni/issues/673 which we can cover in a follow up.

PR #674, opened 2021-05-21
Remove global variables from kubletclient

### LINKED ISSUE #673: kubelet client refactoring: optimization and global variable
@aojea has recommended a few fixes that look to be solid improvements...

1. We're using [these global variables](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/kubeletclient/kubeletclient.go#L23-L26) which is a bad practice, additionally they're being mutated in the kubeletclient_test.go package, and there are some tests which may be running in such a way as to flake.

2. The https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/kubeletclient/kubeletclient.go#L44 `util` package is really only being used for `util.LocalEndpoint()` which is really just a couple methods that manipulate strings, found here @ https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cri/remote/util/util_unix.go#L82-L128 -- we could omit having to include the whole package and instead include these functions.

s1061123:
Fixes #673

### DIFF
--- a/pkg/kubeletclient/kubeletclient.go
+++ b/pkg/kubeletclient/kubeletclient.go
@@ -1,8 +1,8 @@
 package kubeletclient
 
 import (
+	"net/url"
 	"os"
-	"path/filepath"
 	"time"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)


dougbtv (approved):
LGTM! Thanks for straightening this up.

pkg/kubeletclient/kubeletclient.go
```diff
@@ -18,27 +17,26 @@ import (
...
 )
 
 // GetResourceClient returns an instance of ResourceClient interface initialized with Pod resource information
-func GetResourceClient() (types.ResourceClient, error) {
+func GetResourceClient(kubeletSocket string) (types.ResourceClient, error) {
 	// If Kubelet resource API endpoint exist use that by default
 	// Or else fallback with checkpoint file
-	if hasKubeletAPIEndpoint() {
+	if kubeletSocket == "" {
+		kubeletSocket = util.LocalEndpoint(defaultPodResourcesPath, podresources.Socket)
```
aojea, line 28: this returns a string with the scheme of the endpoint i.e
"unix:/pkg/kubelet.sock"
https://play.golang.org/p/zqf-bAWCnVt
does os.Stat() inside `hasKubeletAPIEndpoint()` detects it?

s1061123: oh, good catch! Thank you for your comment! will fix it.

aojea:
lgtm

e0ne (approved):

PR #676, opened 2021-05-23
Describe additional condition when using clusterNetwork

oshothebig:
This PR adds description about a condition that needs to be satisfied when configuring `clusterNetwork` to refer to a CNI JSON config file in `confDir`. I was trying to use `clsuterNetwork` config, but failed for a while as the file name set for `clusterNetwork` didn't match the value for `name` key in the file. It would be helpful to describe the condition currently undocumented.


s1061123 (commented):
It looks good to me. Thank you for your PR!

dougbtv (approved):
LGTM!

s1061123 (approved):

oshothebig:
Thanks for the quick review.

PR #680, opened 2021-05-29
Update ORG_PATH for test-go.sh

xieyanker:


s1061123 (approved):
Good catch! Let's merged.

PR #681, opened 2021-05-29
Update calico daemonset's url

xieyanker:
The calico's master branch structure has updated, the url is invalid.


s1061123 (changes requested):
Thank you for your PR to update the URL!

Almost good to me, but looking calico's page, 'https://docs.projectcalico.org/manifests/calico.yaml' could be more feasible instead of previous versioned manifest. 

https://docs.projectcalico.org/getting-started/kubernetes/self-managed-onprem/onpremises

So could you please change URL for that?

xieyanker:
@s1061123 OK. done!

s1061123 (approved):
LGTM!

PR #682, opened 2021-05-29
Update "kube-system" to defaultSystemNamespace

xieyanker:


s1061123:
Appreciate for your PR, but I'm wondering that we need the changes actually because values, systemNamespaces and multusNamespace, have different meanings.

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/types/conf.go#L325
This code fills default value of systemNamespaces, kube-system.
systemNamespace sets the namespace where the pod should NOT add multus interface. Usually such pod could be in kube-system, hence we put 'kube-system' as default value, but you can configure 'systemNamespaces' in cni config (see [1]).

On the other side, defaultMultusNamespace is the default value of multusNamespace, and multusNamespace specifies the namespace of clusterNetwork/defaultNetworks net-attach-def as written in following documents [1].

[1]https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md

Hence systemNamespaces and multusNamespace (and its default value, defaultMultusNamespace) are different meaning in multus cni (user may configure same value, but it is not always).

Hence if you don't mind it, I am going to close it.

Thank you so much again for your propose!

xieyanker:
@s1061123 Thanks for your detailed reply, I added a new constant `defaultSystemNamespace`. Is this ok?

s1061123:
Hi @xieyanker. I am not clear that why you need to change because 

- `systemNamespace` is CNI config parameter, so user can change it to any value
- `defaultSystemNamespace` is only used once, hence not so useful to change it as constant variable

Could you please let me know why you need to change (and update your commit log if you have)?

xieyanker:
@s1061123 OK, I will close it.

PR #683, opened 2021-06-08
Use the default socket path in GetResourceClient when unspecified

pliurh:
Signed-off-by: Peng Liu <pliu@redhat.com>

### DIFF
--- a/pkg/kubeletclient/kubeletclient.go
+++ b/pkg/kubeletclient/kubeletclient.go
@@ -23,6 +23,9 @@ const (
 
 // GetResourceClient returns an instance of ResourceClient interface initialized with Pod resource information
 func GetResourceClient(kubeletSocket string) (types.ResourceClient, error) {
+	if kubeletSocket == "" {
+		kubeletSocket, _ = util.LocalEndpoint(defaultPodResourcesPath, podresources.Socket)
+	}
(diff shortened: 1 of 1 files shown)


pkg/kubeletclient/kubeletclient.go
```diff
@@ -25,6 +25,9 @@ const (
 func GetResourceClient(kubeletSocket string) (types.ResourceClient, error) {
 	// If Kubelet resource API endpoint exist use that by default
 	// Or else fallback with checkpoint file
+	if kubeletSocket == "" {
```
yuvalk, line 28: this if should probably be placed before the above comment, which seems to reference the next if below

s1061123: well, this is not big issue I suppose, because this code (adding default kubeletSocket) is also for second 'if' statement.
To me, at least, it makes sense.

s1061123 (approved):
It looks good to me (oh, it was existed but removed by merge-conflict).

Thanks, Peng!

e0ne (approved):

pkg/kubeletclient/kubeletclient.go
pliurh: moved the comment

s1061123 (approved):
LGTM, merged!

PR #686, opened 2021-06-17
Updates entrypoint to rebuild kubeconfig when service account token or ca changes

dougbtv:
I was originally going to use inotifywait, however -- it both requires other packages, and I don't know the mechanism by which the token changes. So, this could invalidate an inotifywait if the file is replaced wholesale. So, I decided to md5sum the file and check the previous sum -- and then regenerate.


xagent003:
@dougbtv Why was this PR closed? Is there another fix for this?

Issue #654, opened 2021-04-14
Support for adding instructions for using clusterNetwork/defaultNetwork in multus-cni configuration


chenliu1993:
**What would you like to be added**:
A guide using clusterNetwork/defaultNetowrk to use multus-cni 
**Why is this needed**:
Recently I am using the default/clusterNetwork setting options to set up the multus-cni in k8s cluster. But I fail to install the multus-cni since the configuration.md doesn't contain a field called cniVersion in the top-level which is required by cni. So I think maybe it would be more helpful for a starter like me to understand the meaning and usage for cluster/defaultNetwork options.
**Example**:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md explains the field like below:

```
name (string, required): the name of the network
type (string, required): "multus"
```

and the quick-start-guide gives an example of using delegates.

But If I am using clusterNetwork/defaultNetwork instead of delegates, then a error will point out saying network info cannot be found, thus pod will stuck at ContainerCreating.

After some investigation, I find there should be a cniVersion field at the same level as default/clusterNetwork in the configuration. And delegates doesn't have this issue since delegates contains a cniVersion field in it, at the same level as plugins field in the delegates.

If I am wrong somewhere, please point me out.

Thank you.

dougbtv:
I think you're right @chenliu1993 -- looks like we have a docs deficiency.

make sure when you use `clusterNetwork` or `defaultNetwork` set you also have an accompanying net-attach-def that has the same name you're referencing.

chenliu1993:
Thank you @dougbtv 👍

PR #689, opened 2021-06-23
examples: missing cni version on version without delegates

maiqueb:
Fixes: #654


maiqueb:
I'm not sure this is the gist of the fix; from the discussion on the issue I get the impression the original issue was that how to use a multus config w/ clusterNetwork + defaultNetwork is not explained correctly, but it does look good enough to me.

dougbtv:
Perfect! Thanks for the contribution Miguel!

Issue #649, opened 2021-04-01
Enhancement: Toleration of certain API connectivity issues / errors during CNI ADD


dougbtv:
Currently during a CNI ADD, we only give a number of retries to the k8s api when it returns with [errors.IsServiceUnavailable in multus.go](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L534-L556)

The idea would be to look for critical path GET calls and control retries on specific errors a specific number of times with a tiny backoff (1s -> 3s maybe).

* [IsConnectionRefused](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/net#IsConnectionRefused) and 
* [IsConnectionReset](https://github.com/kubernetes/apimachinery/blob/v0.20.5/pkg/util/net/util.go#L41)
 
And other conditions on the hot path for pod creation.  Helper might be `IsCriticalRequestRetriable()` and it would handle 503, 500, conn reset, and conn refused.

PR #692, opened 2021-06-24
Retry on multiple error types with backoff

maiqueb:
Use exponential backoff when retrying to retrieve the pod for both ADD/DEL commands.

The retries occur when getting the pod information fails with either of the following errors:
  - service unavailable
  - server internal error
  - connection reset
  - connection refused

Fixes: https://github.com/k8snetworkplumbingwg/multus-cni/issues/649

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -40,6 +40,7 @@ import (
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/types"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	k8snet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
@@ -503,6 +504,18 @@ func cmdPluginErr(k8sArgs *types.K8sArgs, confName string, format string, args .
 	return logging.Errorf(msg+format, args...)
 }
 
+func isCriticalRequestRetriable(err error) bool {
+	logging.Debugf("isCriticalRequestRetriable: %v", err)
+	errorTypesAllowingRetry := []func(error) bool{
+		errors.IsServiceUnavailable, errors.IsInternalError, k8snet.IsConnectionReset, k8snet.IsConnectionRefused}
+	for _, f := range errorTypesAllowingRetry {
+		if f(err) {
+			return true
+		}
+	}
+	return false
+}
+
 //CmdAdd ...
 func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (cnitypes.Result, error) {
 	n, err := types.LoadNetConf(args.StdinData)
@@ -536,8 +549,8 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 		pod, err = kubeClient.GetPod(string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
 		if err != nil {
 			var waitErr error
-			// in case of ServiceUnavailable, retry 10 times with 0.5 sec interval
-			if errors.IsServiceUnavailable(err) {
+			// in case of a retriable error, retry 10 times with 0.5 sec interval
+			if isCriticalRequestRetriable(err) {
 				pollDuration := 500 * time.Millisecond
 				pollTimeout := 5 * time.Second
 				waitErr = wait.PollImmediate(pollDuration, pollTimeout, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


pkg/multus/multus.go
```diff
@@ -503,6 +504,22 @@ func cmdPluginErr(k8sArgs *types.K8sArgs, confName string, format string, args .
...
+	for _, f := range errorTypesAllowingRetry {
+		if f(err) {
+			return true
+		}
+	}
+	return false
+}
+
+func generateRetryBackoff(interval time.Duration, timeout time.Duration) wait.Backoff {
+	return wait.Backoff{Duration: interval, Factor: 2, Jitter: 0.1, Steps: 5, Cap: timeout}
```
s1061123, line 520: Can we make 1 item for 1 line (for clear visibility), as

`
        return wait.Backoff{
                Duration: interval,
                Factor: 2,
                Jitter: 0.1,
                Steps: 5,
                Cap: timeout
        }
`

pkg/multus/multus.go
```diff
@@ -503,6 +504,22 @@ func cmdPluginErr(k8sArgs *types.K8sArgs, confName string, format string, args .
...
+		errors.IsServiceUnavailable, errors.IsInternalError, k8snet.IsConnectionReset, k8snet.IsConnectionRefused}
+	for _, f := range errorTypesAllowingRetry {
+		if f(err) {
+			return true
+		}
+	}
+	return false
+}
+
+func generateRetryBackoff(interval time.Duration, timeout time.Duration) wait.Backoff {
```
s1061123, line 519: Currently we use same duration (250msec) and timeout (5sec) for all function. Can we remove both from its arugment if possible? (or make everything in argument?)

pkg/multus/multus.go
```diff
@@ -536,17 +553,17 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
 			var waitErr error
-			// in case of ServiceUnavailable, retry 10 times with 0.5 sec interval
-			if errors.IsServiceUnavailable(err) {
-				pollDuration := 500 * time.Millisecond
+			// in case of a retriable error, retry for 5 seconds, with exponential backoff
+			if isCriticalRequestRetriable(err) {
+				pollDuration := 250 * time.Millisecond
 				pollTimeout := 5 * time.Second
-				waitErr = wait.PollImmediate(pollDuration, pollTimeout, func() (bool, error) {
+				waitErr = wait.ExponentialBackoff(generateRetryBackoff(pollDuration, pollTimeout), func() (bool, error) {
```
s1061123, line 560: @dougbtv What do you think about to change to ExponentialBackoff()?

s1061123:
Thank you so much for the PR! Almost concept is OK to me but I have some comments, so put it. 

maiqueb (commented):
Thanks for the review @s1061123 .

pkg/multus/multus.go
s1061123: @maiqueb Thank you for your reply. I understand that, yeah, it is pretty trade-off. Currently we do not have strict timeout tuning and we do not have strict requirement from other components, hence I guess we don't need to make it parameter for now.

Once we face such case, then we just change it, I guess.

pkg/multus/multus.go
dougbtv: I think we need pollimmediate here too, it's so we can keep retrying quickly during install/upgrade so we can retry when other components are in flux.

maiqueb: OK, I had misunderstood the issue - using a smaller retry frequency, with a smaller timeout (50% less).

pkg/multus/multus.go
```diff
@@ -536,17 +549,17 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
 		if err != nil {
 			var waitErr error
-			// in case of ServiceUnavailable, retry 10 times with 0.5 sec interval
-			if errors.IsServiceUnavailable(err) {
-				pollDuration := 500 * time.Millisecond
-				pollTimeout := 5 * time.Second
+			// in case of a retriable error, retry 10 times with 0.25 sec interval
+			if isCriticalRequestRetriable(err) {
+				pollDuration := 250 * time.Millisecond
+				pollTimeout := 2500 * time.Millisecond
```
dougbtv, line 555: optional nit: this could go in a top level constant so it's in the same place as the pollDuration is now, such as: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L52-L55

dougbtv (approved):
With the change back to pollImediate. I'm OK with where we are, one optional nit.

pkg/multus/multus.go
maiqueb: Done.

s1061123 (approved):
Looks good to me, thank you for incorporate comments!

Issue #700, opened 2021-07-27
security vulnerabilities in latest container image v3.7


blackgold:
Container scan found issues with  packages bind-license and glib2 inside the container image. Following are the vulnerabilities. 

Vulnerability found in os package type (rpm) - bind-license (fixed in: 32:9.11.4-16.P2.el7_8.6)(CVE-2020-8616 - https://access.redhat.com/security/cve/CVE-2020-8616) 
Vulnerability found in os package type (rpm) - bind-license (fixed in: 32:9.11.4-16.P2.el7_8.6)(CVE-2020-8617 - https://access.redhat.com/security/cve/CVE-2020-8617) 
Vulnerability found in os package type (rpm) - bind-license (fixed in: 32:9.11.4-26.P2.el7_9.4)(CVE-2020-8625 - https://access.redhat.com/security/cve/CVE-2020-8625) 
Vulnerability found in os package type (rpm) - bind-license (fixed in: 32:9.11.4-26.P2.el7_9.5)(CVE-2021-25215 - https://access.redhat.com/security/cve/CVE-2021-25215) 
Vulnerability found in os package type (rpm) - glib2 (fixed in: 0:2.56.1-9.el7_9)(CVE-2021-27219 - https://access.redhat.com/security/cve/CVE-2021-27219)

blackgold:
These CVEs don't directly impact multus, however it will be nice to upgrade the os packages in the image.

aostapenko:
We're also awaiting for image fixed. Looks like centos7 repositories don't have fixed packages - doing yum update did not work for me.

aostapenko:
Please disregard my previous message - I misread Dockerfile. Created a pull request addressing this issue https://github.com/k8snetworkplumbingwg/multus-cni/pull/721

PR #704, opened 2021-08-03
Upgrades to all Dockerfiles. Make Golang version consistent across images.

gaby:
* Looking at the different Dockerfiles I noticed that all of them were using different versions of Golang, from 1.13 to 1.15. This PR takes care of using the same version (1.16, same as go.mod file) across all the images.
* This PR also takes care of running `yum update` on the final image to take care of any last minute package updates to the base image.
* The image for `openshift/origin-base` used hasn't been updated in 2 years, updated the tag to `v3.11` which is the latest available, but I have no way of testing this.

On a note, is there a reason the final images use CentOS instead of Alpine or Google/Distroless ?

This PR may fix some of the issues presented in #700

### DIFF
--- a/deployments/Dockerfile.arm64
+++ b/deployments/Dockerfile.arm64
@@ -1,5 +1,5 @@
 # This Dockerfile is used to build the image available on DockerHub
-FROM golang:1.13.4 as build
+FROM golang:1.16.6 as build
 
 # Add everything
 ADD . /usr/src/multus-cni
@@ -14,6 +14,7 @@ RUN  cd /usr/src/multus-cni && \
 FROM arm64v8/centos:7
 LABEL org.opencontainers.image.source https://github.com/k8snetworkplumbingwg/multus-cni
 COPY --from=build /usr/src/multus-cni /usr/src/multus-cni
+RUN yum update
 
 WORKDIR /
 ADD ./images/entrypoint.sh /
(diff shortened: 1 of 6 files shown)


deployments/Dockerfile.arm64
```diff
@@ -14,6 +14,7 @@ RUN  cd /usr/src/multus-cni && \
 FROM arm64v8/centos:7
 LABEL org.opencontainers.image.source https://github.com/k8snetworkplumbingwg/multus-cni
 COPY --from=build /usr/src/multus-cni /usr/src/multus-cni
+RUN yum update
```
cgchinmay, line 17: I have been having issues running yum update for arm based dockerfiles. I am yet to rootcause this, but this the error I get

```
standard_init_linux.go:211: exec user process caused "exec format error"
--
1191 | The command '/bin/sh -c yum update -y &&     yum clean all &&     rm -rf /var/cache/yum' returned a non-zero code: 1
```
Were you able to test it out ?

dougbtv:
Gaby! Thanks for the contribution. We think this makes sense, and having the updated go version.

I approved CI, can you check and see why some are failing? Much appreciated.

Thanks for the pointer on the OpenShift image, you can leave that alone if you wish, and I'll follow up with it, it likely needs to use a different image/tag to come up to date.

We're open to having other Dockerfiles for additional distributions, as well.

gaby:
@dougbtv Just committed a fix for the build failures, Also reverted the changes to the openshift image version.

PR #718, opened 2021-09-01
Replace entrypoint script with initcontainers

maiqueb:
TODO:
  - make sure this works ... didn't actually deploy a pod w/ multus yet.
  - replace the multus conf section
  - replace the builder img / extra dependencies


tools/generate-kubeconfig.go
```diff
@@ -0,0 +1,123 @@
...
+func k8sKubeConfigToken(tokenPath string) (string, error) {
+	data, err := ioutil.ReadFile(tokenPath)
+	if err != nil {
+		return "", fmt.Errorf("failed reading file %s: %w", tokenPath, err)
+	}
+	return string(data), nil
+}
+
+func writeKubeConfig(outputPath string, protocol string, k8sServiceIP string, k8sServicePort int, tlsConfig string, serviceAccountToken string) error {
+	kubeConfigTemplate := `
```
maiqueb, line 90: This is simpler, but I think that using the api object then generating a string out of it would be more correct.

As long as this is tested, I guess we're OK.

images/multus-daemonset.yml
```diff
@@ -190,6 +190,40 @@ spec:
...
+              cpu: "10m"
+              memory: "15Mi"
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - name: cnibin
+              mountPath: /host/opt/cni/bin
+              mountPropagation: Bidirectional
+        - name: generate-kubeconfig
+          image: docker.io/maiqueb/multus:latest
```
s1061123, line 211: can we use `ghcr.io/k8snetworkplumbingwg/multus-cni:stable`?

images/multus-daemonset.yml
```diff
@@ -190,6 +190,40 @@ spec:
           mountPath: /host/opt/cni/bin
         - name: multus-cfg
           mountPath: /tmp/multus-conf
+      initContainers:
+        - name: install-multus-binary
+          image: ghcr.io/k8snetworkplumbingwg/multus-cni:stable
+          command:
+            - "cp"
```
dougbtv, line 197: `cp -f` maybe?

images/multus-daemonset.yml
maiqueb: sure, this is just me being lazy and not using a local registry :) 

I'll update before I un-draft the PR.

deployments/Dockerfile
```diff
@@ -1,21 +1,15 @@
...
 
-FROM centos:centos7
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 LABEL org.opencontainers.image.source https://github.com/k8snetworkplumbingwg/multus-cni
 COPY --from=build /usr/src/multus-cni /usr/src/multus-cni
 WORKDIR /
 
-ADD ./images/entrypoint.sh /
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["sleep"]
```
dougbtv, line 15: Yep, I think this is OK for a default. We'll probably have a running process to do some of the "loopy" stuff that we do in the entrypoint, as we progress, that is...

Like, configuration regeneration: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/entrypoint.sh#L426-L455

Also, something that we can actually omit -- and cover when we have the thick plugin architecture is this hacky PR from me that's not merged: https://github.com/k8snetworkplumbingwg/multus-cni/pull/686

It's a hacky way to update the secret, but in thick-plugin mode, we can just rely on client-go to update it for us, and we won't actually have to generate the secret anymore...

...for now though, let's keep this, just providing a bit of context too, as I look at it.

maiqueb: > Yep, I think this is OK for a default. We'll probably have a running process to do some of the "loopy" stuff that we do in the entrypoint, as we progress, that is...
> 
> Like, configuration regeneration: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/entrypoint.sh#L426-L455

Added it to the PR's description [TODO list](https://github.com/k8snetworkplumbingwg/multus-cni/pull/718#issue-724640335).

> 
> Also, something that we can actually omit -- and cover when we have the thick plugin architecture is this hacky PR from me that's not merged: #686
> 
> It's a hacky way to update the secret, but in thick-plugin mode, we can just rely on client-go to update it for us, and we won't actually have to generate the secret anymore...
> 
> ...for now though, let's keep this, just providing a bit of context too, as I look at it.

Resolving this conversation since I've tracked it in the PR's description TODO list.

images/multus-daemonset.yml
maiqueb: Another thing that comes to mind is we won't be able to update this in one go: unless we immediately cherry-pick this PR to stable - which I don't recommend :) 

I guess for now we should just make sure the e2e part is working correctly, and later on, (after this is cherry-picked to stable) on a separate PR, update the install script.

I also have some conflicting ideas about non-kubernetes installations ... The entry point script would be needed for that. Is that something we want to keep supporting ? @s1061123 / @dougbtv

maiqueb: We've discussed this offline; the plan is:
- issue a multus release before this PR - and all other thick plugin work.
- keep referencing in the official installation this entrypoint based script until the thick plugin approach matures enough to enter the `stable` branch.
- once the thick plugin approach is mature enough, update the multus-daemonset yaml file to reflect the new installation type.

images/multus-daemonset.yml
maiqueb: Pardon my ignorance, but what is the advantage ?

tools/generate-kubeconfig.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	if *kubeCAFilePath == defaultK8sServiceHost {
+		*kubeCAFilePath = serviceAccountPath + "/ca.crt"
+	}
+
+	tlsCfg := "insecure-skip-tls-verify: true"
+	if !*skipTLSVerify {
+		kubeCAFileContents, err := k8sCAFileContentsBase64(*kubeCAFilePath)
+		if err != nil {
+			logInvalidArg(fmt.Errorf("failed grabbing CA file: %s", err.Error()))
```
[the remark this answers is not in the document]
maiqueb: Done.

tools/generate-kubeconfig.go
maiqueb: Keeping it simple and stupid.

maiqueb (commented):
Resolving some comments.

maiqueb:
Rebased to latest master (tagged version 3.8) in [this forced push](https://github.com/k8snetworkplumbingwg/multus-cni/compare/ca3af60af78bc9536f61b5f0bcc53a7bc5936770..b50f071c691381a926feb39be4b4224cd399f995).

hack/build-go.sh
```diff
@@ -41,6 +41,8 @@ if [ "$GO111MODULE" == "off" ]; then
 	export GOBIN=${PWD}/bin
 	export GOPATH=${PWD}/gopath
 	go build -o ${PWD}/bin/multus -tags no_openssl -ldflags "${LDFLAGS}" "$@" ${REPO_PATH}/cmd
+	go build -o ${PWD}/bin/multus-daemon -tags no_openssl -ldflags "${LDFLAGS}" "$@" ${REPO_PATH}/cmd/thick
```
s1061123, line 44: How about to change the file layout as following?
- cmd/multus
- cmd/multus-daemon (I feel that 'multus-daemon' is for thick plugin server, so you may change the name of binary because it might be simple loop for thin plugin, current multus)
- cmd/generate-kubeconfig

maiqueb: I'm OK with that; I originally named it because using the thick plugin option sounded like a certainty at the time. 

I'll try to come up with something better: `multus-controller` possibly, or `pod-watcher`, `pod-monitor` , etc. 

I think I prefer `multus-controller` since it sounds less detailed (if in the future in starts watching some other resource, the binary name becomes misleading).

maiqueb:
@s1061123 / @dougbtv added a new commit where I prototype the usage of https://github.com/koron/go-dproxy . 

Pretty simple to use, and I think it covers our use cases. 

I've isolated its usage in a separate commit so we can remove it if we feel that's not a good idea. If it indeed is, I'll squash all WIP commits into the commit that introduces the multus daemon.

hack/build-go.sh
maiqueb: @s1061123 does this new directory structure make sense to you ?

pkg/config/manager.go
```diff
@@ -0,0 +1,168 @@
+package config
```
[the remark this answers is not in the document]
maiqueb: Done.

pkg/config/generator_test.go
```diff
@@ -0,0 +1,220 @@
+package config
```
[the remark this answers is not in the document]
maiqueb: Done.

pkg/config/generator.go
```diff
@@ -0,0 +1,170 @@
+package config
```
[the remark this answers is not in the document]
maiqueb: Done.

cmd/controller/main.go
```diff
@@ -0,0 +1,213 @@
+package main
```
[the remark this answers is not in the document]
maiqueb: Done.

cmd/config-generation/kubeconfig.go
```diff
@@ -0,0 +1,132 @@
+package main
```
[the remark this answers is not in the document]
maiqueb: Done.

pkg/config/manager_test.go
```diff
@@ -0,0 +1,100 @@
+package config
```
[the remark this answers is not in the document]
maiqueb: Done,

PR #722, opened 2021-09-09
Fix typos

zhangzujian:
Signed-off-by: 张祖建 <zhangzujian.7@gmail.com>

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// This is a "Multi-plugin".The delegate concept refered from CNI project
+// This is a "Multi-plugin".The delegate concept referred from CNI project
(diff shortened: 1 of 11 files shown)


cmd/main.go
```diff
@@ -30,7 +30,7 @@ import (
 
 func main() {
 
-	// Init command line flags to clear vendored packages' one, especially in init()
+	// Init command line flags to clear vendorred packages' one, especially in init()
```
s1061123, line 33: this should be vendored. See https://golang.org/src/README.vendor for the example.

zhangzujian: > this should be vendored. See https://golang.org/src/README.vendor for the example.

fixed

dougbtv (approved):

s1061123:
Merged, thank you so much for the PR!

PR #723, opened 2021-09-13
Update golang version (at least >= 1.16)

s1061123:
Currently golang community supports 1.16 and later, hence, need
to update golang version in our build pipeline. This change
updates golang version to 1.17/1.16.


dougbtv (approved):
LGTM!

PR #724, opened 2021-09-16
Downgrade Kubernetes vendor version

s1061123:
Multus community decide that multus support the Kubernetes that
Kubernetes community supports, so changed vendor code to the
oldest version of currently supported.


dougbtv (approved):
LGTM!

PR #752, opened 2021-11-05
multus cni: add / remove single network / interface

maiqueb:
This PR will grant multus the ability to process a **single**
network/interface pair when a CNI `ADD` / `DELETE` is processed.

This is a first required step to add / remove network interfaces to / from
running PODs.

Multus will differentiate which operation to perform based on the
presence of the `K8S_POD_NETWORK` entry in the `k8sArgs`
args structure: when **it is present**, we will consider load the 
corresponding delegate and add (or remove ...) that secondary 
network interface.

This new entry will be in the `<network name>` format (since the
k8sArgs structure already features an entry for the namespace).

### DIFF
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -16,12 +16,8 @@ package k8sclient
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"net"
 	"os"
-	"regexp"
-	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


Skazza94:
Attempting to remove the stale label.

P.S. When will be this PR merged? It looks very promising 😄

maiqueb:
> Attempting to remove the stale label.
> 
> P.S. When will be this PR merged? It looks very promising smile

After meeting with the multus maintainers, this approach was turned down. We are now focusing on something different, and are actively working on it on branch https://github.com/k8snetworkplumbingwg/multus-cni/tree/feature/multus-4.0. 

It features a refactor of multus to a thick plugin (already available, pr #771 ), and will feature a short refactor of the multus lib allowing the functions that invoke delegates from outside the multus package. 

The plan then is to add a controller that will provide multus with dynamic network attachments for pods (pr #777).

Skazza94:
Hi @maiqueb and thanks for the prompt reply! 😄
Thanks for all the information, I'll wait for multus 4.0 then!

Issue #756, opened 2021-11-11
flaky unit test: "Reacts to updates of the delegated CNI configuration"


maiqueb:
**What happened**:
Flaky unit test on CI: test [Configuration Manager Reacts to updates of the delegated CNI configuration](https://github.com/k8snetworkplumbingwg/multus-cni/blob/7091831a00b6b421167dbc1f8dad90d9c20362c7/pkg/config/manager_test.go#L83) sometimes fails, since it receives a different event (e.g. an update on some other `/tmp/...` file.

**What you expected to happen**:
The unit tests should be deterministic.

PR #757, opened 2021-11-11
flaky: fix delegate CNI conf updates unit test

maiqueb:
The test was just checking that a READ/WRITE `fsnotify.Event` for
the multus configuration was being seen; this patch changes this
behavior, and assures that the delegateCNI configuration update results
in turn on the update of the multus configuration file.

The multus tmp directory for the CNI configuration is also updated.

Fixes: #756

### DIFF
--- a/pkg/config/manager_test.go
+++ b/pkg/config/manager_test.go
@@ -49,15 +49,22 @@ var _ = Describe(suiteName, func() {
   "dns": {}
 }
 `
-		tmpDir = "/tmp"
 	)
 
 	var configManager *Manager
+	var multusConfigDir string
 	var multusConfigFilePath string
 
 	BeforeEach(func() {
+		var err error
+		multusConfigDir, err = ioutil.TempDir("", "multus-config")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(os.MkdirAll(multusConfigDir, 0755)).To(Succeed())
+	})
+
+	BeforeEach(func() {
 		format.TruncatedDiff = false
-		multusConfigFilePath = fmt.Sprintf("%s/%s", tmpDir, primaryCNIPluginName)
+		multusConfigFilePath = fmt.Sprintf("%s/%s", multusConfigDir, primaryCNIPluginName)
 		Expect(ioutil.WriteFile(multusConfigFilePath, []byte(primaryCNIPluginTemplate), userRWPermission)).To(Succeed())
 
 		multusConf := NewMultusConfig(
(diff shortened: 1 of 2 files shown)


maiqueb:
@s1061123 I think tracking updates on a path other than `/tmp/...` is enough. 

Can you give this a try ? I wasn't able to reproduce this locally.

s1061123:
@maiqueb sure. will do that.

s1061123:
@maiqueb test is failed. is that related  yours?
https://github.com/k8snetworkplumbingwg/multus-cni/runs/4191615026?check_suite_focus=true

maiqueb:
> @maiqueb test is failed. is that related yours? https://github.com/k8snetworkplumbingwg/multus-cni/runs/4191615026?check_suite_focus=true

It was. Now it is more robust, but not bullet proof. I'm running 100 executions locally in a loop, and every now and then it fails 0_o . Around 1, 2 % of the times.

I've also noticed the `Eventually` sometimes becomes blocked... Will keep investigating.

maiqueb:
Ready for review @s1061123

pkg/config/manager_test.go
```diff
@@ -80,28 +81,37 @@ var _ = Describe(suiteName, func() {
...
+			close(stopChannel)
+		})
+
+		It("Trigger the re-generation of the Multus CNI configuration", func() {
+			newCNIConfig := "{\"cniVersion\":\"0.4.0\",\"dns\":{},\"ipam\":{},\"name\":\"yoyo-newnet\",\"type\":\"mycni\"}"
+			Expect(ioutil.WriteFile(defaultCniConfig, []byte(newCNIConfig), userRWPermission)).To(Succeed())
+
+			multusCniConfigFile := fmt.Sprintf("%s/%s", multusConfigDir, multusConfigFileName)
+			Eventually(func() (string, error) {
+				bytes, err := ioutil.ReadFile(multusCniConfigFile)
```
s1061123, line 111: variable name `bytes` is overlapped with package name [bytes](https://pkg.go.dev/bytes). Could you please rename it?

s1061123 (changes requested):

pkg/config/manager_test.go
maiqueb: Sure; done in this [forced push](https://github.com/k8snetworkplumbingwg/multus-cni/compare/b248141bf15e374d466a730d955fe51fd9d3899a..0da544985452687d4175bf27e5b4e38892f67ddd).

s1061123 (approved):

dougbtv (approved):

Issue #737, opened 2021-10-15
Interworking with 0.4.0 and 0.3.1 (or previous) CNI version


s1061123:
Since CNI version 0.4.0, CHECK command is introduced. If user configures multus config as follows, then CHECK command will be sent to delegate CNI, flannel and we get error message from kubelet.

What should we do for that...

```
[fedora@kube-master1 ~]$ sudo cat /etc/cni/net.d/00-multus.conf  | jq
{
  "cniVersion": "0.4.0",
  "name": "multus-cni-network",
  "type": "multus",
  "capabilities": {
    "portMappings": true
  },
  "logLevel": "debug",
  "logFile": "/tmp/multus.log",
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
  "delegates": [
    {
      "name": "cbr0",
      "cniVersion": "0.3.1",
      "plugins": [
        {
          "type": "flannel",
          "delegate": {
            "hairpinMode": true,
            "isDefaultGateway": true
          }
        },
        {
          "type": "portmap",
          "capabilities": {
            "portMappings": true
          }
        }
      ]
    }
  ]
}

```

```
  Warning  FailedCreatePodSandBox  2s    kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get network status for pod sandbox k8s_fedora-nonet_default_888e2e8f-4551-46d2-858a-87e5bbeb31e7_0(457da3c23a303d35c61737e78f5a7a2996f881516ab3dcd4f84382943928b42c): delegateCheck: error invoking ConflistCheck - "cbr0": conflistCheck: error in getting result from CheckNetworkList: configuration version "0.3.1" does not support the CHECK command
```

PR #762, opened 2021-11-19
check version incompatibility

bn222:
When nested CNI version is 0.3.1 or less while the top level CNI version
is 0.4.0, error out. Version 0.3.1 or less doesn't support the CHECK
command. This closes issue #737.

### DIFF
--- a/pkg/config/generator.go
+++ b/pkg/config/generator.go
@@ -17,12 +17,15 @@ package config
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
 	"sort"
 	"strings"
 	"time"
+
+	"github.com/blang/semver"
 )
 
 const (
@@ -52,7 +55,7 @@ type MultusConf struct {
 
 // NewMultusConfig creates a basic configuration generator. It can be mutated
 // via the `With...` methods.
-func NewMultusConfig(pluginName string, cniVersion string, kubeconfig string, configurationOptions ...Option) *MultusConf {
+func NewMultusConfig(pluginName string, cniVersion string, kubeconfig string, configurationOptions ...Option) (*MultusConf, error) {
 	multusConfig := &MultusConf{
 		Name:         MultusDefaultNetworkName,
 		CNIVersion:   cniVersion,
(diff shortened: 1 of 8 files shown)


dougbtv:
Hey @bn222 ! Thanks for the contribution.

We'll also need to take care of this in the next generation configuration generation, which can be found approximately here:

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/cmd/controller/main.go
https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/cmd/config-generation/kubeconfig.go

bn222:
Hi @dougbtv, I'm not sure what I should change in the two files you mention. I guess you might actually mean the following file:

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/config/generator.go#L37

I could enforce checking that the delegate configurations are using a compatible CNI version, although that would require me to change the type of Delegates in MultusConf if we want to avoid converting to json and checking the cni version in json.

Let me know what you think!

dougbtv:
Hey Balazs, what I mean is that we have two methods for entrypoint as we're in a transition (which will be superceded by the go code, making the bash script obsolete)

So, in two different modes, either the bash runs, or the go runs (the files I linked)

Meaning that you'll have to implement the equivalent logic in the the go code as well as the bash :+1:

pkg/config/generator.go
```diff
@@ -70,6 +73,39 @@ func NewMultusConfig(pluginName string, cniVersion string, kubeconfig string, co
...
+			delegateVersion, ok := delegatesMap["cniVersion"].(string)
+			if !ok {
+				s := "Couldn't get cni version of delegate"
+				return "", errors.New(s)
+			}
+			v, err := semver.Make(delegateVersion)
+			if err != nil {
+				return "", err
+			}
+			if v.LTE(v031) {
```
maiqueb, line 102: I don't understand why 0.3.1 is special here. Shouldn't we just reject *any* delegates whose version is lower than 0.4.0 ?

pkg/config/generator.go
```diff
@@ -70,6 +73,39 @@ func NewMultusConfig(pluginName string, cniVersion string, kubeconfig string, co
 // Generate generates the multus configuration from whatever state is currently
 // held
 func (mc *MultusConf) Generate() (string, error) {
+	versionFmt := "delegate cni version is %s while top level cni version is %s"
+	v040, _ := semver.Make("0.4.0")
+	v031, _ := semver.Make("0.3.1")
+	topLevel, err := semver.Make(mc.CNIVersion)
```
maiqueb, line 79: please name this `multusCNIVersion` or something alike. 

I don't expect a version number to be in a variable named `topLevel` .

pkg/config/generator.go
```diff
@@ -70,6 +73,39 @@ func NewMultusConfig(pluginName string, cniVersion string, kubeconfig string, co
...
+			if err != nil {
+				return "", err
+			}
+			if v.LTE(v031) {
+				s := fmt.Sprintf(versionFmt, delegateVersion, mc.CNIVersion)
+				return "", errors.New(s)
+			}
+		}
+	}
+
```
maiqueb, line 108: Please encapsulate this in a separate func / method - the simple `Generate` method turned into 95% checking a corner case and 5% generating.

I'd actually call this new function from the constructor, thus preventing the creation of an invalid config.

You can change the signature of the constructor to also return an error.

maiqueb (changes requested):
Other than the comments I'm providing in the code, I can't help to notice that the new functionality  is not being unit tested; you'll have to provide these tests as well.

I also disagree with your choice of where to implement this: I think it should be done in the configuration generator constructor, not in its generate method.

pkg/config/generator.go
bn222: I addressed this by only checking nested version < 0.4.0 when top level version >= 0.4.0.

pkg/config/generator.go
bn222: Done. This caused a few more changes elsewhere as well as the error is propagated.

pkg/config/generator_test.go
```diff
@@ -45,24 +45,28 @@ var primaryCNIConfig = map[string]interface{}{
...
+	if err != nil {
+		return multusConfig, err
+	}
+	err = multusConfig.Mutate(withDelegates(primaryCNIPluginConfig.(map[string]interface{})))
+	return multusConfig, err
 }
 
 func TestBasicMultusConfig(t *testing.T) {
-	multusConfig := newMultusConfigWithDelegates(
+	multusConfig, _ := newMultusConfigWithDelegates(
```
maiqueb, line 58: Now that the method returns an error, it feels weird to swallow it up.

Do you mind asserting it does not error on the tests ? Or better; provide another helper for it ?

maiqueb (changes requested):
It feels a bit weird to not check the error on `newMultusConfigWithDelegates` now that it can return one. 

Unfortunately, I don't have a good alternative to doing that: either check it on the test's body, or introduce another helper for it.

Pretty much everything else is opinionated.

maiqueb (approved):
Thanks for the patience, and for the patch.

@dougbtv I've **only** reviewed the golang related code.

s1061123:
It looks good to me. Apologize that the PR takes a very long time. Let me merge into master. Thank you so much!

PR #763, opened 2021-11-19
Pod UID mismatches should only warn on CNI DEL

dougbtv:
When checking the pod UID, on CNI DEL, we should not error out.

The pod UID can mismatch in the case of a stateful set (for example) and the pod namespace and name remain the same, but the pod UID changes. Erroring on DEL can cause the pod to not be properly cleaned up.

In this case, we can just return without the current results from the API about the pod, and simply operate on the cache in order to make the delete.


s1061123:
LGTM, thanks for the fix!

PR #764, opened 2021-12-05
Update libcni cache when default-route net selection is used

s1061123:
To keep consistency between actual network and CNI result in cache, update libcni cache when multus add/del default routes by
`default-route` network selection.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -636,10 +636,11 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			}
 		}
 
+		netName := ""
 		tmpResult, err = delegateAdd(exec, kubeClient, pod, ifName, delegate, rt, n, cniArgs)
 		if err != nil {
 			// If the add failed, tear down all networks we already added
-			netName := delegate.Conf.Name
+			netName = delegate.Conf.Name
 			if netName == "" {
 				netName = delegate.ConfList.Name
 			}
@@ -649,11 +650,12 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 		}
 
 		// Remove gateway from routing table if the gateway is not used
-		deletegateway := false
+		deleteV4gateway := false
+		deleteV6gateway := false
 		adddefaultgateway := false
-		if delegate.IsFilterGateway {
-			deletegateway = true
-			logging.Debugf("Marked interface %v for gateway deletion", ifName)
+		if delegate.IsFilterV4Gateway {
+			deleteV4gateway = true
+			logging.Debugf("Marked interface %v for v4 gateway deletion", ifName)
 		} else {
 			// Otherwise, determine if this interface now gets our default route.
 			// According to
@@ -661,24 +663,53 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			// the list can be empty; if it is, we'll assume the CNI's config for the default gateway holds,
 			// else we'll update the defaultgateway to the one specified.
 			if delegate.GatewayRequest != nil && delegate.GatewayRequest[0] != nil {
-				deletegateway = true
+				deleteV4gateway = true
 				adddefaultgateway = true
 				logging.Debugf("Detected gateway override on interface %v to %v", ifName, delegate.GatewayRequest)
 			}
 		}
 
-		if deletegateway {
-			tmpResult, err = netutils.DeleteDefaultGW(args, ifName, &tmpResult)
-			if err != nil {
-				return nil, cmdErr(k8sArgs, "error deleting default gateway: %v", err)
+		if delegate.IsFilterV6Gateway {
+			deleteV6gateway = true
+			logging.Debugf("Marked interface %v for v6 gateway deletion", ifName)
+		} else {
+			// Otherwise, determine if this interface now gets our default route.
+			// According to
+			// https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ (4.1.2.1.9)
+			// the list can be empty; if it is, we'll assume the CNI's config for the default gateway holds,
+			// else we'll update the defaultgateway to the one specified.
+			if delegate.GatewayRequest != nil && delegate.GatewayRequest[0] != nil {
+				deleteV6gateway = true
+				adddefaultgateway = true
+				logging.Debugf("Detected gateway override on interface %v to %v", ifName, delegate.GatewayRequest)
 			}
 		}
 
-		// Here we'll set the default gateway
+		// Remove namespace from delegate.Name for Add/Del CNI cache
+		nameSlice := strings.Split(delegate.Name, "/")
+		netName = nameSlice[len(nameSlice) - 1]
+
+		// Remove gateway if `default-route` network selection is specified
+		if deleteV4gateway || deleteV6gateway  {
+			err = netutils.DeleteDefaultGW(args, ifName)
+			if err != nil {
+				return nil, cmdErr(k8sArgs, "error deleting default gateway: %v", err)
+			}
+			err = netutils.DeleteDefaultGWCache(n.CNIDir, rt, netName, ifName, deleteV4gateway, deleteV6gateway)
+			if err != nil {
+				return nil, cmdErr(k8sArgs, "error deleting default gateway in cache: %v", err)
+			}
+		}
+
+		// Here we'll set the default gateway which specified in `default-route` network selection
 		if adddefaultgateway {
-			tmpResult, err = netutils.SetDefaultGW(args, ifName, delegate.GatewayRequest, &tmpResult)
+			err = netutils.SetDefaultGW(args, ifName, delegate.GatewayRequest)
 			if err != nil {
 				return nil, cmdErr(k8sArgs, "error setting default gateway: %v", err)
+			}
+			err = netutils.AddDefaultGWCache(n.CNIDir, rt, netName, ifName, delegate.GatewayRequest)
+			if err != nil {
+				return nil, cmdErr(k8sArgs, "error setting default gateway in cache: %v", err)
 			}
 		}
 
--- a/pkg/netutils/netutils.go
+++ b/pkg/netutils/netutils.go
@@ -16,26 +16,24 @@
 package netutils
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"path/filepath"
+
+	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/skel"
-	cnitypes "github.com/containernetworking/cni/pkg/types"
-	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/vishvananda/netlink"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
-	"net"
-	"strings"
 )
 
 // DeleteDefaultGW removes the default gateway from marked interfaces.
-func DeleteDefaultGW(args *skel.CmdArgs, ifName string, res *cnitypes.Result) (*current.Result, error) {
-	result, err := current.NewResultFromResult(*res)
-	if err != nil {
-		return nil, logging.Errorf("DeleteDefaultGW: Error creating new from current CNI result: %v", err)
-	}
-
+func DeleteDefaultGW(args *skel.CmdArgs, ifName string) error {
 	netns, err := ns.GetNS(args.Netns)
 	if err != nil {
-		return nil, logging.Errorf("DeleteDefaultGW: Error getting namespace %v", err)
+		return logging.Errorf("DeleteDefaultGW: Error getting namespace %v", err)
 	}
 	defer netns.Close()
 
@@ -50,40 +48,27 @@ func DeleteDefaultGW(args *skel.CmdArgs, ifName string, res *cnitypes.Result) (*
 		}
 		return err
 	})
-	var newRoutes []*cnitypes.Route
-	for _, route := range result.Routes {
-		if mask, _ := route.Dst.Mask.Size(); mask != 0 {
-			newRoutes = append(newRoutes, route)
-		}
-	}
-	result.Routes = newRoutes
-	return result, err
+	return err
 }
 
 // SetDefaultGW adds a default gateway on a specific interface
-func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cnitypes.Result) (*current.Result, error) {
-
-	// Use the current CNI result...
-	result, err := current.NewResultFromResult(*res)
-	if err != nil {
-		return nil, logging.Errorf("SetDefaultGW: Error creating new CNI result from current: %v", err)
-	}
-
+func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP) error {
 	// This ensures we're acting within the net namespace for the pod.
 	netns, err := ns.GetNS(args.Netns)
 	if err != nil {
-		return nil, logging.Errorf("SetDefaultGW: Error getting namespace %v", err)
+		return logging.Errorf("SetDefaultGW: Error getting namespace %v", err)
 	}
 	defer netns.Close()
-
-	var newResultDefaultRoutes []*cnitypes.Route
 
 	// Do this within the net namespace.
 	err = netns.Do(func(_ ns.NetNS) error {
 		var err error
 
 		// Pick up the link info as we need the index.
-		link, _ := netlink.LinkByName(ifName)
+		link, err := netlink.LinkByName(ifName)
+		if err != nil {
+			return logging.Errorf("SetDefaultGW: Error getting link %v", err)
+		}
 
 		// Cycle through all the desired gateways.
 		for _, gw := range gateways {
@@ -95,15 +80,6 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
 				Gw:        gw,
 			}
 
-			// Build a new element for the results route
-
-			// Set a correct CIDR depending on IP type
-			_, dstipnet, _ := net.ParseCIDR("::0/0")
-			if strings.Count(gw.String(), ":") < 2 {
-				_, dstipnet, _ = net.ParseCIDR("0.0.0.0/0")
-			}
-			newResultDefaultRoutes = append(newResultDefaultRoutes, &cnitypes.Route{Dst: *dstipnet, GW: gw})
-
 			// Perform the creation of the default route....
 			err = netlink.RouteAdd(&newDefaultRoute)
 			if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/netutils/netutils.go
```diff
@@ -113,7 +89,350 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+}
+
+// AddDefaultGWCache updates libcni cache to add default gateway result
+func AddDefaultGWCache(cacheDir string, rt *libcni.RuntimeConf, netName string, ifName string, gw []net.IP) error {
+	cacheFile := filepath.Join(cacheDir, "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
+
+	cache, err := ioutil.ReadFile(cacheFile)
+	if err != nil {
+		return err
+	}
```
maiqueb, line 306: This is also used when on the `DeleteDefaultGWCache` function - as such, I think it should be exported as a function and called here.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+	result, err := types020.GetResult(resultRaw)
+	Expect(err).NotTo(HaveOccurred())
+
+	for _, r := range result.IP6.Routes {
+		if r.Dst.String() == "::/0" {
+			return true
+		}
+	}
+	return false
+}
```
maiqueb, line 104: plenty is duplicated here; you could export a 3rd function with most of the code and re-use it from the IP version wrappers:

```golang
func test020ResultHasIPv4DefaultRoute(data []byte) bool {
    v4RouteSelectorFromResult := func (res types020.Result) []cnitypes.Route { 
        return result.IP4.Routes
    }
    return test020ResultHasDefaultRoute(data, v4RouteSelectorFromResult, "0.0.0.0/0")
}

func test020ResultHasIPv6DefaultRoute(data []byte) bool {
    v6RouteSelectorFromResult := func (res types020.Result) []cnitypes.Route {
        return result.IP6.Routes
    }
    return test020ResultHasDefaultRoute(data, v6RouteSelectorFromResult, "::0/0")
}

func test020ResultHasDefaultRoute(data []byte, routeSelector func (result types020.Result) []cnitypes.Route, defaultRouteStr string) bool {
	resultRaw, err := types020.NewResult(data)

	result, err := types020.GetResult(resultRaw)

	for _, r := range routeSelector(result) {
		if r.Dst.String() == defaultRouteStr {
			return true
		}
	}
	return false
}
```

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+
+	for _, r := range result.Routes {
+		if r.Dst.String() == "::/0" {
+			return true
+		}
+	}
+	return false
+}
+
+var _ = Describe("netutil netlink function testing", func() {
```
maiqueb, line 136: Thanks for adding these unit tests.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+
+				//"dst": "10.0.0.0/16"
+				err = testAddRoute(link,
+					net.IPv4(10, 0, 0, 0), net.CIDRMask(16, 32),
+					net.IPv4(10, 0, 0, 1))
+				Expect(err).NotTo(HaveOccurred())
+
+				return nil
+			})
+			Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 200: Would it make sense to move all this test setup stage into a `BeforeEach` block ?

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
```
maiqueb, line 240: same here; I wonder if this shouldn't be moved to a `BeforeEach` block.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+	result, err := current.GetResult(resultRaw)
+	Expect(err).NotTo(HaveOccurred())
+
+	for _, r := range result.Routes {
+		if r.Dst.String() == "::/0" {
+			return true
+		}
+	}
+	return false
+}
```
maiqueb, line 134: plenty is duplicated here; you could export a 3rd function with most of the code and re-use it from the IP version wrappers:

```golang
func testResultHasIPv4DefaultRoute(data []byte) bool {
    return testResultHasDefaultRoute(data, "0.0.0.0/0")
}

func testResultHasIPv6DefaultRoute(data []byte) bool {
    return testResultHasDefaultRoute(data, "::0/0")
}

func testResultHasDefaultRoute(data []byte, defaultRouteStr string) bool {
	resultRaw, err := types020.NewResult(data)

	result, err := types020.GetResult(resultRaw)

	for _, r := range result.Routes {
		if r.Dst.String() == defaultRouteStr {
			return true
		}
	}
	return false
}
```

maiqueb (changes requested):
While I have added a lot of comments, I'm very happy with this fix.

The unit tests added are amazing.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package netutils
+
+import (
+	"encoding/json"
+	//"fmt"
```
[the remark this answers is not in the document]
s1061123: fixed.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+
+// helper function
+func testAddRoute(link netlink.Link, ip net.IP, mask net.IPMask, gw net.IP) error {
+    dst := &net.IPNet{
+        IP:   ip,
+        Mask: mask,
+    }
+    route := netlink.Route{LinkIndex: link.Attrs().Index, Dst: dst, Gw: gw}
+    err := netlink.RouteAdd(&route)
+    return err
```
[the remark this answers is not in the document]
s1061123: fixed

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+}
+
+func testAddAddr(link netlink.Link, ip net.IP, mask net.IPMask) error {
+    var address = &net.IPNet{IP: ip, Mask: mask}
+    var addr = &netlink.Addr{IPNet: address}
+    err := netlink.AddrAdd(link, addr)
+    return err
```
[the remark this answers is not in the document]
s1061123: fixed.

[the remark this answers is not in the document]
s1061123: fixed.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+        Mask: mask,
+    }
+    route := netlink.Route{LinkIndex: link.Attrs().Index, Dst: dst, Gw: gw}
+    err := netlink.RouteAdd(&route)
+    return err
+}
+
+func testAddAddr(link netlink.Link, ip net.IP, mask net.IPMask) error {
+    var address = &net.IPNet{IP: ip, Mask: mask}
+    var addr = &netlink.Addr{IPNet: address}
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+				err = testAddRoute(link,
+					net.IPv4(0, 0, 0, 0), net.CIDRMask(0, 0),
+					net.IPv4(10, 0, 0, 1))
+				Expect(err).NotTo(HaveOccurred())
+
+				//"dst": "10.0.0.0/16"
+				err = testAddRoute(link,
+					net.IPv4(10, 0, 0, 0), net.CIDRMask(16, 32),
+					net.IPv4(10, 0, 0, 1))
+				Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1042 @@
...
+        },
+        {
+          "dst": "30::0:0:0/64",
+          "gw": "10::1:1:1"
+        }
+      ]
+    }
+  }
+}`)
+			newResult, err := deleteDefaultGWCacheBytes(origResult, true, false)
```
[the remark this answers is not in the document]
s1061123: This is trade off among readability, efficiency and code complexity.

We could write with enum as follows but it has three state, so it is redundant. 

```
const (
   IPv4 = iota
   IPv6
   IPv4v6
```

We may use uint with bit calculation (0x1=IPv4, 0x10=IPv6), but it is not easy to read, too I guess.

We could also write as follows, but it is not only easy-to-read, also hard-to-write/debug, at least to me.

```
func deleteDefaultGWCacheBytes(cacheFile []byte, []func(*[]interface{}) error) ([]byte, error) {
...
}

func deleteDefaultGWCacheBytesCheckIPv4(routes *[]interface{}) error {
...
}

func deleteDefaultGWCacheBytesCheckIPv6(routes *[]interface{}) error {
...
}

newResult, err := deleteDefaultGWCacheBytes(origResult, [deleteDefaultGWCacheBytesCheckIPv4, deleteDefaultGWCacheBytesCheckIPv6])
```

So I keep it as it is for now. If you want to change, please file a new PR for that. I am happy to review that!

pkg/types/conf.go
```diff
@@ -504,13 +504,28 @@ func addCNIArgsInConfList(inBytes []byte, cniArgs *map[string]interface{}) ([]by
...
+			delegates[i].IsFilterV6Gateway = true
+		} else {
+			if len(delegate.GatewayRequest) == 1 {
+				// only 1 GW configrued (IPv4 or IPv6)
+				if delegate.GatewayRequest[0].To4() != nil {
+					delegates[i].IsFilterV6Gateway = true
+				} else {
+					delegates[i].IsFilterV4Gateway = true
+				}
+			} else {
```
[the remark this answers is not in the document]
s1061123: So what is the user expectation for two or more GW configured in same family? Currently CNI plugin and multus don't support ECMP yet, hence it should be error. Changed to raise error if multiple GWs configured in same family.

pkg/multus/multus.go
```diff
@@ -649,37 +650,67 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
+			// According to
+			// https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ (4.1.2.1.9)
+			// the list can be empty; if it is, we'll assume the CNI's config for the default gateway holds,
+			// else we'll update the defaultgateway to the one specified.
+			if delegate.GatewayRequest != nil && delegate.GatewayRequest[0] != nil {
+				deleteV6gateway = true
+				adddefaultgateway = true
+				logging.Debugf("Detected gateway override on interface %v to %v", ifName, delegate.GatewayRequest)
+			}
+		}
```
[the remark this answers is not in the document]
s1061123: I don't think so because both codes are a bit different. Should we write such like that?

```
func XXX(family string, delegate DelegateNetConf, filterGW bool, deleteGW *bool, adddefaultGW *bool) {
                if filterGW {
                        *deleteGW = true
                        logging.Debugf("Marked interface %v for %s gateway deletion", family, ifName)
                } else {
                        // Otherwise, determine if this interface now gets our default route.
                        // According to
                        // https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ (4.1.2.1.9)
                        // the list can be empty; if it is, we'll assume the CNI's config for the default gateway holds,
                        // else we'll update the defaultgateway to the one specified.
                        if delegate.GatewayRequest != nil && delegate.GatewayRequest[0] != nil {
                                *deleteGW = true
                                *adddefaultGW= true
                                logging.Debugf("Detected gateway override on interface %v to %v", ifName, delegate.GatewayRequest)
                        }
                }
```

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,350 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+	if !ok  {
+		// No route in result, hence we do nothing
+		return result, nil
+	}
+	routes, ok := result["routes"].([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("wrong routes format: %v", result["routes"])
+	}
+
+	newRoutes := routes
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,350 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+	// delete IPv4 default routes
+	if ipv4 {
+		for i, r := range routes {
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,350 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+				dst, ok := route["dst"].(string)
+				if !ok {
+					return nil, fmt.Errorf("wrong dst format: %v", route["dst"])
+				}
+				if dst == "::0/0" {
+					newRoutes = append(newRoutes[:i], newRoutes[i+1:]...)
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,350 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+						dst, ok := route["dst"].(string)
+						if !ok {
+							return nil, fmt.Errorf("wrong dst format: %v", route["dst"])
+						}
+						if dst == "::0/0" {
+							newRoutes = append(routes[:i], routes[i+1:]...)
+						}
+					}
+				}
+				ip6["routes"] = newRoutes
```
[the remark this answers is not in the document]
s1061123: changed.

pkg/netutils/netutils_test.go
s1061123: Your code above has less duplicated, however, your code size and my code size are pretty similar, so I suppose it is not a big issue.

pkg/netutils/netutils_test.go
s1061123: Each test setup code is different, hence if we cannot use `BeforeEach`. Is there any way to change inside `BeforeEach` for each testing? (To me, I suppose we don't just use `BeforeEach` in such case...)

pkg/netutils/netutils_test.go
s1061123: ditto.

pkg/netutils/netutils_test.go
s1061123: ditto. I suppose not big difference from code size point of view, so the refactoring is not strongly required.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1026 @@
...
+}
+
+func testAddAddr(link netlink.Link, ip net.IP, mask net.IPMask) error {
+    return netlink.AddrAdd(link, &netlink.Addr{IPNet: &net.IPNet{IP: ip, Mask: mask}})
+}
+
+func testGetResultFromCache(data []byte) []byte {
+	var cachedInfo map[string]interface{}
+	err := json.Unmarshal(data, &cachedInfo)
+	ExpectWithOffset(1,err).NotTo(HaveOccurred())
```
maiqueb, line 57: nit: you could inline these:
```golang
	ExpectWithOffset(1, json.Unmarshal(data, &cachedInfo)).NotTo(HaveOccurred())
```

I'd again use `Expect(...).To(Succeed())` rather than `Expect(...).NotTo(HaveOccurred())` but this is a personal opinion; you're free to Ignore both suggestions if you prefer.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1026 @@
...
+	})
+
+	Context("test DeleteDefaultGW", func() {
+		It("verify default gateway is removed", func() {
+			Expect(targetNS.Do(func(ns.NetNS) error {
+				defer GinkgoRecover()
+				link, err := netlink.LinkByName(IFNAME)
+				Expect(err).NotTo(HaveOccurred())
+				err = netlink.LinkSetUp(link)
+				Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 176: optional nit: these could be inlined. Sorry for having missed it in the first pass.

pkg/types/conf.go
```diff
@@ -504,15 +504,35 @@ func addCNIArgsInConfList(inBytes []byte, cniArgs *map[string]interface{}) ([]by
...
+		} else {
+			gateways := len(delegate.GatewayRequest)
+			if gateways == 1 {
+				// only 1 GW configrued (IPv4 or IPv6)
+				if delegate.GatewayRequest[0].To4() != nil {
+					delegates[i].IsFilterV6Gateway = true
+				} else {
+					delegates[i].IsFilterV4Gateway = true
+				}
+			} else if gateways == 2 {
```
maiqueb, line 525: What if a user requests 2 gateways, but for whatever reason provide 2 gateways for the same family ? Wouldn't that scenario "match" this if clause ?

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,318 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+	}
+
+	if cniVersion == "0.1.0" || cniVersion == "0.2.0" {
+		// fallback to processing result for old cni version(0.1.0/0.2.0)
+		return addDefaultGWResult020(result, gw)
+	}
+
+	if cniVersion != "0.3.0" && cniVersion != "0.3.1" && cniVersion != "0.4.0" && cniVersion != "1.0.0" {
+		return nil, fmt.Errorf("not supported version: %s", cniVersion)
+	}
```
maiqueb, line 334: As more CNI versions become available, we're "forced" to update our code to support that, right ?... 
I would prefer *a* way that does not require constant maintenance.

For instance, right now, you're missing release `1.0.1`: https://github.com/containernetworking/cni/releases/tag/v1.0.1

pkg/netutils/netutils.go
```diff
@@ -113,7 +89,318 @@ func SetDefaultGW(args *skel.CmdArgs, ifName string, gateways []net.IP, res *cni
...
+	routes := []interface{}{}
+	_, ok = result["routes"]
+	if ok  {
+		routes, ok = result["routes"].([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("wrong routes format: %v", result["routes"])
+		}
+	}
+
+	newRoutes := routes
```
maiqueb, line 345: I may be missing something, but I still don't see the purpose of this variable.

pkg/netutils/netutils.go
maiqueb: ping (just a suggestion you can ignore).

```golang
func AddDefaultGWCache(cacheDir string, rt *libcni.RuntimeConf, netName string, ifName string, gw []net.IP) error {
    cache, err := readCache(cacheDir, netName, rt.ContainerID, rt.IfName)
...
}

func readCache(dir string, networkName string, containerID string, ifaceName string) ([]byte, err) {
	return ioutil.ReadFile(
            filepath.Join(
                cacheDir, "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName)))
```

maiqueb (commented):
Thanks for the work, most of my follow up comments are just opinionated nits you can ignore if you prefer.

pkg/netutils/netutils.go
s1061123: Of course we need to update, because CNI project does not have unified way to convert the cache (well, actually CNI cache function is not exported to public, actually, from CNI point of view, we should not touch CNI cache...).

> For instance, right now, you're missing release 1.0.1: https://github.com/containernetworking/cni/releases/tag/v1.0.1
Not missing, because 1.0.1 does not change result schema, hence 1.0.1 uses still 1.0.0. See https://github.com/containernetworking/cni/blob/v1.0.1/pkg/types/100/types.go#L29

maiqueb: > Not missing, because 1.0.1 does not change result schema, hence 1.0.1 uses still 1.0.0. See https://github.com/containernetworking/cni/blob/v1.0.1/pkg/types/100/types.go#L29

Nice, wasn't aware. Thanks.

pkg/netutils/netutils_test.go
```diff
@@ -0,0 +1,1025 @@
...
+    return netlink.RouteAdd(&route)
+}
+
+func testAddAddr(link netlink.Link, ip net.IP, mask net.IPMask) error {
+    return netlink.AddrAdd(link, &netlink.Addr{IPNet: &net.IPNet{IP: ip, Mask: mask}})
+}
+
+func testGetResultFromCache(data []byte) []byte {
+	var cachedInfo map[string]interface{}
+	err := json.Unmarshal(data, &cachedInfo)
```
maiqueb, line 56: left-over. either delete it, or `ExpectWithOffset(1, err)...` below.

maiqueb (approved):
Thank you.

PR #768, opened 2021-12-20
Add toleration against NoExecute

oshothebig:
This PR adds toleration against NoExecute for `kube-multus-ds` DaemonSet. I think it's a common practice for CNI plugins to set tolerations against `NoSchedule` and `NoExecute` as well.

The final goal of this PR is to include the toleration in the release manifest like https://github.com/k8snetworkplumbingwg/multus-cni/blob/v3.8/images/multus-daemonset.yml, but I'm not so sure the files changed in this PR are right to do that. If my guess is wrong, could you let me know where the right location is?

### DIFF
--- a/deployments/multus-daemonset-crio.yml
+++ b/deployments/multus-daemonset-crio.yml
@@ -166,6 +166,8 @@ spec:
       tolerations:
       - operator: Exists
         effect: NoSchedule
+      - operator: Exists
+        effect: NoExecute
       serviceAccountName: multus
       containers:
       - name: kube-multus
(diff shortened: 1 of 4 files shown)


dougbtv:
Thanks for the PR! We took a look at the maintainer's meeting and we're going to think about the potential impact and we'll return to the topic next maintainer's meeting.

s1061123 (approved):
Thank you for the PR. It looks good to me, so I'll merge it soon.
Once the new version is released, this change will be in the release manifest you mentioned.

But please keep in mind that the daemonset is just a template for multus, for 'tasting', 'testing' and so on, hence we don't care about whole cluster life cycles for production environment. Hence you may need to customize this daemonset yaml for production environment, to inter-work with cluster network CNI plugins.

We will mention it explicitly in the issue, https://github.com/k8snetworkplumbingwg/multus-cni/issues/770.

PR #771, opened 2022-01-03
Refactor multus as a thick plugin

maiqueb:
This PR's scope is to refactor multus as a thick plugin.

This is a first required step to add / remove network interfaces to / from running PODs.

Follow-up PRs will build upon this, and implement the multus controller that will add / remove network interfaces to / from running pods.

**Info for reviewers**: 
~~This PR is still in draft mode because:~~
- ~~some legacy unit tests are failing. Ongoing investigation.~~
- ~~it misses unit tests for the added code~~
- the way the delegate CNIs are spawned **may** change in the future: mounting vs switching mount namespaces (accessing `/proc/<pid>/root/...` can also be an option - also requires sharing host's PID ns).

### DIFF
--- a/deployments/multus-daemonset-thick-plugin.yml
+++ b/deployments/multus-daemonset-thick-plugin.yml
@@ -114,6 +114,7 @@ spec:
         name: multus
     spec:
       hostNetwork: true
+      hostPID: true
       tolerations:
         - operator: Exists
           effect: NoSchedule
@@ -130,6 +131,8 @@ spec:
             - "-multus-autoconfig-dir=/host/etc/cni/net.d"
             - "-multus-log-to-stderr=true"
             - "-multus-log-level=verbose"
+            - "-binDir=/host/opt/cni/bin"
+            - "-cniDir=/host/var/lib/cni/multus"
           resources:
             requests:
               cpu: "100m"
@@ -144,6 +147,11 @@ spec:
               mountPath: /host/etc/cni/net.d
             - name: cnibin
               mountPath: /host/opt/cni/bin
+            - name: host-var-run-multus-cni
+              mountPath: /host/var/run/multus-cni
+            - name: host-var-run-netns
+              mountPath: /var/run/netns
+              mountPropagation: HostToContainer
       initContainers:
         - name: install-multus-binary
           image: ghcr.io/k8snetworkplumbingwg/multus-cni:thick
@@ -178,6 +186,21 @@ spec:
             - name: cni
               mountPath: /host/etc/cni/net.d
               mountPropagation: Bidirectional
+        - name: config-cni-socket-permissions
+            image: ghcr.io/k8snetworkplumbingwg/multus-cni:thick
+            command:
+              - "chmod"
+              - "0700"
+              - "/host/var/run/multus-cni"
+            resources:
+              requests:
+                cpu: "10m"
+                memory: "15Mi"
+            securityContext:
+              privileged: true
+            volumeMounts:
+              - name: host-var-run-multus-cni
+                mountPath: /host/var/run/multus-cni
       terminationGracePeriodSeconds: 10
       volumes:
         - name: cni
@@ -186,4 +209,9 @@ spec:
         - name: cnibin
           hostPath:
             path: /opt/cni/bin
-
+        - name: host-var-run-multus-cni
+          hostPath:
+            path: /var/run/multus-cni
+        - name: host-var-run-netns
+          hostPath:
+            path: /var/run/netns/
--- a/pkg/cni/server.go
+++ b/pkg/cni/server.go
@@ -16,7 +16,6 @@
 package cni
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -24,9 +23,9 @@ import (
 	"net/http"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/containernetworking/cni/pkg/invoke"
+	"github.com/containernetworking/cni/pkg/skel"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	cnicurrent "github.com/containernetworking/cni/pkg/types/current"
 	"github.com/gorilla/mux"
@@ -35,7 +34,6 @@ import (
 	k8s "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/k8sclient"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/multus"
-	multustypes "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/types"
 )
 
 // HandleCNIRequest is the CNI server handler function; it is invoked whenever
@@ -128,7 +126,6 @@ func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer req.cancel()
 
 	result, err := s.requestFunc(req)
 	if err != nil {
@@ -146,8 +143,8 @@ func cniRequestToPodRequest(cniRequest *Request, kubeclient *k8s.ClientInfo, exe
 
 	req := &PodRequest{
 		Command:    command(cmd),
-		kubeclient: kubeclient,
 		exec:       exec,
+		kubeClient: kubeclient,
 	}
 
 	req.ContainerID, ok = cniRequest.Env["CNI_CONTAINERID"]
@@ -164,36 +161,39 @@ func cniRequestToPodRequest(cniRequest *Request, kubeclient *k8s.ClientInfo, exe
 		req.IfName = "eth0"
 	}
 
-	cniArgs, err := gatherCNIArgs(cniRequest.Env)
+	cniArgs, found := cniRequest.Env["CNI_ARGS"]
+	if !found {
+		return nil, fmt.Errorf("missing CNI_ARGS")
+	}
+	req.Args = cniArgs
+	req.StdinData = cniRequest.Config
+
+	cniEnv, err := gatherCNIArgs(cniRequest.Env)
 	if err != nil {
 		return nil, err
 	}
-
-	req.Namespace, ok = cniArgs["K8S_POD_NAMESPACE"]
-	if !ok {
+	podNamespace, found := cniEnv["K8S_POD_NAMESPACE"]
+	if !found {
 		return nil, fmt.Errorf("missing K8S_POD_NAMESPACE")
 	}
 
-	req.Name, ok = cniArgs["K8S_POD_NAME"]
-	if !ok {
+	podName, found := cniEnv["K8S_POD_NAME"]
+	if !found {
 		return nil, fmt.Errorf("missing K8S_POD_NAME")
 	}
 
-	req.UID, err = podUID(kubeclient, cniArgs, req.Namespace, req.Name)
+	uid, err := podUID(kubeclient, cniEnv, podNamespace, podName)
 	if err != nil {
 		return nil, err
 	}
 
-	req.SandboxID = cniRequest.Env["K8S_POD_INFRA_CONTAINER_ID"]
+	sandboxID := cniRequest.Env["K8S_POD_INFRA_CONTAINER_ID"]
 
-	conf, err := multustypes.LoadNetConf(cniRequest.Config)
-	if err != nil {
-		return nil, fmt.Errorf("broken stdin args")
-	}
+	req.K8S_POD_NAMESPACE = cnitypes.UnmarshallableString(podNamespace)
+	req.K8S_POD_NAME = cnitypes.UnmarshallableString(podName)
+	req.K8S_POD_INFRA_CONTAINER_ID = cnitypes.UnmarshallableString(sandboxID)
+	req.K8S_POD_UID = cnitypes.UnmarshallableString(uid)
 
-	req.CNIConf = conf
-	req.timestamp = time.Now()
-	req.ctx, req.cancel = context.WithTimeout(context.Background(), time.Minute)
 	return req, nil
 }
 
@@ -233,72 +233,52 @@ func podUID(kubeclient *k8s.ClientInfo, cniArgs map[string]string, podNamespace,
 	return uid, nil
 }
 
+func (pr *PodRequest) cniSkel() *skel.CmdArgs {
+	return &skel.CmdArgs{
+		ContainerID: pr.ContainerID,
+		Netns:       pr.Netns,
+		IfName:      pr.IfName,
+		Args:        pr.Args,
+		Path:        pr.Path,
+		StdinData:   pr.StdinData,
+	}
+}
+
 func (pr *PodRequest) cmdAdd() ([]byte, error) {
-	namespace := pr.Namespace
-	podName := pr.Name
+	namespace := string(pr.K8S_POD_NAMESPACE)
+	podName := string(pr.K8S_POD_NAME)
 	if namespace == "" || podName == "" {
-		return nil, fmt.Errorf("required CNI variable missing")
+		return nil, fmt.Errorf("required CNI variable missing. pod name: %s; pod namespace: %s", podName, namespace)
 	}
 
-	pod, err := multus.GetPod(pr.kubeclient, pr.Namespace, pr.Name, pr.UID, false)
+	logging.Debugf("CmdAdd for [%s/%s]. CNI conf: %+v", namespace, podName, *pr)
+	result, err := multus.CmdAdd(pr.cniSkel(), pr.exec, pr.kubeClient)
 	if err != nil {
-		return nil, fmt.Errorf("error getting pod [%s/%s]: %v", pr.Namespace, pr.Name, err)
-	}
-
-	multusAddCmd := multus.NewCmd(
-		pr.ContainerID,
-		pr.SandboxID,
-		pr.IfName,
-		pr.Netns,
-		pr.Name,
-		pr.Namespace,
-		pr.UID,
-	)
-	result, err := multus.Add(multusAddCmd, pr.CNIConf, pod, pr.exec, pr.kubeclient)
-	if err != nil {
-		return nil, fmt.Errorf("error configuring pod [%s/%s] networking: %v", pr.Namespace, pr.Name, err)
+		return nil, fmt.Errorf("error configuring pod [%s/%s] networking: %v", namespace, podName, err)
 	}
 	return serializeResult(result)
 }
 
 func (pr *PodRequest) cmdDelete() error {
-	logging.Debugf("CmdDel: %+v", *pr.CNIConf)
-
-	pod, err := multus.GetPod(pr.kubeclient, pr.Namespace, pr.Name, pr.UID, true)
-	if err != nil {
-		return fmt.Errorf("error getting pod [%s/%s]: %v", pr.Namespace, pr.Name, err)
+	namespace := string(pr.K8S_POD_NAMESPACE)
+	podName := string(pr.K8S_POD_NAME)
+	if namespace == "" || podName == "" {
+		return fmt.Errorf("required CNI variable missing. pod name: %s; pod namespace: %s", podName, namespace)
 	}
 
-	multusDeleteCmd := multus.NewCmd(
-		pr.ContainerID,
-		pr.SandboxID,
-		pr.IfName,
-		pr.Netns,
-		pr.Name,
-		pr.Namespace,
-		pr.UID,
-	)
-	return multus.Delete(multusDeleteCmd, pr.CNIConf, pod, pr.exec, pr.kubeclient)
+	logging.Debugf("CmdDel for [%s/%s]. CNI conf: %+v", namespace, podName, *pr)
+	return multus.CmdDel(pr.cniSkel(), pr.exec, pr.kubeClient)
 }
 
 func (pr *PodRequest) cmdCheck() error {
-	namespace := pr.Namespace
-	podName := pr.Name
+	namespace := string(pr.K8S_POD_NAMESPACE)
+	podName := string(pr.K8S_POD_NAME)
 	if namespace == "" || podName == "" {
-		return fmt.Errorf("required CNI variable missing")
+		return fmt.Errorf("required CNI variable missing. pod name: %s; pod namespace: %s", podName, namespace)
 	}
 
-	logging.Debugf("CmdCheck for [%s/%s]. CNI conf: %+v", namespace, podName, *pr.CNIConf)
-
-	multusCheckCmd := multus.NewCmd(
-		pr.ContainerID,
-		pr.SandboxID,
-		pr.IfName,
-		pr.Netns,
-		pr.Name,
-		pr.Namespace,
-		pr.UID)
-	return multus.Check(multusCheckCmd, pr.CNIConf, pr.exec)
+	logging.Debugf("CmdCheck for [%s/%s]. CNI conf: %+v", namespace, podName, *pr)
+	return multus.CmdCheck(pr.cniSkel(), pr.exec, pr.kubeClient)
 }
 
 func serializeResult(result cnitypes.Result) ([]byte, error) {
--- a/pkg/cni/server.go
+++ b/pkg/cni/server.go
@@ -1,0 +1,312 @@
+// Copyright (c) 2021 Multus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cni
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/containernetworking/cni/pkg/invoke"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
+	cnicurrent "github.com/containernetworking/cni/pkg/types/current"
+	"github.com/gorilla/mux"
+
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/config"
+	k8s "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/k8sclient"
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/multus"
+	multustypes "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/types"
+)
+
+// HandleCNIRequest is the CNI server handler function; it is invoked whenever
+// a CNI request is processed.
+func HandleCNIRequest(request *PodRequest) ([]byte, error) {
+	var result []byte
+	var err error
+
+	logging.Verbosef("%+v %s starting CNI request %+v", request, request.Command, request)
+	switch request.Command {
+	case "ADD":
+		result, err = request.cmdAdd()
+	case "DEL":
+		err = request.cmdDelete()
+	case "CHECK":
+		err = request.cmdCheck()
+	default:
+	}
+	logging.Verbosef("%s finished CNI request %+v, result: %q, err: %v", request.Command, request, string(result), err)
+	if err != nil {
+		// Prefix errors with request info for easier failure debugging
+		return nil, fmt.Errorf("%+v ERRORED: %v", request, err)
+	}
+	return result, nil
+}
+
+// ServerListener creates a listener to a unix socket located in `socketPath`
+func ServerListener(socketPath string) (net.Listener, error) {
+	l, err := net.Listen("unix", socketPath)
+	if err != nil {
+		return nil, logging.Errorf("failed to listen on pod info socket: %v", err)
+	}
+	if err := os.Chmod(socketPath, config.UserRWPermission); err != nil {
+		_ = l.Close()
+		return nil, logging.Errorf("failed to listen on pod info socket: %v", err)
+	}
+	return l, nil
+}
+
+// NewCNIServer creates and returns a new Server object which will listen on a socket in the given path
+func NewCNIServer(rundir string) (*Server, error) {
+	kubeClient, err := k8s.InClusterK8sClient()
+	if err != nil {
+		return nil, fmt.Errorf("error getting k8s client: %v", err)
+	}
+
+	return newCNIServer(rundir, kubeClient, nil)
+}
+
+func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec) (*Server, error) {
+	router := mux.NewRouter()
+	s := &Server{
+		Server: http.Server{
+			Handler: router,
+		},
+		rundir:      rundir,
+		requestFunc: HandleCNIRequest,
+		kubeclient:  kubeClient,
+		exec:        exec,
+	}
+
+	router.NotFoundHandler = http.HandlerFunc(http.NotFound)
+	router.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		result, err := s.handleCNIRequest(r)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("%v", err), http.StatusBadRequest)
+			return
+		}
+
+		// Empty response JSON means success with no body
+		w.Header().Set("Content-Type", "application/json")
+		if _, err := w.Write(result); err != nil {
+			_ = logging.Errorf("Error writing HTTP response: %v", err)
+		}
+	}).Methods("POST")
+
+	return s, nil
+}
+
+func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
+	var cr Request
+	b, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(b, &cr); err != nil {
+		return nil, err
+	}
+	req, err := cniRequestToPodRequest(&cr, s.kubeclient, s.exec)
+	if err != nil {
+		return nil, err
+	}
+	defer req.cancel()
+
+	result, err := s.requestFunc(req)
+	if err != nil {
+		// Prefix error with request information for easier debugging
+		return nil, fmt.Errorf("%+v %v", req, err)
+	}
+	return result, nil
+}
+
+func cniRequestToPodRequest(cniRequest *Request, kubeclient *k8s.ClientInfo, exec invoke.Exec) (*PodRequest, error) {
+	cmd, ok := cniRequest.Env["CNI_COMMAND"]
+	if !ok {
+		return nil, fmt.Errorf("unexpected or missing CNI_COMMAND")
+	}
+
+	req := &PodRequest{
+		Command:    command(cmd),
+		kubeclient: kubeclient,
+		exec:       exec,
+	}
+
+	req.ContainerID, ok = cniRequest.Env["CNI_CONTAINERID"]
+	if !ok {
+		return nil, fmt.Errorf("missing CNI_CONTAINERID")
+	}
+	req.Netns, ok = cniRequest.Env["CNI_NETNS"]
+	if !ok {
+		return nil, fmt.Errorf("missing CNI_NETNS")
+	}
+
+	req.IfName, ok = cniRequest.Env["CNI_IFNAME"]
+	if !ok {
+		req.IfName = "eth0"
+	}
+
+	cniArgs, err := gatherCNIArgs(cniRequest.Env)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Namespace, ok = cniArgs["K8S_POD_NAMESPACE"]
+	if !ok {
+		return nil, fmt.Errorf("missing K8S_POD_NAMESPACE")
+	}
+
+	req.Name, ok = cniArgs["K8S_POD_NAME"]
+	if !ok {
+		return nil, fmt.Errorf("missing K8S_POD_NAME")
+	}
+
+	req.UID, err = podUID(kubeclient, cniArgs, req.Namespace, req.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	req.SandboxID = cniRequest.Env["K8S_POD_INFRA_CONTAINER_ID"]
+
+	conf, err := multustypes.LoadNetConf(cniRequest.Config)
+	if err != nil {
+		return nil, fmt.Errorf("broken stdin args")
+	}
+
+	req.CNIConf = conf
+	req.timestamp = time.Now()
+	req.ctx, req.cancel = context.WithTimeout(context.Background(), time.Minute)
+	return req, nil
+}
+
+func gatherCNIArgs(env map[string]string) (map[string]string, error) {
+	cniArgs, ok := env["CNI_ARGS"]
+	if !ok {
+		return nil, fmt.Errorf("missing CNI_ARGS: '%s'", env)
+	}
+
+	mapArgs := make(map[string]string)
+	for _, arg := range strings.Split(cniArgs, ";") {
+		parts := strings.Split(arg, "=")
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid CNI_ARG '%s'", arg)
+		}
+		mapArgs[strings.TrimSpace(parts[0])] = strings.TrimSpace(parts[1])
+	}
+	return mapArgs, nil
+}
+func podUID(kubeclient *k8s.ClientInfo, cniArgs map[string]string, podNamespace, podName string) (string, error) {
+	// UID may not be passed by all runtimes yet. Will be passed
+	// by CRIO 1.20+ and containerd 1.5+ soon.
+	// CRIO 1.20: https://github.com/cri-o/cri-o/pull/5029
+	// CRIO 1.21: https://github.com/cri-o/cri-o/pull/5028
+	// CRIO 1.22: https://github.com/cri-o/cri-o/pull/5026
+	// containerd 1.6: https://github.com/containerd/containerd/pull/5640
+	// containerd 1.5: https://github.com/containerd/containerd/pull/5643
+	uid, found := cniArgs["K8S_POD_UID"]
+	if !found {
+		pod, err := kubeclient.GetPod(podNamespace, podName)
+		if err != nil {
+			return "", fmt.Errorf("missing pod UID; attempted to recover it from the K8s API, but failed: %w", err)
+		}
+		return string(pod.UID), nil
+	}
+
+	return uid, nil
+}
+
+func (pr *PodRequest) cmdAdd() ([]byte, error) {
+	namespace := pr.Namespace
+	podName := pr.Name
+	if namespace == "" || podName == "" {
+		return nil, fmt.Errorf("required CNI variable missing")
+	}
+
+	pod, err := multus.GetPod(pr.kubeclient, pr.Namespace, pr.Name, pr.UID, false)
+	if err != nil {
+		return nil, fmt.Errorf("error getting pod [%s/%s]: %v", pr.Namespace, pr.Name, err)
+	}
+
+	result, err := multus.NewMultusCmd(
+		pr.ContainerID,
+		pr.SandboxID,
+		pr.IfName,
+		pr.Netns,
+		pr.Name,
+		pr.Namespace,
+		pr.UID,
+	).Add(pr.CNIConf, pod, pr.exec, pr.kubeclient)
+	if err != nil {
+		return nil, fmt.Errorf("error configuring pod [%s/%s] networking: %v", pr.Namespace, pr.Name, err)
+	}
+	return serializeResult(result)
+}
+
+func (pr *PodRequest) cmdDelete() error {
+	logging.Debugf("CmdDel: %+v", *pr.CNIConf)
+
+	pod, err := multus.GetPod(pr.kubeclient, pr.Namespace, pr.Name, pr.UID, true)
+	if err != nil {
+		return fmt.Errorf("error getting pod [%s/%s]: %v", pr.Namespace, pr.Name, err)
+	}
+
+	return multus.NewMultusCmd(
+		pr.ContainerID,
+		pr.SandboxID,
+		pr.IfName,
+		pr.Netns,
+		pr.Name,
+		pr.Namespace,
+		pr.UID,
+	).Delete(pr.CNIConf, pod, pr.exec, pr.kubeclient)
+}
+
+func (pr *PodRequest) cmdCheck() error {
+	namespace := pr.Namespace
+	podName := pr.Name
+	if namespace == "" || podName == "" {
+		return fmt.Errorf("required CNI variable missing")
+	}
+
+	logging.Debugf("CmdCheck for [%s/%s]. CNI conf: %+v", namespace, podName, *pr.CNIConf)
+
+	return multus.NewMultusCmd(
+		pr.ContainerID,
+		pr.SandboxID,
+		pr.IfName,
+		pr.Netns,
+		pr.Name,
+		pr.Namespace,
+		pr.UID).Check(pr.CNIConf, pr.exec)
+}
+
+func serializeResult(result cnitypes.Result) ([]byte, error) {
+	realResult, err := cnicurrent.NewResultFromResult(result)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate the CNI result: %w", err)
+	}
+
+	responseBytes, err := json.Marshal(&Response{Result: realResult})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal pod request response: %v", err)
+	}
+	return responseBytes, nil
+}
--- a/pkg/cni/server.go
+++ b/pkg/cni/server.go
@@ -34,28 +34,30 @@ import (
 	k8s "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/k8sclient"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/multus"
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/types"
 )
 
 // HandleCNIRequest is the CNI server handler function; it is invoked whenever
 // a CNI request is processed.
-func HandleCNIRequest(request *PodRequest) ([]byte, error) {
+func HandleCNIRequest(cmd string, k8sArgs *types.K8sArgs, cniCmdArgs *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) ([]byte, error) {
 	var result []byte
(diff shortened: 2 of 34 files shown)


pkg/cni/socket.go
```diff
@@ -0,0 +1,69 @@
...
+func FilesystemPreRequirements(rundir string) error {
+	socketpath := SocketPath(rundir)
+	if err := os.RemoveAll(rundir); err != nil && !os.IsNotExist(err) {
+		info, err := os.Stat(rundir)
+		if err != nil {
+			return fmt.Errorf("failed to stat old pod info socket directory %s: %v", rundir, err)
+		}
+		// Owner must be root
+		tmp := info.Sys()
+		statt, ok := tmp.(*syscall.Stat_t)
```
nicklesimba, line 42: I will post questions as comments here whenever I do not understand something. Let me know if I should be directing these questions elsewhere. :)

Is this code guaranteed to be running on a Unix environment? Where can I find that info?

maiqueb: > I will post questions as comments whenever I do not understand something. It isn't necessarily code review, so let me know if I should be directing these questions elsewhere. :)
> 
> Is this code guaranteed to be running on a Unix environment? Where can I find that info?

AFAIU yes - at least we are using netlink, which only runs on linux.

I have no idea if / where the multus project has such documentation. @dougbtv please keep me honest.

pkg/cni/types.go
```diff
@@ -0,0 +1,104 @@
...
+// PodRequest represents a request for ADD / DEL / CHECK for a Pod
+type PodRequest struct {
+	// The CNI command of the operation
+	Command command
+	// kubernetes namespace name
+	PodNamespace string
+	// kubernetes pod name
+	PodName string
+	// kubernetes pod UID
+	PodUID string
```
s1061123, line 28: These variable name could be 'Namespace', 'Name' and 'UID'. Simply remove 'Pod' because this structure is for Pod.

pkg/cni/types.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Server represents an HTTP server listening to a unix socket. It will handle
+// the CNI shim requests issued when a pod is added / removed.
+type Server struct {
+	http.Server
+	requestFunc       cniRequestFunc
+	rundir            string
+	useOVSExternalIDs int32
+	kubeAuth          *KubeAPIAuth
```
s1061123, line 78: `kubeAuth` may not required to keep in Server structure. Usually user gives kubeconfig (from service account), so it should be enough.

pkg/cni/types.go
```diff
@@ -0,0 +1,104 @@
...
+// Request sent to the Server by the multus-shim
+type Request struct {
+	// CNI environment variables, like CNI_COMMAND and CNI_NETNS
+	Env map[string]string `json:"env,omitempty"`
+	// CNI configuration passed via stdin to the CNI plugin
+	Config []byte `json:"config,omitempty"`
+}
+
+// KubeAPIAuth contains information necessary to create a Kube API client
+type KubeAPIAuth struct {
```
s1061123, line 60: `KubeAPIAuth` may not be required. as following reason (see line 78 comment)

pkg/cni/types.go
```diff
@@ -0,0 +1,104 @@
...
+const CNIUpdate command = "UPDATE"
+
+// CNIDel is the command representing delete operation on a pod that is to be torn down
+const CNIDel command = "DEL"
+
+// CNICheck is the command representing check operation on a pod
+const CNICheck command = "CHECK"
+
+// Plugin represents the connection between the CNI shim and server.
+type Plugin struct {
```
s1061123, line 96: Why do you create structure for that? Is there any detailed plan to enhance it?

pkg/cni/types.go
```diff
@@ -0,0 +1,104 @@
...
+const CNIAdd command = "ADD"
+
+// CNIUpdate is the command representing update operation for an existing pod
+const CNIUpdate command = "UPDATE"
+
+// CNIDel is the command representing delete operation on a pod that is to be torn down
+const CNIDel command = "DEL"
+
+// CNICheck is the command representing check operation on a pod
+const CNICheck command = "CHECK"
```
s1061123, line 93: I suppose `const CNIAdd` seems to be redundant. We just use 'ADD' string, not make it const variable.

pkg/k8sclient/k8sclient.go
```diff
@@ -96,8 +96,17 @@ func (e *NoK8sNetworkError) Error() string { return string(e.message) }
...
+	podUID := string(k8sArgs.K8S_POD_UID)
+
+	return SetPodNetworkStatusAnnotation(client, podName, podNamespace, podUID, netStatus, conf)
+}
+
+// SetPodNetworkStatusAnnotation sets network status into Pod annotation
+func SetPodNetworkStatusAnnotation(client *ClientInfo, podName string, podNamespace string, podUID string, netStatus []nettypes.NetworkStatus, conf *types.NetConf) error {
 	var err error
-	logging.Debugf("SetNetworkStatus: %v, %v, %v, %v", client, k8sArgs, netStatus, conf)
+	logging.Debugf("SetNetworkStatus: %v, %v, %v", client, netStatus, conf)
```
s1061123, line 109: SetNetworkStatus -> SetNetworkStatusAnnotation

pkg/k8sclient/k8sclient.go
```diff
@@ -96,8 +96,17 @@ func (e *NoK8sNetworkError) Error() string { return string(e.message) }
 
 // SetNetworkStatus sets network status into Pod annotation
 func SetNetworkStatus(client *ClientInfo, k8sArgs *types.K8sArgs, netStatus []nettypes.NetworkStatus, conf *types.NetConf) error {
```
s1061123, line 98: I don't know why you make into two function SetNetworkStatus and SetNetworkStatusAnnotation. I believe SetNetworkStatus/SetNetworkStatusAnnotation is not so much used (less than 10 times), so I suppose we just make one function.

pkg/types/conf.go
```diff
@@ -184,41 +184,24 @@ func mergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetCo
 // CreateCNIRuntimeConf create CNI RuntimeConf for a delegate. If delegate configuration
 // exists, merge data with the runtime config.
 func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, rc *RuntimeConfig, delegate *DelegateNetConf) (*libcni.RuntimeConf, string) {
```
s1061123, line 186: Why you make into two function: CreateCNIRuntimeConf and CNIRuntimeConf?

To me, CreateCNIRuntimeConf seems to be redundant because it is just a one line.

```
func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, rc *RuntimeConfig, delegate *DelegateNetConf) (*libcni.RuntimeConf, string) {
	return CNIRuntimeConf(args.ContainerID, string(k8sArgs.K8S_POD_INFRA_CONTAINER_ID), string(k8sArgs.K8S_POD_NAME),  string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_UID), args.Netns, ifName, rc, delegate)
}
```

pkg/types/conf.go
```diff
@@ -184,41 +184,24 @@ func mergeCNIRuntimeConfig(runtimeConfig *RuntimeConfig, delegate *DelegateNetCo
...
-			autoDeviceInfo := fmt.Sprintf("%s-%s_%s", delegate.Name, args.ContainerID, ifName)
-			delegateRc.CNIDeviceInfoFile = nadutils.GetCNIDeviceInfoPath(autoDeviceInfo)
-			cniDeviceInfoFile = delegateRc.CNIDeviceInfoFile
-			logging.Debugf("Adding auto-generated CNIDeviceInfoFile: %s", delegateRc.CNIDeviceInfoFile)
-		}
-	} else {
-		delegateRc = rc
-	}
+// CNIRuntimeConf creates the CNI `RuntimeConf` for the given ADD / DEL request.
+func CNIRuntimeConf(containerID, sandboxID, podName, podNamespace, podUID, netNs, ifName string, rc *RuntimeConfig, delegate *DelegateNetConf) (*libcni.RuntimeConf, string) {
```
s1061123, line 195: As the function name, CNIRuntimeConf is not clear. As above comment, It should be CreateCNIRuntimeConf() because this function creates CNI RuntimeConf.

pkg/types/conf.go
```diff
@@ -246,6 +229,43 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 	return rt, cniDeviceInfoFile
 }
 
+// RuntimeConf creates the CNI `RuntimeConf` for the given ADD / DEL request.
+func RuntimeConf(netNs, podNamespace, podName, containerID, sandboxID, podUID, ifName string) *libcni.RuntimeConf {
```
s1061123, line 233: Ditto. See comment in CNIRuntimeConf.

pkg/multus/multus.go
```diff
@@ -458,19 +458,34 @@ func delegateDel(exec invoke.Exec, pod *v1.Pod, ifName string, delegateConf *typ
 	return err
 }
 
+// CNICmd holds the required CNI spec data required to configure networking in
```
s1061123, line 461: CNICmd seems to be CNI command, 'ADD', 'DEL', 'CHECK'. How about to change it to CNIRuntimeArgs or MultusCNIArgs?

pkg/multus/multus.go
```diff
@@ -583,50 +606,148 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
+		InterfaceName: ifaceName,
+		NetnsPath:     netnsPath,
+		PodName:       podName,
+		PodNamespace:  podNamespace,
+		PodUID:        podUID,
+	}
+}
+
+// CmdAdd is the handler of the CNI ADD command
+func (mc *CNICmd) CmdAdd(cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) (cnitypes.Result, error) {
```
s1061123, line 708: To me, `mc *CNICmd` seems to be part of args, not method variable.

pkg/multus/multus.go
```diff
@@ -744,65 +865,62 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
+}
 
-	for idx, delegate := range in.Delegates {
-		ifName := getIfname(delegate, args.IfName, idx)
-
-		rt, _ := types.CreateCNIRuntimeConf(args, k8sArgs, ifName, in.RuntimeConfig, delegate)
-		err = delegateCheck(exec, ifName, delegate, rt, in)
-		if err != nil {
+// CmdCheck is the handler of the CNI CHECK command
+func (mc *CNICmd) CmdCheck(cniConf *types.NetConf, exec invoke.Exec) error {
```
s1061123, line 891: To me, `mc *CNICmd` seems to be part of args, not method variable.

pkg/multus/multus.go
```diff
@@ -744,65 +865,62 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
-//CmdDel ...
-func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) error {
-	in, err := types.LoadNetConf(args.StdinData)
-	logging.Debugf("CmdDel: %v, %v, %v", args, exec, kubeClient)
-	if err != nil {
-		return err
-	}
-
+// CmdDelete is the handler of the CNI DEL command
+func (mc *CNICmd) CmdDelete(cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) error {
```
s1061123, line 913: To me, `mc *CNICmd` seems to be part of args, not method variable.
We could keep the function name as it is. Is there any clear reason why we need to change the function name?

pkg/multus/multus.go
```diff
@@ -812,51 +930,26 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
...
-
-	kubeClient, err = k8s.GetK8sClient(in.Kubeconfig, kubeClient)
-	if err != nil {
-		return cmdErr(nil, "error getting k8s client: %v", err)
-	}
-
-	pod, err := getPod(kubeClient, k8sArgs, true)
-	if err != nil {
-		return err
-	}
```
s1061123, line 838: Why do you remove these code from here?

s1061123 (changes requested):

pkg/cni/server.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// UID may not be passed by all runtimes yet. Will be passed
+	// by CRIO 1.20+ and containerd 1.5+ soon.
+	// CRIO 1.20: https://github.com/cri-o/cri-o/pull/5029
+	// CRIO 1.21: https://github.com/cri-o/cri-o/pull/5028
+	// CRIO 1.22: https://github.com/cri-o/cri-o/pull/5026
+	// containerd 1.6: https://github.com/containerd/containerd/pull/5640
+	// containerd 1.5: https://github.com/containerd/containerd/pull/5643
+	req.PodUID = cniArgs["K8S_POD_UID"]
```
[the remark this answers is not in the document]
maiqueb: Makes sense. Most common scenario is for it to be present, and that'll be the cheapest; if not found, we'll use the client to query the API server and try to get the value. 

Very nice suggestion.

pkg/k8sclient/k8sclient.go
maiqueb: One of them uses `k8sArgs`, the other doesn't. The one that uses k8s args calls the other one, passing the explicit values.

I do this in order to keep the current caller code (in `multus.go`) untouched.

pkg/types/conf.go
maiqueb: same as in https://github.com/k8snetworkplumbingwg/multus-cni/pull/771#discussion_r782236223

pkg/types/conf.go
maiqueb: I can do it like that; was just trying to follow [idiomatic go for getters](https://go.dev/doc/effective_go#Getters).

Forced to agree that this is not a getter - it creates an object. Will rename.

pkg/multus/multus.go
maiqueb: I don't follow this comment. I'm making `mc` the receiver for this function, which makes it a method.

The overall idea is for both the "old" binary (current one) and the new multus daemon should create a `CNICmd` with the required immutable data (args / cmd / netns / container id / ...) and just call the respective methods.

What do you mean w/ "be part of args" ?

pkg/multus/multus.go
maiqueb: same as in https://github.com/k8snetworkplumbingwg/multus-cni/pull/771#discussion_r782252596

pkg/multus/multus.go
maiqueb: same as in https://github.com/k8snetworkplumbingwg/multus-cni/pull/771#discussion_r782252596

I can preserve the function name if you prefer.

pkg/multus/multus.go
maiqueb: I see your point. I can rename the struct accordingly.

I guess `MultusCNIArgs` makes more sense to me (in my own personal mental model).

pkg/multus/multus.go
s1061123: Before the PR, it is function. Now it is method, even though you can change it as following:

```
func CmdAdd(mc *CNICmd, cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) (cnitypes.Result, error) {
```

pkg/cni/types.go
maiqueb: I think you're right; this can be dropped. I have no plans to extend this struct.

pkg/cni/types.go
maiqueb: I guess I'd prefer to have more uses of the CNIAdd const rather than call `"ADD"`; DRY principle. 

But I'll follow your suggestion if you insist.

pkg/multus/multus.go
maiqueb: I think you're right; I accidentally delete this for `CNI_DEL`. Will put it back.

pkg/cni/server.go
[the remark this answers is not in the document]
maiqueb: Done.

pkg/k8sclient/k8sclient.go
maiqueb: Done.

maiqueb (commented):
Reworked the PR + replied to @s1061123 's questions.

Thanks again for the review.

pkg/cni/types.go
s1061123: But to prevent to repeat, adding more code is not make sense to me.
"ADD" is 5 chars, CNIAdd is 6 chars. In this case, I suppose we could repeat it.

pkg/cni/types.go
s1061123: > Using the encapsulated socket path seems simpler, more readable, and uses less boilerplate code.

I don't think so. Currently functions, CmdAdd/CmdDel/CmdCheck/DoCNI, does not refer `p *Plugin` often, actually.

CmdAdd/CmdDel/CmdCheck: refer once (p.DoCNI)
DoCNI: refer once (p.SocketPath)

So I suppose we could decapsulate it and pass socket as argument to CmdAdd/Del/Check/DoCNI.

maiqueb: It's true that its not used often, but being able to simplify 4 calls is better than not doing it.

Furthermore, the caller code (located on the shim) below is simpler (and less):
```golang 
	p := cni.Plugin{SocketPath: cni.SocketPath(*runDir)}
	skel.PluginMain(
		p.CmdAdd,
		p.CmdCheck,
		p.CmdDel,
		cniversion.All, "meta-plugin that delegates to other CNI plugins")
```

Than what is required following your suggestion:
```golang
	skel.PluginMain(
		func(args *skel.CmdArgs) error {
			result, err := cni.CmdAdd(cni.SocketPath(*runDir))
			if err != nil {
				return err
			}
			return result.Print()
		},
		func(args *skel.CmdArgs) error {
			return cni.CmdCheck(cni.SocketPath(*runDir))
		},
		func(args *skel.CmdArgs) error { return cni.CmdDel(cni.SocketPath(*runDir)) },
		cniversion.All, "meta-plugin that delegates to other CNI plugins")
```

pkg/cni/types.go
maiqueb: I don't think the important parameter is the number of characters - I just think that using magic strings is worse than using a const, since I can update the value of the const once rather than searching the code N times for the repetitions of the magic string. 

I'm aware it isn't probably that this string will ever be updated, but still, as a practice, I'd take a const over magic strings every time.

I'll remove the consts and add more magic strings, since you insist.

maiqueb: Done.

pkg/multus/multus.go
```diff
@@ -583,50 +606,148 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
+		InterfaceName: ifaceName,
+		NetnsPath:     netnsPath,
+		PodName:       podName,
+		PodNamespace:  podNamespace,
+		PodUID:        podUID,
+	}
+}
+
+// Add is the handler of the CNI ADD command
+func (mc *Cmd) Add(cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) (cnitypes.Result, error) {
```
s1061123, line 708: Could you please change it to `Add(mc *Cmd, cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) (cnitypes.Result, error) {`?

pkg/multus/multus.go
```diff
@@ -744,119 +865,102 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
+}
 
-	for idx, delegate := range in.Delegates {
-		ifName := getIfname(delegate, args.IfName, idx)
-
-		rt, _ := types.CreateCNIRuntimeConf(args, k8sArgs, ifName, in.RuntimeConfig, delegate)
-		err = delegateCheck(exec, ifName, delegate, rt, in)
-		if err != nil {
+// Check is the handler of the CNI CHECK command
+func (mc *Cmd) Check(cniConf *types.NetConf, exec invoke.Exec) error {
```
s1061123, line 891: ditto as Add()

pkg/multus/multus.go
```diff
@@ -744,119 +865,102 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
 }
 
-//CmdDel ...
-func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) error {
-	in, err := types.LoadNetConf(args.StdinData)
-	logging.Debugf("CmdDel: %v, %v, %v", args, exec, kubeClient)
-	if err != nil {
-		return err
+// Delete is the handler of the CNI DEL command
+func (mc *Cmd) Delete(cniConf *types.NetConf, pod *v1.Pod, exec invoke.Exec, kubeclient *k8s.ClientInfo) error {
```
s1061123, line 913: ditto as Add()

s1061123 (changes requested):

pkg/multus/multus.go
```diff
@@ -458,19 +458,34 @@ func delegateDel(exec invoke.Exec, pod *v1.Pod, ifName string, delegateConf *typ
...
+	SandboxID     string
+}
+
 // delPlugins deletes plugins in reverse order from lastdIdx
 // Uses netRt as base RuntimeConf (coming from NetConf) but merges it
 // with each of the delegates' configuration
-func delPlugins(exec invoke.Exec, pod *v1.Pod, args *skel.CmdArgs, k8sArgs *types.K8sArgs, delegates []*types.DelegateNetConf, lastIdx int, netRt *types.RuntimeConfig, multusNetconf *types.NetConf) error {
-	logging.Debugf("delPlugins: %v, %v, %v, %v, %v, %d, %v", exec, pod, args, k8sArgs, delegates, lastIdx, netRt)
+func (mc *Cmd) delPlugins(exec invoke.Exec, pod *v1.Pod, delegates []*types.DelegateNetConf, lastIdx int, netRt *types.RuntimeConfig, multusNetconf *types.NetConf) error {
+	logging.Debugf("delPlugins: %v, %v, %v, %d, %v", exec, pod, delegates, lastIdx, netRt)
```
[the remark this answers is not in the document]
maiqueb: Done. I've stated why I think this is a bad choice, but I'm complying with the maintainer.

pkg/multus/multus.go
maiqueb: Done.

pkg/multus/multus.go
maiqueb: Done.

pkg/multus/multus.go
maiqueb: Done.

s1061123:
1. `shim` binary is not used in deployment

As far as I looked `deployments/multus-daemonset-thick-pluign.yml`, https://github.com/maiqueb/multus-cni/blob/thick-arch/deployments/multus-daemonset-thick-plugin.yml, you uses `multus` binary for CNI plugin, which is thin plugin, current binary, not `cmd/shim` binary, thick CNI plugin you designed. `shim` binary is just in multus container, not in CNI plugin directory of host.

In thick plugin mechanism, as far as I know of, `cmd/shim` is invoked from kubelet/crio and shim send a request to server and then create the interface, but your deployment does not do that.

Could you please change the deployment and test it (if you design the unix socket can be changed by option, then please test it).

s1061123:
`shim` plugin seems not to check input parameters, which is given from kubelet/crio based on CNI config. I recommend to do json syntax check before sending request to the server.

s1061123:
Regarding the configuration of thick pluign case, you need to design the following:

- 'shim' CNI config (in json)
- 'multus-daemon' server config (optional)

'shim' CNI only requires the socket path to communicate to multus-daemon, I suppose, CNI config get smaller than current multus. At least, we can remove kubeconfig from the current multus and make multus-server consume in-cluster config directory.

Regarding of multus-server config, it is pretty optional, just a good to have feature.

maiqueb:
> 1. `shim` binary is not used in deployment
> 
> As far as I looked `deployments/multus-daemonset-thick-pluign.yml`, https://github.com/maiqueb/multus-cni/blob/thick-arch/deployments/multus-daemonset-thick-plugin.yml, you uses `multus` binary for CNI plugin, which is thin plugin, current binary, not `cmd/shim` binary, thick CNI plugin you designed. `shim` binary is just in multus container, not in CNI plugin directory of host.
> 
> In thick plugin mechanism, as far as I know of, `cmd/shim` is invoked from kubelet/crio and shim send a request to server and then create the interface, but your deployment does not do that.
> 
> Could you please change the deployment and test it (if you design the unix socket can be changed by option, then please test it).

The e2e tests use a different daemonset spec - located [here](https://github.com/k8snetworkplumbingwg/multus-cni/pull/771/files#diff-b7fd7f3cf6f429156d71fd2cbb61b938a80ae7e97ab1a347f6c6a6a5f46d1d01). So yes, you're right in the sense that the regular daemonset (`deployments/...` path) spec must be updated, but the changes I'm providing are being tested in the e2e tests.

I'll update the multus deployment image as well; will update once I have it.

maiqueb:
> Regarding the configuration of thick pluign case, you need to design the following:
> 
> * 'shim' CNI config (in json)
> * 'multus-daemon' server config (optional)
> 
> 'shim' CNI only requires the socket path to communicate to multus-daemon, I suppose, CNI config get smaller than current multus. At least, we can remove kubeconfig from the current multus and make multus-server consume in-cluster config directory.
> 
> Regarding of multus-server config, it is pretty optional, just a good to have feature.

Taking into account your spot-on suggestions, I was thinking of using:
```golang
// ShimNetConf for the SHIM cni config file written in json
type ShimNetConf struct {
       types.NetConf

       MultusSocketDir string `json:"socketDir"`
}

// ControllerNetConf for the controller cni configuration
type ControllerNetConf struct {
       types.NetConf

       // support chaining for master interface and IP decisions
       // occurring prior to running ipvlan plugin
       RawPrevResult *map[string]interface{} `json:"prevResult"`
       PrevResult    *current.Result         `json:"-"`

       ConfDir string `json:"confDir"`
       CNIDir  string `json:"cniDir"`
       BinDir  string `json:"binDir"`
       // RawDelegates is private to the NetConf class; use Delegates instead
       RawDelegates    []map[string]interface{} `json:"delegates"`
       Delegates       []*DelegateNetConf       `json:"-"`
       ClusterNetwork  string                   `json:"clusterNetwork"`
       DefaultNetworks []string                 `json:"defaultNetworks"`
       LogFile         string                   `json:"logFile"`
       LogLevel        string                   `json:"logLevel"`
       LogToStderr     bool                     `json:"logToStderr,omitempty"`
       RuntimeConfig   *RuntimeConfig           `json:"runtimeConfig,omitempty"`
       // Default network readiness options
       ReadinessIndicatorFile string `json:"readinessindicatorfile"`
       // Option to isolate the usage of CR's to the namespace in which a pod resides.
       NamespaceIsolation       bool     `json:"namespaceIsolation"`
       RawNonIsolatedNamespaces string   `json:"globalNamespaces"`
       NonIsolatedNamespaces    []string `json:"-"`

       // Option to set system namespaces (to avoid to add defaultNetworks)
       SystemNamespaces []string `json:"systemNamespaces"`
       // Option to set the namespace that multus-cni uses (clusterNetwork/defaultNetworks)
       MultusNamespace string `json:"multusNamespace"`

       // Option to point to the path of the unix domain socket through which the
       // multus client / server communicate.
       MultusSocketDir string `json:"socketDir"`
}

```

maiqueb:
> `shim` plugin seems not to check input parameters, which is given from kubelet/crio based on CNI config. I recommend to do json syntax check before sending request to the server.

Done, I'm now attempting to unmarshall the stdin data to the multus configuration.

Also updated the `deployments/multus-daemonset-thick-plugin.yml` file as requested. It can't be used because the image it uses will only be created after the PR is merged ...

maiqueb:
> `shim` plugin seems not to check input parameters, which is given from kubelet/crio based on CNI config. I recommend to do json syntax check before sending request to the server.

Done, I'm now attempting to unmarshall the stdin data to the multus configuration.

Also updated the deployments/multus-daemonset-thick-plugin.yml file as requested. It can't be used because the image it uses will only be created after the PR is merged ...

Please also check https://github.com/k8snetworkplumbingwg/multus-cni/pull/771#issuecomment-1017732183; as far as I understand, that's the only remaining concern about this implementation.

@s1061123

s1061123:
Thank you so much for the update! Let me comment inline:

```
// ControllerNetConf for the controller cni configuration
type ControllerNetConf struct {
       types.NetConf

       // support chaining for master interface and IP decisions
       // occurring prior to running ipvlan plugin
       RawPrevResult *map[string]interface{} `json:"prevResult"`
       PrevResult    *current.Result         `json:"-"`
```

I don't know why you keep `types.NetConf`, `RawPrevResult` and `PrevResult` in server config. As far as I know, these parameter is used in CNI request, hence it is not necessary for server config because these information (of request) will be given by shim CNI side. 
  

```
       ConfDir string `json:"confDir"`
       CNIDir  string `json:"cniDir"`
       BinDir  string `json:"binDir"`

       // RawDelegates is private to the NetConf class; use Delegates instead
       RawDelegates    []map[string]interface{} `json:"delegates"`
       Delegates       []*DelegateNetConf       `json:"-"`
```

`Delegates` is not sued for server config as you know it is not annotated in json. 

```
       ClusterNetwork  string                   `json:"clusterNetwork"`
       DefaultNetworks []string                 `json:"defaultNetworks"`
       LogFile         string                   `json:"logFile"`
       LogLevel        string                   `json:"logLevel"`
       LogToStderr     bool                     `json:"logToStderr,omitempty"`
       RuntimeConfig   *RuntimeConfig           `json:"runtimeConfig,omitempty"`
```

`RuntimeConfig` is also not used as serverconfig.


```
       // Default network readiness options
       ReadinessIndicatorFile string `json:"readinessindicatorfile"`
       // Option to isolate the usage of CR's to the namespace in which a pod resides.
       NamespaceIsolation       bool     `json:"namespaceIsolation"`
       RawNonIsolatedNamespaces string   `json:"globalNamespaces"`
       NonIsolatedNamespaces    []string `json:"-"`

       // Option to set system namespaces (to avoid to add defaultNetworks)
       SystemNamespaces []string `json:"systemNamespaces"`
       // Option to set the namespace that multus-cni uses (clusterNetwork/defaultNetworks)
       MultusNamespace string `json:"multusNamespace"`

       // Option to point to the path of the unix domain socket through which the
       // multus client / server communicate.
       MultusSocketDir string `json:"socketDir"`
}
```

In addition, I would like to know what are going to consume `ControllerNetConf` by `multus-daemon`. 

- Is that given to multus-daemon as command line argument (./multus-daemon --config <json file config>? or
- the json will be consumed from configmap (is it fixed name configmap)?
- or just used as CNI Config (multus-daemon is invoked as usual CNI plugin?)

maiqueb:
Rebased master in [this forced push](https://github.com/k8snetworkplumbingwg/multus-cni/compare/d059b3bf43012d42828a1222fdd3b7411d9f3ef0..afced69cfc43c027b37a62c5ee63386bbf6ce716).

maiqueb:
...
> In addition, I would like to know what are going to consume `ControllerNetConf` by `multus-daemon`.
> 
> * Is that given to multus-daemon as command line argument (./multus-daemon --config ? or

What would the file look like ? JSON vs TOML for instance. 

> * the json will be consumed from configmap (is it fixed name configmap)?

I take it you mean reading the config map, and accessing its `data` variable ( a `map[string]string` ).
​
I see this as the preferable long term option. Never the less, I find it a bit odd that reading the configuration would require the k8s API ... I guess I'm just not used to it. 
​
The alternative of mounting the config map as a volume is, basically, just the previous option but with the configuration defined / accessible via the API (which has advantages ...).

> * or just used as CNI Config (multus-daemon is invoked as usual CNI plugin?)

This one is my least preferred option. I find it convoluted and misleading to users / developers / maintainers.​
I think my preference is a JSON encoded file-based approach, whose file is mounted from a config map.

Thoughts @s1061123 ?

pkg/cni/server.go
```diff
@@ -0,0 +1,315 @@
...
+			Handler: router,
+		},
+		rundir:      rundir,
+		requestFunc: HandleCNIRequest,
+		kubeclient:  kubeClient,
+		exec:        exec,
+	}
+
+	router.NotFoundHandler = http.HandlerFunc(http.NotFound)
+	router.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
s1061123: That could be an idea. I suppose it is good feature.

s1061123 (changes requested):

pkg/cni/types.go
```diff
@@ -0,0 +1,73 @@
...
+	timestamp time.Time
+	// ctx is a context tracking this request's lifetime
+	ctx context.Context
+	// cancel should be called to cancel this request
+	cancel context.CancelFunc
+	// kubeclient has the kubernetes API client
+	kubeclient *k8sclient.ClientInfo
+	// exec execs
+	exec invoke.Exec
+}
```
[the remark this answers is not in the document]
maiqueb: Instead of relying on a payload that fits libcni - `skel.CmdArgs` -  I am adding one that fits multus, and have added this mix of runtime config (`ContainerID`, `NetNS`, `IfName`), `K8sArgs` (pod name, pod namespace, pod UID, and pod sandbox ID), utilities required for handling a request - kubeclient, exec, etc. My intention was to gather together all the required data to handle a request to configure / cleanup networking for a pod.

I don't want to handle CmdArgs [StdinData](https://github.com/k8snetworkplumbingwg/multus-cni/blob/5feb1343cd8318a578d2c06b8c678450e7217068/vendor/github.com/containernetworking/cni/pkg/skel/skel.go#L42) nor `args`. Nor want to use `types.UnmarshableStrings` unless I really have to.

I did not understand if your issue is with me introducing those two types (`Cmd` and `PodRequest`) or if you want both of them gone.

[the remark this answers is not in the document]
s1061123: Regarding Stdin, it should be parsed and it is in https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/types/types.go#L27-L59 so you don't need to handle raw stdin as long as you use `types.LoadNetConf()`. 

Gathering the required info as structure may make sense to me, but I don't know why you need two structure for that.
To gathering info, one structure seems to be enough, at least to me. But you introduce two structures.
As you mentioned above, your purpose to introduce them is 'My intention was to gather together all the required data to handle a request to configure / cleanup networking for a pod.', but your reason is not enough to me to explain 'why you introduce two structure'.

I suppose you have more reason why you need two structure for one purpose. So first, I would like to clarify your reason.
Please let me know why you need both.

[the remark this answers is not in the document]
maiqueb: `Cmd` holds the runtime data (part of libcni's runtime config (ContainerID, NetNS, IfName) plus multus `K8sArgs`, holding the relevant runtime cni arguments (pod name / uid / namespace + sandbox ID). I am forced to agree its name could be better.

`PodRequest` holds that, plus the multus configuration, kubeclient, request timestamp (logging purposes), request context + cancel function, and cni exec caller.

To me, these are different structure, with different goals and are used in different scopes. I can embed one within the other; imo, it would make the relationship between them more explicit. 

Your thoughts @s1061123 ?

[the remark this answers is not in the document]
s1061123: If you design code from scratch, the structure may make sense, however,
currently pkg/multus already have CmdAdd, CmdDel and CmdCheck based on
libcni's API structure, skel.CmdArgs. In CmdAdd/CmdDel/CmdCheck, NetConf
structure, DelegateNetConf structure and other structures are used to collect
required information, such as Kubernetes pod info and so on.

Once you substitute your request to skel.CmdArgs and call
CmdAdd/CmdDel/CmdCheck, current multus code does all processing for you.
You don't need to handle stdin data and NetConf because current code handles
for you because current multus has the code without any new structure
introduction.

Your PR seems to change the current code a lot to introduce the new structure,
not utilizing current structure. To me, I don't know why you need these
structures. Please let me know why you need these structures, because current
multus does these process without your structures.

In addition, I still need your answer for why the strucutre is introduced
because some variables still ambiguous use, like 'timestamp time.Time' and
'exec invoke.Exec' in 'PodRequest'. Both variable is set once and never used
(refer/update) as far as I looked your code however, I believe that you
have clear usecase for that and it must be used in your PR with clear
purposes, because you added these variables in the PR.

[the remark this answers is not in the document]
maiqueb: ...

> Once you substitute your request to skel.CmdArgs and call CmdAdd/CmdDel/CmdCheck, current multus code does all processing for you. You don't need to handle stdin data and NetConf because current code handles for you because current multus has the code without any new structure introduction.

While I'm aware of that advantage, modelling the internal communication of the multus processes using a libcni API structure does not seem a good choice (to me). I do not think the multus shim / daemon should speak CNI between themselves.

> 
> Your PR seems to change the current code a lot to introduce the new structure, not utilizing current structure. To me, I don't know why you need these structures. Please let me know why you need these structures, because current multus does these process without your structures.

I'm proposing to keep the current structure, and use an intermediate canonical representation, which both the thin and thick options use: they both speak CNI in its API (both the multus-shim and multus thin plugin) and **uses** this intermediate representation as an implementation detail.

> 
> In addition, I still need your answer for why the strucutre is introduced because some variables still ambiguous use, like 'timestamp time.Time' and 'exec invoke.Exec' in 'PodRequest'. Both variable is set once and never used (refer/update) as far as I looked your code however, I believe that you have clear usecase for that and it must be used in your PR with clear purposes, because you added these variables in the PR.

`exec` is used (and passed to) libcni, check the following links:
  - https://github.com/k8snetworkplumbingwg/multus-cni/pull/771/files#diff-ef8a5ed2f4065ae76aaaefd57d7cacf3956c1c0a217fe8a45bd75b74197e858dR257
  - https://github.com/k8snetworkplumbingwg/multus-cni/pull/771/files#diff-ef8a5ed2f4065ae76aaaefd57d7cacf3956c1c0a217fe8a45bd75b74197e858dR281
  - https://github.com/k8snetworkplumbingwg/multus-cni/pull/771/files#diff-ef8a5ed2f4065ae76aaaefd57d7cacf3956c1c0a217fe8a45bd75b74197e858dR301

I forgot to log how long does it take to process a request - that's the purpose of the timestamp attribute. I'll add that asap; thanks for spotting it.

Are there any more attributes you consider ambiguous ?

pkg/cni/server.go
```diff
@@ -0,0 +1,315 @@
...
+		err = request.cmdCheck()
+	default:
+	}
+	logging.Verbosef("%s finished CNI request %+v, result: %q, err: %v", request.Command, request, string(result), err)
+	if err != nil {
+		// Prefix errors with request info for easier failure debugging
+		return nil, fmt.Errorf("%+v ERRORED: %v", request, err)
+	}
+	return result, nil
+}
```
maiqueb, line 63: How do you plan I do this @s1061123 without adding a new struct ? How can I (on the multus-daemon side) know what to handle if I'm just using `skel.Args` ?

maiqueb: ... obviously having 3 endpoint handlers (instead of a single common one) would do it.

pkg/cni/types.go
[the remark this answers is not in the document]
s1061123: > While I'm aware of that advantage, modelling the internal communication of the multus processes using a libcni API structure does not seem a good choice (to me). I do not think the multus shim / daemon should speak CNI between themselves.

I don't know why. Especially you design that the multus shim / daemon speak CNI between themselves, even though you told that 'I do not think the multus shim / daemon should speak CNI between themselves.' If you think the multus shim / daemon should *NOT* speak CNI between themselves, then you just design and write a code for that.

I would like to know why you feel that modelling the internal communication of the multus processes using a libcni API structure does not seem a good choice.
Please let me know clear and objective reason why.

> exec is used (and passed to) libcni, check the following links:

I am asking that why you add exec variable in PodRequest, not why you use exec.
As you mentioned and wrote in a comment, PodRequest contains per Pod basis request, which means PodRequest should have pod related information.

Looking the 'exec' variable in PodRequest, it seems to following:

1. set the variable at [cniRequestToPodRequest()](https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L150), which is come from [newCNIServer()](https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L97), that is nil as https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L85
2. then this goes to cmdAdd()/cmdCheck()/cmdDel(), without any update.

Hence, the exec variable always initial variable from server, that is nil.
I'm wondering why you set nil in the variable and hold it.

As far as I know of libcni code, exec does not contain any information related to per Pod basis information however you put this variable as per Pod basis information. I would like to know why you add because you may have some special reason.

During to write reply, you already changed the PR. So I will check your code again and reply it.

pkg/cni/server.go
s1061123: If I design the code, I will do following step:
# I don't add error handling to simplify the process but of course, it is required.

1. (as you did,) CNI shim send CNI request through [type Request](https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/types.go#L51-L57)
1. based on https://github.com/containernetworking/cni/blob/master/pkg/skel/skel.go#L57-L154, wrote function that convert 'type Request' to 'skel.CmdArgs'. Let's name reqToCmdArgs() for instance. This function should return triplet, (cmd, CmdArgs, error).
1. At handleCNIRequest() in pkg/cni/server.go, after json.Unmarshal() to get Request structure, then call reqToCmdArgs() to get the cmd and CmdArgs.
1. Based on cmd, call one of CmdAdd/CmdDel/CmdCheck with CmdArgs.

This process does not introduce 'type Cmd' and 'type PodRequest'.

s1061123:
> @s1061123 the failing unit tests test a scenario where skel.Args features an invalid netns for the pod, but a correct namespace is available in the CNI_NETNS environment variable.

Regarding these unit-tests, I'm already working on it, so no problem, please ignore it.

pkg/cni/types.go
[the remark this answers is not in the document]
maiqueb: > I don't know why. Especially you design that the multus shim / daemon speak CNI between themselves, even though you told that 'I do not think the multus shim / daemon should speak CNI between themselves.' If you think the multus shim / daemon should _NOT_ speak CNI between themselves, then you just design and write a code for that.

Hm, I thought I was doing that. They're speaking `PodRequest` / `PodResponse` - which **do** encapsulate the libcni's structs.

> 
> I would like to know why you feel that modelling the internal communication of the multus processes using a libcni API structure does not seem a good choice. Please let me know clear and objective reason why.

Being a CNI plugin, multus must understand CNI. As I see it, **at the edges** - it must be invoked via CNI API, and must in turn invoke another plugin, via CNI API. 

I don't think modelling the internal communication of 2 processes using libcni's structs is a good choice because that is exposing an implementation detail.

Furthermore, I'd like to remind you that I'm using pod request to group together all I need to handle a request - and not everything is CNI related - I need to inject the kubeclient, have a timestamp, exec (keep reading for the explanation on this ), etc. The multus daemon will eventually invoke the multus library funcs (`CmdAdd` / ... ), and for that, it'll need to pass kubeclient and exec - which, I need to mock in the tests. 

I've embedded the libcni / k8sArgs into the `PodRequest`. Is that good enough for you ? 

> 
> > exec is used (and passed to) libcni, check the following links:

> 
> I am asking that why you add exec variable in PodRequest, not why you use exec. As you mentioned and wrote in a comment, PodRequest contains per Pod basis request, which means PodRequest should have pod related information.

I miss a few logic steps on the claim that "PodRequest should have pod related information". My goal is to have everything together to **handle a request** ...  That includes the arguments required for `multus.CmdAdd` / `multus.CmdDel` / `multus.CmdCheck` - which are also kubeclient and exec. I'd also like to have a timestamp, so I can figure out how much time we spend on a request.

Could it be I should rename the struct ? 

> 
> Looking the 'exec' variable in PodRequest, it seems to following:
> 
> 1. set the variable at [cniRequestToPodRequest()](https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L150), which is come from [newCNIServer()](https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L97), that is nil as https://github.com/maiqueb/multus-cni/blob/thick-arch/pkg/cni/server.go#L85
> 2. then this goes to cmdAdd()/cmdCheck()/cmdDel(), without any update.
> 
> Hence, the exec variable always initial variable from server, that is nil. I'm wondering why you set nil in the variable and hold it.

On the production code, you're right. But on the tests I pass a `fakeExec` to the server, which allows me to mock the exec handler and unit test the server / client communication. 

https://github.com/k8snetworkplumbingwg/multus-cni/pull/771/files#diff-8998a16bc50da50c38ab8158dfb00d4a307f3dd01f4d7ec708e0005a503b18d3R210

> 
> As far as I know of libcni code, exec does not contain any information related to per Pod basis information however you put this variable as per Pod basis information. I would like to know why you add because you may have some special reason.

I just need a way to control what exec is used on the requests - as such I inject the exec from the server into every request. 

> 
> During to write reply, you already changed the PR. So I will check your code again and reply it.

pkg/cni/server.go
maiqueb: I see. Now the question is: why is returning a triplet better than returning a struct that has those things embedded ? 

My issue w/ returning more than 2 arguments: I loose track of what they are. 

It's hard for me to use these things correctly, because what the attribute is is directly correlated to the order it is returned. I'd rather do that by name, which has meaning.

dougbtv: I see both sides. So... returning a struct could be useful if it has a strong correlation with having to pass that object around between a bunch of different things. This does add some complexity though, and you have to look up the structure definition.

However, since we have named returned values in go code, we can always see those in the function definition, which does give them meaning. 

So I'd probably lean towards the triplet unless there's a case for tying the values together in a number of other places.

maiqueb: > I see both sides. So... returning a struct could be useful if it has a strong correlation with having to pass that object around between a bunch of different things. This does add some complexity though, and you have to look up the structure definition.

... the same way I have to look up whatever the function's returning. 

> 
> However, since we have named returned values in go code, we can always see those in the function definition, which does give them meaning.

For that I have to navigate to the function's definition. I.e. the caller knows nothing about the returned types until he moves to the function's definition. 

Named return arguments are not a feature I would happily use. And I think it is not being used anywhere in the project.

> 
> So I'd probably lean towards the triplet unless there's a case for tying the values together in a number of other places.

I personally don't see how, but I will not choose this particular hill to die on. Whatever you guys say. I do think it reads worse, but I've already stated my opinion.

s1061123: > I personally don't see how, but I will not choose this particular hill to die on. Whatever you guys say. I do think it reads worse, but I've already stated my opinion.

I understand that you want not to use multiple return value for the golang code, however, the reason seems to be subjective, based on  your impression, not objective, hence we cannot agree on that because we don't have such impression.

Return multiple value is pretty usual language feature and not so unique feature that only in golang. As far as I know of, multiple return value (or similar language functionality) is supported not only in go, also in lisp, Standard ML, haskell, ocaml, ruby, python and rust. 

Many software projects written in golang uses multiple return value as usual as following:

- https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/pods/helpers.go#L65
- https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/services/template.go#L297
- https://cs.opensource.google/go/go/+/refs/tags/go1.17.6:src/net/pipe.go;l=113

If you suppose multiple values is the particular hill to die on, then these projects may be the one of them. If you want to change these culture, here is not the place to discuss about it. Please go appropriate places to discuss. 

Considering code volumes (less code is easy to maintain) and simplicity (i.e. less structure we need to maintain), we want to use multiple return value as usual golang does, without any structures.

maiqueb: > If you suppose multiple values is the particular hill to die on, then these projects may be the one of them. If you want to change these culture, here is not the place to discuss about it. Please go appropriate places to discuss.

I don't think you understood me; I said "I will **not** choose this particular hill to die on. Whatever you guys say. ". This means that while I disagree with you, I'll comply. 

I'm waiting for the rest of the code to be reviewed before updating the PR.

s1061123: I suppose that the update (to incorporate with my comment) seems to be big and it also changes some codes not only in cni/server.go also types/conf.go and others. So could you please proceed your change first?

deployments/multus-daemonset-thick-plugin.yml
```diff
@@ -178,6 +186,21 @@ spec:
...
+              - "/host/var/run/multus-cni"
+            resources:
+              requests:
+                cpu: "10m"
+                memory: "15Mi"
+            securityContext:
+              privileged: true
+            volumeMounts:
+              - name: host-var-run-multus-cni
+                mountPath: /host/var/run/multus-cni
```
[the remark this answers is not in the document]
s1061123: > - where to put the code that sets up the multus socket directory permissions
> - how we setup that directory

How about to make it in server side (not initContainer, just in multus-daemon)? In this case, we just provide a way to configure as command line option.

s1061123 (changes requested):
Please provide some document in 'docs/thick-plugin.md' that describe

- what is the thick plugin
- how to use it (deploy, config parameters and so on)

deployments/multus-daemonset-thick-plugin.yml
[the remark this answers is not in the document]
s1061123: And let me ask you another question. Why do you need to create the directory in initContainer?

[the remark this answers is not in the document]
maiqueb: > And let me ask you another question. Why do you need to create the directory in initContainer?

The directory is bind mounted into the host. AFAIU, that is what creates the directory, not "me". 

Regarding having the server setup the permissions for the directory, I've replied to that already in https://github.com/k8snetworkplumbingwg/multus-cni/pull/771#discussion_r800837670; I think it's past it's responsibility. The server should consume an already existing - and configured - directory, and fail if the permissions are insecure. 

What are your reasons to avoid using this particular init container ?

maiqueb:
[Rebased latest master](https://github.com/k8snetworkplumbingwg/multus-cni/compare/20e87a7b40d2c98cfb6ee71b3994a6e834c5c25e..c359b56e810ceba70a28b7cabb9e9b708accd853).

maiqueb:
Added documentation and dropped unused server configuration in [this forced push](https://github.com/k8snetworkplumbingwg/multus-cni/compare/c359b56e810ceba70a28b7cabb9e9b708accd853..7fcfe77ca802f8039a15ff7139ca6cd4ca9af4f8).

pkg/cni/server.go
maiqueb: Done - no more `PodRequest`, `command`, and now everything is passed via function parameters. 

Resolving this conversation.

pkg/cni/types.go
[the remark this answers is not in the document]
maiqueb: Removed `PodRequest`. Thus, I think this conversation can be resolved.

Let me know if I'm wrong.

dougbtv (approved):

s1061123 (approved):
Looks good to me, so let's move it to 4.0 branch.

PR #772, opened 2022-01-04
run gofmt on the code

maiqueb:
Format the golang code located in `pkg/...` and `cmd/...`


dougbtv (approved):
LGTM. Unsure how it was missed, but, we should have this :+1:

PR #777, opened 2022-01-13
hotplug feature: add pod networks annotation controller

maiqueb:
Features:
  - a very opinionated refactor of multus, allowing its delegate handling functions to be called outside the package
  - a controller listening to multus annotations. When the multus annotations changes, it will invoke the corresponding delegate
  - e2e tests
  - two runtimes (required to read to network namespace of the pod being updated - which is a CNI parameter)
    - containerd
    - crio
  - unit tests of every new package

**TODO**: 
  - figure out what is holding this PR
  - act upon revive's comments


maiqueb:
[This forced push](https://github.com/k8snetworkplumbingwg/multus-cni/compare/24340982a473655141a01c20aa36e745595f643f..f99f08ef3ea53f7eafa2d1eaf6c6e5e0c0d5e8a3) features:
  - removed the status condition updates
  - fixed revive's comments (causing the unit test jobs to fail).

maiqueb:
@s1061123 I think the approach you shared offline is in line w/ this commit: https://github.com/k8snetworkplumbingwg/multus-cni/pull/777/commits/df801b2e92860769fe18b2b429bf33d1ec995a82

I would appreciate if you'd help me understand how to build the missing data on the controller.

Furthermore, a logic split is required on the multus `CmdAdd` function.

If we take this route, when the time comes to handle the default route update, another serious refactor on the multus pkg will be required.

pkg/containerruntimes/crio/crio_test.go
```diff
@@ -0,0 +1,61 @@
...
+	When("a live container is provisioned in the runtime", func() {
+		const (
+			containerID = "1234"
+			netnsPath   = "bottom-drawer"
+		)
+		BeforeEach(func() {
+			runtime = newDummyCrioRuntime(fake.WithCachedContainer(containerID, netnsPath))
+		})
+
+		It("cannot extract the network namespace of a container", func() {
```
maiqueb, line 41: the it clause description is wrong, probably a copy paste error.

maiqueb:
This is being implemented in https://github.com/maiqueb/multus-dynamic-networks-controller/, now that a 3rd party controller with access to the multus thick-plugin socket can trigger delegate add / delete.

PR #782, opened 2022-01-21
WIP| e2e, tests: re-write macvlan tests in golang

maiqueb:
This PR is a first step towards re-writing the bash e2e tests to golang.

Currently it does:
  - port the macvlan test
    - it checks the `networkStatus` annotation rather than checking the IP inside the pod. It checks the multus API, not the inner pod state.
  - adds a git action to run this new golang based test both for thick / thin plugin alternatives.

ToDo:
  - use a random namespace per each test. This would speed up the tests (I wouldn't have to wait for the pods to be gone ...).
  - port the simple pod tests
  - port the default route tests
  - decide if it is OK just relying on the API - network status annotation - or we need to check the pod's inner state.

### DIFF
--- a/.github/workflows/kind-e2e.yml
+++ b/.github/workflows/kind-e2e.yml
@@ -39,6 +39,9 @@ jobs:
         working-directory: ./e2e
         run: ./test-default-route1.sh
 
+      - name: Run golang e2e tests
+        run: ./hack/test-e2e.sh
+
       - name: cleanup cluster and registry
         run: |
           kind delete cluster
--- a/.github/workflows/legacy-kind-e2e.yml
+++ b/.github/workflows/legacy-kind-e2e.yml
@@ -39,6 +39,9 @@ jobs:
         working-directory: ./e2e
         run: ./test-default-route1.sh
 
+      - name: Run golang e2e tests
+        run: ./hack/test-e2e.sh
+
       - name: cleanup cluster and registry
         run: |
           kind delete cluster
--- a/e2e/multus_test.go
+++ b/e2e/multus_test.go
@@ -1,0 +1,231 @@
+// Copyright (c) 2021 Multus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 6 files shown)


s1061123:
Thank you for your PR. As far as I looked your code, it seems to have different aspects from current e2e test.

As you tell in above comment, your tests checks networkStatus annotation in pod based on assumption that multus always create right network status. From multus testing point of view, we also need to check that multus creates that right configuration based on net-attach-def. In this case, we should check actual network status rather than networkStatus because multus (or used vendor library) may have bugs around network status generation, and e2e tests means 'end-to-end' tests, so we should focus on end-to-end, from config text to actual interface config, so I believe that we should keep current test and current test fits to use shell script based approach.

But on the other side, having go based test provides other various tests I suppose. How about to create `e2e-go` or some other directory to put the file?

maiqueb:
> Thank you for your PR. As far as I looked your code, it seems to have different aspects from current e2e test.
> 
> As you tell in above comment, your tests checks networkStatus annotation in pod based on assumption that multus always create right network status. From multus testing point of view, we also need to check that multus creates that right configuration based on net-attach-def. In this case, we should check actual network status rather than networkStatus because multus (or used vendor library) may have bugs around network status generation, and e2e tests means 'end-to-end' tests, so we should focus on end-to-end, from config text to actual interface config, so I believe that we should keep current test and current test fits to use shell script based approach.

Right. I half expected something like this. Please also note I haven't deleted the existing tests. 

> 
> But on the other side, having go based test provides other various tests I suppose. How about to create `e2e-go` or some other directory to put the file?

Perfectly fine for me.


> In addition, I strongly recommend you to start the discussion before implementation because without any consensus in community, dynamic changes tends not to accepted.

Note taken; I only intended to provide a minimal example of what I consider a decent looking test to start the discussion.

maiqueb:
> Thank you for your PR. As far as I looked your code, it seems to have different aspects from current e2e test.
> 
> As you tell in above comment, your tests checks networkStatus annotation in pod based on assumption that multus always create right network status. From multus testing point of view, we also need to check that multus creates that right configuration based on net-attach-def. In this case, we should check actual network status rather than networkStatus because multus (or used vendor library) may have bugs around network status generation, and e2e tests means 'end-to-end' tests, so we should focus on end-to-end, from config text to actual interface config, so I believe that we should keep current test and current test fits to use shell script based approach.
> 
> But on the other side, having go based test provides other various tests I suppose. How about to create `e2e-go` or some other directory to put the file?

I don't object to have them co-exist. I'll need to re-use the bash script that creates the cluster...

Would you be ok with having this directory **inside** e2e ? We could use the hierarchy to understand what goes where. 

@s1061123

PR #785, opened 2022-02-07
Replace setenv with runtimeConfig set

s1061123:
setenv refers environment variables, which is unique in process,
not unique to go routine. Hence it may causes some issue in multi
threaded case, hence it is replaced with libcni's runtimeConfig
value set to set these variables at libcni side, after process
fork.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -206,7 +206,7 @@ func confCheck(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.N
 }
 
 func confDel(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.NetConf, exec invoke.Exec) error {
-	logging.Debugf("conflistDel: %v, %s", rt, string(rawNetconf))
+	logging.Debugf("confDel: %v, %s", rt, string(rawNetconf))
 	// In part, adapted from K8s pkg/kubelet/dockershim/network/cni/cni.go
 	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
 	binDirs = append([]string{multusNetconf.BinDir}, binDirs...)
@@ -285,23 +285,15 @@ func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *t
 	return err
 }
 
-func delegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, ifName string, delegate *types.DelegateNetConf, rt *libcni.RuntimeConf, multusNetconf *types.NetConf, cniArgs string) (cnitypes.Result, error) {
-	logging.Debugf("delegateAdd: %v, %s, %v, %v", exec, ifName, delegate, rt)
-	if os.Setenv("CNI_IFNAME", ifName) != nil {
-		return nil, logging.Errorf("delegateAdd: error setting environment variable CNI_IFNAME")
-	}
+func delegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, delegate *types.DelegateNetConf, rt *libcni.RuntimeConf, multusNetconf *types.NetConf) (cnitypes.Result, error) {
+	logging.Debugf("delegateAdd: %v, %v, %v", exec, delegate, rt)
 
-	if err := validateIfName(os.Getenv("CNI_NETNS"), ifName); err != nil {
-		return nil, logging.Errorf("delegateAdd: cannot set %q interface name to %q: %v", delegate.Conf.Type, ifName, err)
+	if err := validateIfName(rt.NetNS, rt.IfName); err != nil {
+		return nil, logging.Errorf("delegateAdd: cannot set %q interface name to %q: %v", delegate.Conf.Type, rt.IfName, err)
 	}
 
 	// Deprecated in ver 3.5.
 	if delegate.MacRequest != "" || delegate.IPRequest != nil {
-		if cniArgs != "" {
-			cniArgs = fmt.Sprintf("%s;IgnoreUnknown=true", cniArgs)
-		} else {
-			cniArgs = "IgnoreUnknown=true"
-		}
 		if delegate.MacRequest != "" {
 			// validate Mac address
 			_, err := net.ParseMAC(delegate.MacRequest)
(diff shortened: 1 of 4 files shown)


pkg/multus/multus.go
```diff
@@ -285,32 +285,23 @@ func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *t
...
+		return nil, logging.Errorf("delegateAdd: cannot set %q interface name to %q: %v", delegate.Conf.Type, rt.IfName, err)
 	}
 
 	// Deprecated in ver 3.5.
 	if delegate.MacRequest != "" || delegate.IPRequest != nil {
-		if cniArgs != "" {
-			cniArgs = fmt.Sprintf("%s;IgnoreUnknown=true", cniArgs)
-		} else {
-			cniArgs = "IgnoreUnknown=true"
-		}
```
maiqueb, line 304: I assume `rt.Args` already features this `IgnoreUnknown=true` key value pair. Would you confirm ? 

If not ... Don't we want to preserve the existing behavior ?

pkg/types/conf.go
```diff
@@ -218,6 +219,32 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 		},
 	}
 
+	// get CNI_ARGS and set it if it is not exists in rt.Args
```
maiqueb, line 222:
```suggestion
        // get CNI_ARGS and set it if it does not exist in rt.Args
```

Or just drop the comment.

pkg/multus/multus_test.go
```diff
@@ -68,30 +68,37 @@ type fakeExec struct {
 	delIndex        int
 	chkIndex        int
 	expectedDelSkip int
-	plugins         []*fakePlugin
+	//plugins         []*fakePlugin
```
maiqueb, line 71: Curious about what are your reasons to keep this around in the code .

pkg/multus/multus_test.go
```diff
@@ -160,718 +181,103 @@ func (f *fakeExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData
 
 	dec := json.NewDecoder(reader)
 	enc := json.NewEncoder(writer)
-	err = dec.Decode(&m)
-	Expect(err).NotTo(HaveOccurred())
-	for k := range m {
-		if k == "prevResult" {
-			delete(m, k)
-		}
-	}
```
maiqueb, line 169: I don't understand this huge diff until (old) line 817. 

Either I'm missing something or this is the same code as in https://github.com/k8snetworkplumbingwg/multus-cni/pull/785/files#diff-4a220b8816e7d9c591fef947d5847068f9e454f9a2ff482aed709128fabc4c88R184-R192

Could this be just some spaces vs tabs issue ?

maiqueb (commented):
I'm worried about the > 1500 lines changes in the tests.

Some was the code in `multus_test,go` was deleted and added in the same place, which is weird.

Logic wise, I think this patch is good.

pkg/multus/multus.go
s1061123: We can assume at that time, but no guarantee. Please let me know if you have something that defines the assumption.

pkg/multus/multus_test.go
s1061123: changed.

pkg/types/conf.go
s1061123: fixed.

pkg/multus/multus_test.go
s1061123: It is just remove deduped testing.  Check the master branch, you can find several tests are identical, for example "returns the previous result using CmdCheck"

- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus_test.go#L402
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus_test.go#L559
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus_test.go#L717
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus_test.go#L874
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus_test.go#L1031

pkg/multus/multus.go
maiqueb: Well, I assumed simply because you're changing the current behavior (i.e. `IgnoreUnknown=true` is *always* available on the CNI args of the delegate). With your patch, it stops being there. 

I just want to be sure that's OK. AFAIU, changing this behavior is OK if this attribute is not used for anything - I don't know if it is - hence the question.

pkg/multus/multus_test.go
maiqueb: I'm just saying that the code I've linked above (pasting [here](https://github.com/k8snetworkplumbingwg/multus-cni/pull/785/files#diff-4a220b8816e7d9c591fef947d5847068f9e454f9a2ff482aed709128fabc4c88R183-R191) the updated link) seems to be equal before and after this patch - without a line of comment - but is somehow still listed as a difference. I find that odd, and pointed it out.

I'm OK with removing the duplicated tests - but I haven't checked if **all** that is deleted are those tests ( or any other duplicates).

dougbtv: I understand the gist of what Tomo's pointing out, which is that there's a number of duplicated tests with the same name. I think it's appropriate to remove them.

But you're right Miguel, it's super hard to read the diff on GH.

maiqueb: > I understand the gist of what Tomo's pointing out, which is that there's a number of duplicated tests with the same name. I think it's appropriate to remove them.

I have no issue - at all - about removing the duplicate tests. I want them gone as much as you do. For the sake of clarity: we all agree on that: **no point preserving duplicate tests**. 

I personally would have done it in a separate PR - since it would be a lot easier to review this one -  but I haven't complained about that until now, and I **believe that's besides the point**.

> 
> But you're right Miguel, it's super hard to read the diff on GH.

I'm showing in the links at the bottom of https://github.com/k8snetworkplumbingwg/multus-cni/pull/785#discussion_r800928607 there are **more** changes in the multus tests than just deleting duplicates. All I'm saying is there's something weird there, since the code seems the same to my eyes: it could be a github bug, could be spaces vs tabs, could be anything else. I don't know, but it caught my eye. 

If the maintainers (you) feel / know there's nothing wrong there, by all means, proceed.

PR #803, opened 2022-03-03
only warn when netns can't be opened

dougbtv:
Should make a best effort during CNI DEL


s1061123 (approved):
It looks good to me. Thanks!

PR #805, opened 2022-03-04
fix the usage of flag "overrideNetworkName"

cyclinder:
now the usage of flag "overrideNetworkName" is incomplete. so this pr will fix it.


s1061123 (approved):
It looks good to me, thanks!

PR #806, opened 2022-03-04
Refine unit test in pkg/multus

s1061123:
Some of tests are duplicated hence deduped. In addition, sort
these unit tests based on cniVersion.


dougbtv (approved):
LGTM, thanks Tomo

PR #807, opened 2022-03-04
Fix thick plugin to run kind-e2e test

s1061123:


cmd/multus-daemon/main.go
```diff
@@ -108,8 +107,9 @@ func main() {
 		os.Exit(3)
 	}
 
-	if *multusConfigFile == defaultMultusConfigFile {
-		if *cniVersion == defaultMultusCNIVersion {
+	// Generate multus CNI config from current CNI config
+	if *multusConfigFile == "auto" {
+		if *cniVersion == "" {
```
dougbtv, line 112: +1. I think it's OK to reduce down to the string value in these cases because of legibility.

If there's a higher reliance on repeating these strings we could fall back to the previous method.

deployments/multus-daemonset-thick-plugin.yml
```diff
@@ -186,21 +186,6 @@ spec:
...
-              - "/host/var/run/multus-cni"
-            resources:
-              requests:
-                cpu: "10m"
-                memory: "15Mi"
-            securityContext:
-              privileged: true
-            volumeMounts:
-              - name: host-var-run-multus-cni
-                mountPath: /host/var/run/multus-cni
```
dougbtv, line 203: Tomo explained that this might not be necessary in all cases, so let's try without this, and we could return to it if we're seeing that this mode isn't set properly in the future.

pkg/server/config/manager.go
```diff
@@ -41,29 +41,31 @@ type Manager struct {
 	multusConfigDir      string
 	multusConfigFilePath string
 	primaryCNIConfigPath string
+	cniVersion	string
+	forceCNIVersion	bool
```
dougbtv, line 45: nit: gofmt looks funny on this

pkg/server/socket.go
```diff
@@ -31,31 +30,8 @@ const (
...
-		// Check permissions
-		if info.Mode()&fullReadWriteExecutePermissions != thickPluginSocketRunDirPermissions {
-			return fmt.Errorf("insecure permissions on pod info socket directory %s: %v", rundir, info.Mode())
-		}
-
-		// Finally remove the socket file so we can re-create it
-		if err := os.Remove(socketpath); err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("failed to remove old pod info socket %s: %v", socketpath, err)
-		}
+		return fmt.Errorf("failed to remove old pod info socket directory %s: %v", rundir, err)
```
dougbtv, line 34: nit: If we're removing all the failure cases, we could improve the message to say `failed to remove old pod info socket directory (please stat the socketfile and check the mode is restrictive and UID != 0)`

dougbtv (approved):
LGTM. I left a few comments that can be followed up on in upcoming changes to the feature branch. Thanks!

Issue #814, opened 2022-03-22
Enhancement: Add custom log settings


cyclinder:
**What would you like to be added**:
Hi everyone, I think it would be better if some parameters of logging settings could be customized in multus，such as `MaxSize`,`MaxBackups`,`MaxAge`. 

**Why is this needed**:
At present, multus can only set the `log file path`, does not support setting log other configurations, refer to
please refer to: https://github.com/k8snetworkplumbingwg/multus-cni/blob/45428a53ceedcf6359a7c2f75e8b84d64a10148b/pkg/logging/logging.go#L144-L149

If you think it's worth, I'm willing to do the work :)

PR #817, opened 2022-03-27
add some flags for customize the log settings

cyclinder:
Fixes: https://github.com/k8snetworkplumbingwg/multus-cni/issues/814

adding some flags for customize the log settings, such as`MaxSize`, `MaxBackups`, `MaxAge` etc.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -40,6 +40,10 @@ const (
 	defaultMultusGlobalNamespaces       = ""
 	defaultMultusKubeconfigPath         = "/etc/cni/net.d/multus.d/multus.kubeconfig"
 	defaultMultusLogFile                = ""
+	defaultMultusLogMaxSize             = 100 // megabytes
+	defaultMultusLogMaxAge              = 5   // days
+	defaultMultusLogMaxBackups          = 5
+	defaultMultusLogCompress            = true
 	defaultMultusLogLevel               = ""
 	defaultMultusLogToStdErr            = false
 	defaultMultusMasterCNIFile          = ""
@@ -55,6 +59,10 @@ const (
 	multusConfigFileVarName       = "multus-conf-file"
 	multusGlobalNamespaces        = "global-namespaces"
 	multusLogFile                 = "multus-log-file"
+	multusLogMaxSize              = "multus-log-max-size"
+	multusLogMaxAge               = "multus-log-max-age"
+	multusLogMaxBackups           = "multus-log-max-backups"
+	multusLogCompress             = "multus-log-Compress"
 	multusLogLevel                = "multus-log-level"
 	multusLogToStdErr             = "multus-log-to-stderr"
 	multusKubeconfigPath          = "multus-kubeconfig-file-host"
@@ -76,6 +84,10 @@ func main() {
 	logToStdErr := flag.Bool(multusLogToStdErr, defaultMultusLogToStdErr, "If the multus logs are also to be echoed to stderr.")
 	logLevel := flag.String(multusLogLevel, defaultMultusLogLevel, "One of: debug/verbose/error/panic. Used only with --multus-conf-file=auto.")
 	logFile := flag.String(multusLogFile, defaultMultusLogFile, "Path where to multus will log. Used only with --multus-conf-file=auto.")
+	logMaxSize := flag.Int(multusLogMaxSize, defaultMultusLogMaxSize, "the maximum size in megabytes of the log file before it gets rotated")
+	logMaxAge := flag.Int(multusLogMaxAge, defaultMultusLogMaxAge, "the maximum number of days to retain old log files in their filename")
+	logMaxBackups := flag.Int(multusLogMaxBackups, defaultMultusLogMaxBackups, "the maximum number of old log files to retain")
+	logCompress := flag.Bool(multusLogCompress, defaultMultusLogCompress, "compress determines if the rotated log files should be compressed using gzip")
 	cniVersion := flag.String(multusCNIVersion, defaultMultusCNIVersion, "Allows you to specify CNI spec version. Used only with --multus-conf-file=auto.")
 	additionalBinDir := flag.String(multusAdditionalBinDirVarName, defaultMultusAdditionalBinDir, "Additional binary directory to specify in the configurations. Used only with --multus-conf-file=auto.")
 	readinessIndicator := flag.String(multusReadinessIndicatorFile, defaultMultusReadinessIndicatorFile, "Which file should be used as the readiness indicator. Used only with --multus-conf-file=auto.")
@@ -138,6 +150,28 @@ func main() {
 		if *readinessIndicator != defaultMultusReadinessIndicatorFile {
 			configurationOptions = append(
 				configurationOptions, config.WithReadinessFileIndicator(*readinessIndicator))
+		}
+
+		// logOptions
+
+		var logOptionFuncs []config.LogOptionFunc
+		if *logMaxAge != defaultMultusLogMaxAge {
+			logOptionFuncs = append(logOptionFuncs, config.WithLogMaxAge(logMaxAge))
+		}
+		if *logMaxSize != defaultMultusLogMaxSize {
+			logOptionFuncs = append(logOptionFuncs, config.WithLogMaxSize(logMaxSize))
+		}
+		if *logMaxBackups != defaultMultusLogMaxBackups {
+			logOptionFuncs = append(logOptionFuncs, config.WithLogMaxBackups(logMaxBackups))
+		}
+		if *logCompress != defaultMultusLogCompress {
+			logOptionFuncs = append(logOptionFuncs, config.WithLogCompress(logCompress))
+		}
+
+		if len(logOptionFuncs) > 0 {
+			logOptions := &config.LogOptions{}
+			config.MutateLogOptions(logOptions, logOptionFuncs...)
+			configurationOptions = append(configurationOptions, config.WithLogOptions(logOptions))
 		}
 
 		multusConfig, err := config.NewMultusConfig(multusPluginName, *cniVersion, *multusKubeconfig, configurationOptions...)
--- a/pkg/config/generator.go
+++ b/pkg/config/generator.go
@@ -33,6 +33,9 @@ const (
 	singleConfigCapabilityKey = "capabilities"
 )
 
+// LogOptionFunc mutates the `LoggingOptions` object
+type LogOptionFunc func(logOptions *LogOptions)
+
 // Option mutates the `conf` object
 type Option func(conf *MultusConf)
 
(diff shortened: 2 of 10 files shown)


s1061123:
Thank you for the PR, @cyclinder 

As far as I looked your PR, your change introduces option in cmd/controller. In this case, multus CNI's logging is NOT changed because 'cmd/controller' is not the multus CNI plugin. 'cmd/controller' is helper component to watch delegating cluster network config and generate multus config if required (i.e. "-multus-autoconfig-dir" option).

If you want to add option, you need to check 
- ['cmd/main.go'](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/cmd/main.go) for actual multus CNI plugin's main function
- ['type NetConf'](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/types/types.go#L26-L59) in 'pkg/types' for multus CNI config (i.e. CNI json config)

In addition, if you want to add new CNI config into multus CNI, I recommend to make it hierarchy such as:

```
{
    "type": "multus",
     (snip, but multus main CNI config),
    "loggingOptions": {
      "maxSize": ...
      "maxAge": ...
     }
  }
```

cyclinder:
@s1061123 Thanks for your suggestion！

According to your advices, I make some change.  This PR  is still  in progress and may still need some changes,  so I changed it to Draft.

 Can you take a look again？I would be grateful for your advice!

cmd/controller/main.go
```diff
@@ -40,8 +41,12 @@ const (
 	defaultMultusGlobalNamespaces       = ""
 	defaultMultusKubeconfigPath         = "/etc/cni/net.d/multus.d/multus.kubeconfig"
 	defaultMultusLogFile                = ""
+	defaultMultusLogMaxSize             = 100 // megabytes
+	defaultMultusLogMaxAge              = 5   // days
+	defaultMultusLogMaxBackups          = 5
+	defaultMultusLogCompress            = true
 	defaultMultusLogLevel               = ""
-	defaultMultusLogToStdErr            = false
+	defaultMultusLogToStdErr            = true
```
s1061123, line 49: Could you revert it because of 'false' as default value?

pkg/config/generator.go
```diff
@@ -33,36 +34,41 @@ const (
...
-	ReadinessIndicatorFile   string          `json:"readinessindicatorfile,omitempty"`
-	Type                     string          `json:"type"`
+	BinDir                   string                `json:"binDir,omitempty"`
+	Capabilities             map[string]bool       `json:"capabilities,omitempty"`
+	CNIVersion               string                `json:"cniVersion"`
+	Delegates                []interface{}         `json:"delegates"`
+	LogFile                  string                `json:"logFile,omitempty"`
+	LogLevel                 string                `json:"logLevel,omitempty"`
+	LogToStderr              bool                  `json:"logToStderr,omitempty"`
+	LoggingOptions           *types.LoggingOptions `json:"loggingOptions,omitempty"`
```
s1061123, line 52: How about to use 'LogOptions' instead of 'Logging' to make it consistent with 'LogFile' 'LogLevel' and so on?

pkg/types/types.go
```diff
@@ -69,6 +67,17 @@ type RuntimeConfig struct {
...
+type LoggingOptions struct {
+	LogFile     string `json:"logFile"`
+	LogLevel    string `json:"logLevel"`
+	LogToStderr bool   `json:"logToStderr,omitempty"`
+	MaxAge      int    `json:"maxAge,omitempty"`
+	MaxSize     int    `json:"maxSize,omitempty"`
+	MaxBackups  int    `json:"maxBackups,omitempty"`
+	Compress    bool   `json:"compress,omitempty"`
+}
+
```
s1061123, line 80: I would like to keep current logParameters in NetConf and add logging specific options are in LoggingOptions (logOptions) because this changes incompatible to previous version.

So setLogStderr()/setLogLevel() should be reverted as SetLogStderr()/SetLogLevel()/SetLogFile()

s1061123 (changes requested):

s1061123:
In addition, your code seems not to support without logFile option. Please consider multus log without file option because all user want NOT to have log file (e.g. someone just want to put stderr, without any log file).

cmd/controller/main.go
cyclinder: Sure!

cyclinder: My previous thought about this is wrong.. :)

pkg/config/generator.go
cyclinder: Agree, Done

pkg/types/types.go
cyclinder: Sorry, I didn't consider the compatibility.., Now I'm reverted it

pkg/types/conf_test.go
```diff
@@ -116,6 +116,12 @@ var _ = Describe("config operations", func() {
 			"type": "multus",
 			"logLevel": "debug",
 			"logFile": "/var/log/multus.log",
+			"logOptions": {
+			    "maxAge": 5,
+                "maxSize": 100,
+                "maxBackups": 5,
+                "compress": true
+			}
```
s1061123, line 124: it seems to be different indent. Did you do `gofmt -s -w` command to fix golang format?

pkg/types/types.go
```diff
@@ -69,6 +70,14 @@ type RuntimeConfig struct {
 	CNIDeviceInfoFile string          `json:"CNIDeviceInfoFile,omitempty"`
 }
 
+// LogOptions specifies the configuration of the log
+type LogOptions struct {
+	MaxAge     int  `json:"maxAge,omitempty"`
+	MaxSize    int  `json:"maxSize,omitempty"`
+	MaxBackups int  `json:"maxBackups,omitempty"`
+	Compress   bool `json:"compress,omitempty"`
+}
```
s1061123, line 79: Could you please provide document for new option, in Multus-cni Configuration Reference
, https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md?

s1061123, line 79: In this structure, you defined each value as 'omitempty', so if user does not specify any of them, then 0 or false will be set to the variable, as https://pkg.go.dev/encoding/json#Marshal .

This means if user uses current multus config (which does not have any LogOption) everything is set to 0 or false. This is not what current user want to (at least, MaxAge and MaxSize are 0, hence current log is gone).

Could you think about the situation and provide how to solve that?

pkg/config/generator.go
```diff
@@ -138,6 +151,46 @@ func WithGlobalNamespaces(globalNamespaces string) Option {
...
+}
+
+// WithLogCompress mutates the inner state to set the
+// logCompress attribute
+func WithLogCompress(compress bool) LogOptionFunc {
+	return func(logOptions *LogOptions) {
+		logOptions.Compress = compress
+	}
+}
+
```
s1061123, line 193: Could you please add test code for that?

pkg/config/generator.go
```diff
@@ -162,6 +215,14 @@ func WithLogFile(logFile string) Option {
...
 }
 
+// WithLogOptions mutates the inner state to set the
+// LogOptions attribute
+func WithLogOptions(logOptions *LogOptions) Option {
+	return func(conf *MultusConf) {
+		conf.LogOptions = logOptions
+	}
+}
+
```
s1061123, line 225: Could you please add test code for that?

s1061123 (changes requested):
Thank you for incorporate my previous comments. Now I add several comments.
I guess that you just did some simple testing and missing some tests (what's happen if user uses multus config without loggingOption). So could you please test your code with several condition (e.g. changes multus CNI config or so)?

pkg/types/conf_test.go
cyclinder: Sure(little strange), I did fix the golang format, but it seems to not work well

pkg/types/types.go
cyclinder: Sure.

cyclinder:
Hi @s1061123 , I have basically completed all the suggestions you mentioned above, and now all the tests have passed, do you have some other suggestions?

pkg/logging/logging_test.go
```diff
@@ -79,4 +80,37 @@ var _ = Describe("logging operations", func() {
 		currentLevel := loggingLevel
 		Expect(currentLevel).To(Equal(GetLoggingLevel()))
 	})
+
```
s1061123, line 83: Did you test with missing some options?

For example:

```
		logOptions := &LogOptions{
			MaxAge:     1,
			MaxBackups: 1,
			Compress:   true,
		}
```

In this case, user omit MaxSize so that user expects that multus set default value to MaxSize (not MaxSize=0).

cyclinder: Thanks for your point, Done

pkg/logging/logging.go
```diff
@@ -41,9 +41,48 @@ const (
...
+		if options.MaxSize != 0 {
+			logger.MaxSize = options.MaxSize
+		} else {
+			logger.MaxSize = 100
+		}
+		if options.MaxBackups != 0 {
+			logger.MaxBackups = options.MaxBackups
+		} else {
+			logger.MaxBackups = 5
+		}
```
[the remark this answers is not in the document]
cyclinder: yeah! I made some tests for using pointer for structure type in my local machine, it works well! Thanks!

cyclinder:
@s1061123 Would you mind taking a look?

s1061123:
Sure, will do that within a few days.

pkg/logging/logging.go
```diff
@@ -41,9 +41,52 @@ const (
...
+		} else {
+			logger.Compress = true
+		}
+	} else {
+		// give some default value
+		logger.MaxSize = 100
+		logger.MaxAge = 5
+		logger.MaxBackups = 5
+		logger.Compress = true
+	}
```
s1061123, line 85: This may be more simple.

```
func SetLogOptions(options *LogOptions) {
        // give some default value
        logger.MaxSize = 100
        logger.MaxAge = 5
        logger.MaxBackups = 5
        logger.Compress = true
        if options != nil {
                if options.MaxAge != nil {
                        logger.MaxAge = *options.MaxAge
                }
                if options.MaxSize != nil {
                        logger.MaxSize = *options.MaxSize
                }
                if options.MaxBackups != nil {
                        logger.MaxBackups = *options.MaxBackups
                }
                if options.Compress != nil {
                        logger.Compress = *options.Compress
                }
        }

        loggingW = logger
}
```

s1061123 (changes requested):
In your PR, `logOption{}` is generated in multus Config when the user configures without log option. In such case, better to remove `logOption{}` from generated config.

I create some code proposal for that. This code does not create empty `logOption{}` hence you don't need to change generator_test.go and manager_test.go to add `logOption{}`.
https://github.com/s1061123/multus-cni/tree/propose-logoption

pkg/logging/logging.go
cyclinder: Sure!  Thank you for your comment and I totally agree with your suggestion.

PR #823, opened 2022-04-04
Remove error handling for getPod to force to proceed cmdDel.

### LINKED ISSUE #822: cmdDel does not exit gracefully when K8s API is not available
When the multus cannot get K8sAPI at CmdDel, then multus exit with error in CmdDel. 

In such case, multus should use cache data and try to remove interface without error, usually.  In addition, CNI Spec said that `Plugins should generally complete a DEL action without error even if some resources are missing.`, hence it should do so.

s1061123:
In cmdDel, CNI Spec mentioned that plugin should proceed cmdDel without any error, hence the change removes error returning
at cmdDel.

fix #822


dougbtv (approved):
LGTM!

PR #825, opened 2022-04-05
Fixes log message on CNI Check

dougbtv:
Should say `CheckNetwork` instead of `DelNetwork`


s1061123 (approved):
Looks good to me. Thanks! Merged.

PR #828, opened 2022-04-08
Add chroot option in multus-daemon

### LINKED ISSUE #829: '/var/run' should be '/run'.

**Why is this needed**:

s1061123:
This change introduces 'chroot' option in multus-daemon to execute
delegate CNI with host filesystem context.

fix #829

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -102,13 +102,14 @@ metadata:
 data:
   daemon-config.json: |
     {
+        "chrootDir": "/hostroot",
         "confDir": "/host/etc/cni/net.d",
         "logToStderr": true,
         "logLevel": "debug",
         "logFile": "/tmp/multus.log",
-        "binDir": "/host/opt/cni/bin",
-        "cniDir": "/host/var/lib/cni/multus",
-        "socketDir": "/host/var/run/multus/"
+        "binDir": "/opt/cni/bin",
+        "cniDir": "/var/lib/cni/multus",
+        "socketDir": "/host/run/multus/"
     }
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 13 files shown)


deployments/multus-daemonset-thick.yml
```diff
@@ -163,16 +164,19 @@ spec:
           volumeMounts:
             - name: cni
               mountPath: /host/etc/cni/net.d
-            - name: cnibin
```
dougbtv, line 166: Hey Tomo, why do we remove the bindir, but we keep the other directories? e.g. netns, confdir, etc.

deployments/multus-daemonset-thick.yml
```diff
@@ -198,6 +202,9 @@ spec:
         - name: cnibin
           hostPath:
             path: /opt/cni/bin
+        - name: hostdir
```
dougbtv, line 205: maybe a more specific name, `hostrootdir` or something, seems kind of overloaded between `hostPath` and `hostdir`

deployments/multus-daemonset-thick.yml
```diff
@@ -207,6 +214,9 @@ spec:
         - name: host-var-run
```
dougbtv, line 214: I realize this is unrelated to this changeset, but... I'm wondering if this /var/run mount is obsolete. I think we used to use it for a crio reboot hack? Maybe we can remove in 4.0

e2e/templates/multus-daemonset-thick.yml.j2
```diff
@@ -89,8 +89,8 @@ data:
         "logLevel": "debug",
         "logFile": "/tmp/multus.log",
         "binDir": "/host/opt/cni/bin",
```
dougbtv, line 91: should this still reference `/host/`?

deployments/multus-daemonset-thick.yml
s1061123: Thanks. Changed to hostroot.

deployments/multus-daemonset-thick.yml
s1061123: We still used to '/host/var/run/multus' for CNI cache. So we cannot remove it, but agreed that we could change to '/run'.
Let me add issue for that.

s1061123: Added as #829

deployments/multus-daemonset-thick.yml
s1061123: In this yaml case (multus-thick with chroot), when the delegate CNI is executed, multus changed the root to host root, with chroot(), hence, we don't need to configure binDir for pod mount namespace. We just configure host directory in this case. Of course, we just use multus-daemon without chroot option, user need to configure.

deployments/multus-daemonset-thick.yml
s1061123: It is taken care of the PR.

dougbtv (approved):
Thanks for the touch up on the directory as well as the explanations. LGTM!

PR #833, opened 2022-04-15
delete type conversion

jinglina:
remove redundant type conversion


s1061123 (approved):
Thank you so much for the PR! Looks good to me, let's merge it.

Issue #778, opened 2022-01-14
Multus doesn't support Kubernetes 1.21 or later (PodDevicesEntry.Data.PodDeviceEntries.DeviceIDs type has changed)


gmaquinay:
I've faced a problem, deploying multus with Kubernetes 1.21.
Checkpoint structure changed in Kube since commit a8b8995ef241e93e9486d475126450f33f24ef4e (2020 July 14th):
```
+// DevicesPerNUMA represents device ids obtained from device plugin per NUMA node id
+type DevicesPerNUMA map[int64][]string
+
 // PodDevicesEntry connects pod information to devices
 type PodDevicesEntry struct {
        PodUID        string
        ContainerName string
        ResourceName  string
-       DeviceIDs     []string
+       DeviceIDs     DevicesPerNUMA
        AllocResp     []byte
 }
```
DeviceIDs is not any more a slice of strings but its now a map:
```
// DevicesPerNUMA represents device ids obtained from device plugin per NUMA node id
type DevicesPerNUMA map[int64][]string
```

Thus, multus is not any more able to read `/var/lib/kubelet/device-plugins/kubelet_internal_checkpoint` and fails to provision requested network interfaces.

Is there a proper way to work around this? If not, I can propose a fix.
Please let me know.

Thx

zshi-redhat:
> 
> Thus, multus is not any more able to read `/var/lib/kubelet/device-plugins/kubelet_internal_checkpoint` and fails to provision requested network interfaces.
> 
> Is there a proper way to work around this? If not, I can propose a fix. Please let me know.

Did you try the kubelet API endpoint[1] path (default option) to get device ID? 
[1] https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/kubeletclient/kubeletclient.go#L31

I agree that the checkpoint file structure needs to be fixed in multus-cni.

s1061123:
Hi @gmaquinay thank you for the issue and offer to fix. Could you please propose this fix to us? I appreciate your help!

To work around the issue, you can use kubelet API endpoint as @zshi-redhat mentioned. Currently multus-cni and sr-iov device plugin prefers to use kubelet API endpoint to pass the device ID and checkpoint is the fallback method. So we recommend to enable  kubelet API endpoint to use multus-cni and sr-iov device plugin.

PR #834, opened 2022-04-18
Fix sr-iov support

### LINKED ISSUE #665: SR-IOV handling possibly breaks with K8s 1.21+ due to checkpoint file structure change
Following commit changed the type of the DeviceIDs parameter in Kubelet from []string to map[int][]string:
https://github.com/kubernetes/kubernetes/commit/a8b8995ef241e93e9486d475126450f33f24ef4e

As result I guess every project depending on directly parsing Kubelet's checkpoint file breaks when used with K8s 1.21+, because the file cannot be unmarshalled anymore into an old structure such as:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/a52680b3bfa026087879f89fae331b0dfc8dffe0/pkg/checkpoint/checkpoint.go#L36

At least mine sure as hell broke, thought I also let you know :)
As a reference here is the new format:
https://github.com/kubernetes/kubernetes/blob/a41a3ed4ae77a7902577bdb50b35fc5f42ed7c60/pkg/kubelet/cm/devicemanager/checkpoint/checkpoint.go#L34

s1061123:
Fix thick plugin daemonset to add volume mapping required for
sr-iov and fix code to update network status.
In addition, fix checkpoint structures to support K8s without
kubelet pod resources API.

fix https://github.com/k8snetworkplumbingwg/multus-cni/issues/665 and https://github.com/k8snetworkplumbingwg/multus-cni/issues/778

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -168,6 +168,10 @@ spec:
               mountPath: /host/run
             - name: host-var-lib-cni-multus
               mountPath: /var/lib/cni/multus
+            - name: host-var-lib-kubelet
+              mountPath: /var/lib/kubelet
+            - name: host-run-k8s-cni-cncf-io
+              mountPath: /run/k8s.cni.cncf.io
             - name: host-run-netns
               mountPath: /run/netns
               mountPropagation: HostToContainer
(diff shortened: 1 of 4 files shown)


dougbtv (approved):
LGTM! Thanks!

PR #835, opened 2022-04-18
Static pod e2e

nicklesimba:
Created a bash e2e test that creates a static pod, checks for net1 interface, and cleans up afterward.

### DIFF
--- a/e2e/setup_cluster.sh
+++ b/e2e/setup_cluster.sh
@@ -39,6 +39,12 @@ containerdConfigPatches:
 nodes:
   - role: control-plane
   - role: worker
+    kubeadmConfigPatches:
+    - |
+      kind: InitConfiguration
+      nodeRegistration:
+        kubeletExtraArgs:
+          pod-manifest-path: "/etc/kubernetes/manifests/"
   - role: worker
 EOF
 
--- a/e2e/static-pod-nad.yml
+++ b/e2e/static-pod-nad.yml
@@ -1,0 +1,15 @@
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: bridge-nad
+spec: 
+  config: '{
+    "cniVersion": "0.3.1",
+    "name": "testnet",
+    "type": "bridge",
+    "bridge": "testnet0",
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.10.0.0/16"
+    }
+}'
--- a/e2e/test-static-pod.sh
+++ b/e2e/test-static-pod.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -o errexit
+
+echo "Creating network attachment definition"
+kubectl create -f static-pod-nad.yml
+
+echo "Creating static pod config file"
+docker cp simple-static-pod.yml kind-worker:/etc/kubernetes/manifests/static-web.yaml
+
+echo "Waiting for static pod to start"
+kubectl wait --for=condition=Ready --namespace=default pod/static-web-kind-worker
+
+echo "Checking the pod annotation for net1 interface"
+kubectl exec static-web-kind-worker --namespace=default -- ip a show dev net1
+
+echo "Deleting static pod"
+docker exec kind-worker /bin/bash -c "rm /etc/kubernetes/manifests/static-web.yaml"
+
+echo "Deleting network attachment definition"
+kubectl delete -f static-pod-nad.yml
+
+echo "Test complete"
--- a/.github/workflows/kind-e2e.yml
+++ b/.github/workflows/kind-e2e.yml
@@ -35,6 +35,10 @@ jobs:
         working-directory: ./e2e
         run: ./test-simple-macvlan1.sh
 
+      - name: Test static pod
+        working-directory: ./e2e
+        run: ./test-static-pod.sh
+
       - name: Test default route1
         working-directory: ./e2e
         run: ./test-default-route1.sh
--- a/e2e/simple-static-pod.yml
+++ b/e2e/simple-static-pod.yml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: Pod
+metadata:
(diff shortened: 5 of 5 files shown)


e2e/static-pod-nad.yml
```diff
@@ -0,0 +1,16 @@
...
+  config: '{
+    "cniVersion": "0.3.1",
+    "name": "testnet",
+    "type": "bridge",
+    "bridge": "testnet0",
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.10.0.0/16"
+    }
+}'
```
s1061123, line 16: may need EOL (end of line).

e2e/static-pod-nad.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: bridge-nad
+  namespace: default
```
s1061123, line 5: we may not need 'namespace: default' because of 'default'.

e2e/test-static-pod.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -o errexit
+
+echo "Creating network attachment definition"
+kubectl create -f static-pod-nad.yml
+
+echo "Creating static pod config file"
+docker cp simple-static-pod.yml multus-whereabouts-worker:/etc/kubernetes/manifests/static-web.yaml
```
s1061123, line 8: `multus-whereabouts-worker` is not correct container name of multus e2e and that's why CI is failed.
Could you please check it?

e2e/test-static-pod.sh
```diff
@@ -0,0 +1,29 @@
...
+
+echo "Configuring kubelet on the node to use correct config file"
+docker cp static-pod-kubelet multus-whereabouts-worker:/etc/kubernetes/kubelet
+
+echo "Restarting kubelet"
+docker exec -it multus-whereabouts-worker /bin/bash -c "systemctl restart kubelet"
```
s1061123, line 14: To change the kubelet config, modify/restart could be a way, but I suppose we could configure these parameter at cluster creation (and it could be simple). This config does NOT conflict any test scenarios so we can configure it at the beginning of e2e testing. How about to modify 'setup_cluster.sh' to create kind cluster with this config?

Here are the pointers for that:
- https://kind.sigs.k8s.io/docs/user/configuration/#kubeadm-config-patches
- https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/
- https://github.com/kubernetes/release/blob/master/cmd/kubepkg/templates/latest/rpm/kubeadm/10-kubeadm.conf#L7-L9

e2e/test-static-pod.sh
```diff
@@ -0,0 +1,29 @@
...
+kubectl exec static-web-multus-whereabouts-worker --namespace=default -- ip a show dev net1
+
+echo "Deleting static pod"
+docker exec -it multus-whereabouts-worker /bin/bash -c "rm /etc/kubernetes/manifests/static-web.yaml"
+# https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/e2e/simple-pod.yml
+
+echo "Deleting network attachment definition"
+kubectl delete network-attachment-definition --namespace=default bridge-nad
+
+echo "Test complete"
```
s1061123, line 29: may need EOL (end of line).

e2e/static-pod-nad.yml
nicklesimba: I don't think the cluster defaults to the "default" namespace after creation. For example, on CI, after creating the cluster, the namespace is set to kind-kind. I think the net-attach-def then gets created in this namespace. For this reason I think line 5 may be needed.

e2e/setup_cluster.sh
```diff
@@ -39,6 +39,14 @@ containerdConfigPatches:
...
   - role: control-plane
   - role: worker
+    kubeadmConfigPatches:
+    - |
+      kind: InitConfiguration
+      nodeRegistration:
+        kubeletExtraArgs:
+          pod-manifest-path: "/etc/kubernetes/manifests/"
+          cluster-dns: "10.254.0.10"
+          cluster-domain: "kube.local"
```
s1061123, line 49: 'cluster-dns' and 'cluster-domain' might not required I suppose (pod-manifest-path is only required).
Is there any reason to add these?

e2e/test-static-pod.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Waiting for static pod to start"
+kubectl wait --for=condition=Ready --namespace=default pod/static-web-kind-worker
+
+echo "Checking the pod annotation for net1 interface"
+kubectl exec static-web-kind-worker --namespace=default -- ip a show dev net1
+
+echo "Deleting static pod"
+docker exec kind-worker /bin/bash -c "rm /etc/kubernetes/manifests/static-web.yaml"
+# https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/e2e/simple-pod.yml
```
s1061123, line 18: Can we remove this line? ('# https...')

e2e/test-static-pod.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Deleting network attachment definition"
+kubectl delete network-attachment-definition --namespace=default bridge-nad
```
s1061123, line 21: `kubectl delete -f static-pod-nad.yml` could be more consistent to `kubectl create -f static-pod-nad.yml`.

e2e/setup_cluster.sh
nicklesimba: Those params are in the static pod quickstart guide from official k8s docs. I'll try running the test without these parameters and see what happens...

nicklesimba: Seems like it works fine without, as you suggested.

s1061123 (approved):
It looks good to me. Thank you for incorporation of my comments! Merged.

PR #836, opened 2022-04-22
types: fix usage of strings.Split() for parsing CNI_ARGS

dcbw:
strings.Split() returns a slice, in this case with two elements of
the key and value. As such we shouldn't range over the slice when
the code is expecting a 2-element slice of key/value.

Otherwise we get errors for valid CNI_ARGS like:

```
2022-04-22T11:53:54Z [error] CreateCNIRuntimeConf: CNI_ARGS K8S_POD_NAMESPACE=openshift-etcd K8S_POD_NAMESPACE 17 is not recognized as CNI arg, skipped
2022-04-22T11:53:54Z [error] CreateCNIRuntimeConf: CNI_ARGS K8S_POD_NAMESPACE=openshift-etcd openshift-etcd 14 is not recognized as CNI arg, skipped
```

Fixes: d7d2a99ab5b8 ("Replace setenv with runtimeConfig set")

@s1061123 @dougbtv

### DIFF
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -219,7 +219,7 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 		},
 	}
 
-	// get CNI_ARGS and set it if it does not exist in rt.Args
+	// Populate rt.Args with CNI_ARGS if the rt.Args value is not set
 	cniArgs := os.Getenv("CNI_ARGS")
 	if cniArgs != "" {
 		for _, arg := range strings.Split(cniArgs, ";") {
@@ -232,14 +232,18 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 
 			envKey := string(keyval[0])
 			envVal := string(keyval[1])
-			isExists := false
-			for _, rtArg := range rt.Args {
-				if rtArg[0] == envKey {
-					isExists = true
+			found := false
+			for i := range rt.Args {
+				// Update existing key if its value is empty
+				if rt.Args[i][0] == envKey && rt.Args[i][1] == "" && envVal != "" {
+					logging.Debugf("CreateCNIRuntimeConf: add new val: %s", arg)
+					rt.Args[i][1] = envVal
+					found = true
+					break
 				}
 			}
-			if isExists != false {
-				logging.Debugf("CreateCNIRuntimeConf: add new val: %s", arg)
+			if !found {
+				// Add the new key if it didn't exist yet
 				rt.Args = append(rt.Args, [2]string{envKey, envVal})
 			}
 		}
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -223,24 +223,24 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
 	cniArgs := os.Getenv("CNI_ARGS")
 	if cniArgs != "" {
 		for _, arg := range strings.Split(cniArgs, ";") {
-			for _, keyval := range strings.Split(arg, "=") {
-				if len(keyval) != 2 {
-					logging.Errorf("CreateCNIRuntimeConf: CNI_ARGS %s %s %d is not recognized as CNI arg, skipped", arg, keyval, len(keyval))
-					continue
-				}
+			// SplitN to handle = within values, like BLAH=foo=bar
+			keyval := strings.SplitN(arg, "=", 2)
+			if len(keyval) != 2 {
+				logging.Errorf("CreateCNIRuntimeConf: CNI_ARGS %s %s %d is not recognized as CNI arg, skipped", arg, keyval, len(keyval))
+				continue
+			}
 
-				envKey := string(keyval[0])
-				envVal := string(keyval[1])
-				isExists := false
-				for _, rtArg := range rt.Args {
-					if rtArg[0] == envKey {
-						isExists = true
-					}
+			envKey := string(keyval[0])
+			envVal := string(keyval[1])
+			isExists := false
+			for _, rtArg := range rt.Args {
+				if rtArg[0] == envKey {
+					isExists = true
 				}
-				if isExists != false {
-					logging.Debugf("CreateCNIRuntimeConf: add new val: %s", arg)
-					rt.Args = append(rt.Args, [2]string{envKey, envVal})
-				}
+			}
+			if isExists != false {
+				logging.Debugf("CreateCNIRuntimeConf: add new val: %s", arg)
+				rt.Args = append(rt.Args, [2]string{envKey, envVal})
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/types/conf_test.go
```diff
@@ -639,6 +640,39 @@ var _ = Describe("config operations", func() {
...
+
+		os.Setenv("CNI_ARGS", "K8S_POD_NAME=dummy;K8S_POD_NAMESPACE=namespacedummy;K8S_POD_INFRA_CONTAINER_ID=123456789")
+		rt, _ := CreateCNIRuntimeConf(args, nil, "", &RuntimeConfig{}, nil)
+		fmt.Println("rt.ContainerID: ", rt.ContainerID)
+		Expect(rt.ContainerID).To(Equal("123456789"))
+		Expect(rt.NetNS).To(Equal(args.Netns))
+		Expect(rt.IfName).To(Equal(""))
+		Expect(rt.Args[0]).To(Equal([2]string{"K8S_POD_NAME", "dummy"}))
+		Expect(rt.Args[0]).To(Equal([2]string{"K8S_POD_NAMESPACE", "namespacedummy"}))
+		Expect(rt.Args[0]).To(Equal([2]string{"K8S_POD_INFRA_CONTAINER_ID", "123456789"}))
```
s1061123, line 673: rt.Args[] is formatted as https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/types/conf.go#L213-L219 hence the testing code might be following (I also add more ARGS test data in it)

```
                k8sArgs := &K8sArgs{K8S_POD_NAME: "dummy", K8S_POD_NAMESPACE: "namespacedummy", K8S_POD_INFRA_CONTAINER_ID: "123456789"}
                os.Setenv("CNI_ARGS", "K8S_POD_NAME=dummy;K8S_POD_NAMESPACE=namespacedummy;K8S_POD_INFRA_CONTAINER_ID=123456789;DUMMY_ARGS1=test;DUMMY_ARGS2=arg2key=arg2val")
                rt, _ := CreateCNIRuntimeConf(args, k8sArgs, "", &RuntimeConfig{}, nil)
                fmt.Println("rt.ContainerID: ", rt.ContainerID)
                Expect(rt.ContainerID).To(Equal("123456789"))
                Expect(rt.NetNS).To(Equal(args.Netns))
                Expect(rt.IfName).To(Equal(""))
                Expect(rt.Args[0]).To(Equal([2]string{"IgnoreUnknown", "true"}))
                Expect(rt.Args[1]).To(Equal([2]string{"K8S_POD_NAMESPACE", "namespacedummy"}))
                Expect(rt.Args[2]).To(Equal([2]string{"K8S_POD_NAME", "dummy"}))
                Expect(rt.Args[3]).To(Equal([2]string{"K8S_POD_INFRA_CONTAINER_ID", "123456789"}))
                Expect(rt.Args[4]).To(Equal([2]string{"K8S_POD_UID", ""}))
                Expect(rt.Args[5]).To(Equal([2]string{"DUMMY_ARGS1", "test"}))
                Expect(rt.Args[6]).To(Equal([2]string{"DUMMY_ARGS2", "arg2key=arg2val"}))
```

pkg/types/conf.go
```diff
@@ -223,25 +223,24 @@ func CreateCNIRuntimeConf(args *skel.CmdArgs, k8sArgs *K8sArgs, ifName string, r
...
-				for _, rtArg := range rt.Args {
-					if rtArg[0] == envKey {
-						isExists = true
-					}
-				}
-				if isExists != false {
-					logging.Debugf("CreateCNIRuntimeConf: add new val: %s", arg)
-					rt.Args = append(rt.Args, [2]string{envKey, envVal})
+			envKey := string(keyval[0])
+			envVal := string(keyval[1])
```
s1061123, line 233: This code could not parse `CNI_ARGS=ARGS1=ARG1VAL;ARG2=ARG2VAL1=ARG2VAL2` (well, current code does not support it, of course).

So how about that?

```
                        keyval := strings.Split(arg, "=")
                        if len(keyval) < 2 {
                                logging.Errorf("CreateCNIRuntimeConf: CNI_ARGS %s %s %d is not recognized as CNI arg, skipped", arg, keyval, len(keyval))
                                continue
                        }

                        envKey := string(keyval[0])
                        envVal := string(strings.Join(keyval[1:], "="))
```

s1061123 (commented):
Thank you for the fix! Nice catch the bugs.

I added some comments. Could you please take a look into it?

pkg/types/conf_test.go
dcbw: Yeah, good catch. I didn't run locally with root privs which would have caught this, I was depending on CI :)

pkg/types/conf.go
dcbw: Fixed, PTAL. Used strings.SplitN(..., 2) so we'll end up with ["BLAH", "foo=bar"] for those cases.

pkg/types/conf_test.go
```diff
@@ -639,6 +640,44 @@ var _ = Describe("config operations", func() {
...
+		Expect(rt.ContainerID).To(Equal("123456789"))
+		Expect(rt.NetNS).To(Equal(args.Netns))
+		Expect(rt.IfName).To(Equal(""))
+		fmt.Println("rt.ContainerID: ", rt.ContainerID)
+		Expect(rt.Args[0]).To(Equal([2]string{"IgnoreUnknown", "true"}))
+		Expect(rt.Args[1]).To(Equal([2]string{"K8S_POD_NAMESPACE", "namespacedummy"}))
+		Expect(rt.Args[2]).To(Equal([2]string{"K8S_POD_NAME", "dummy"}))
+		Expect(rt.Args[3]).To(Equal([2]string{"K8S_POD_INFRA_CONTAINER_ID", "123456789"}))
+		Expect(rt.Args[4]).To(Equal([2]string{"K8S_POD_UID", "aaaaa"}))
+		Expect(rt.Args[5]).To(Equal([2]string{"BLAHBLAH", "foo=bar"}))
```
s1061123, line 678: The following is required to pass the CI.

```
                Expect(rt.Args[1]).To(Equal([2]string{"K8S_POD_NAMESPACE", ""}))
                Expect(rt.Args[2]).To(Equal([2]string{"K8S_POD_NAME", ""}))
                Expect(rt.Args[3]).To(Equal([2]string{"K8S_POD_INFRA_CONTAINER_ID", ""}))
                Expect(rt.Args[4]).To(Equal([2]string{"K8S_POD_UID", ""}))
```

Otherwise we need to replace `k8sArgs := &K8sArgs{}` with `k8sArgs := &K8sArgs{K8S_POD_NAME: "dummy", K8S_POD_NAMESPACE: "namespacedummy", K8S_POD_INFRA_CONTAINER_ID: "123456789"}` because `CreateCNIRuntimeConf()` assumes that Kubernetes related CNI Args (which is in `K8sArgs{}`) is already retrieved at https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L563

Hence CreateCNIRuntimeConf() takes k8sArgs which already sets Kubernetes CNI_ARGS value.

s1061123 (changes requested):

pkg/types/conf_test.go
dcbw: Updated the PR; the code now replaces the K8S args if their value is empty, and it appends CNI_ARGS values that weren't already found in rt.Args. PTAL thanks!

s1061123 (approved):
Looks good to me, thanks!

PR #839, opened 2022-04-26
Refine Kubernetes APIs

s1061123:
This change bumps up Kubernetes API to 1.22.8 and make multus independent from k8s.io/kubernetes.


dougbtv (approved):
LGTM! Took it for a test drive in my lab, looking good. Interestingly enough the image size wasn't terribly improved, but, I"m happy to remove the dependency regardless -- thanks for cinching that up.

PR #841, opened 2022-05-03
The cachefile name should be the delegate configuration name

dougbtv:
It was previously using the net-attach-def name, which doesn't align with the cache file. Causing default-route selection to not succeed.

To reproduce the error, run through the steps in [the specifying a default route docs](https://github.com/k8snetworkplumbingwg/multus-cni/blob/130db696ca739fb3bcb550104ace34f03c00a360/docs/how-to-use.md#specifying-a-default-route-for-a-specific-attachment).

Previous to this fix, it should fail with:

```
error deleting default gateway in cache: open /var/lib/cni/multus/results/macvlan-conf-1b516b1ef0c52920145b3502a71182c92858a542067ea1f134d0dcd27e61f3c4-net1: no such file or directory
```

In this case the string `macvlan-conf` comes from the net-attach-def name, but it should come from the delegate name in order to match the cache filename on disk.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -617,14 +617,14 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			}
 		}
 
-		netName := ""
+		// We collect the delegate netName for the cachefile name as well as following errors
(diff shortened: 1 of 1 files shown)


s1061123 (approved):
LGTM, thanks!

PR #844, opened 2022-05-10
Fix wrong location for daemonset in readme

danielmellado:
This PR fixes an outdated reference in the docs where
multus-daemonset.yml was expected to be found under the images folder.


maiqueb (approved):
Good catch.

PR #846, opened 2022-05-11
Bumps net-attach-def client library to v1.1.1

dougbtv:
This accounts for CNI v1.0 formatted IP address information (no longer requires version field)


dougbtv:
First attempt is a mess... looking into it.

dougbtv:
OK, looks better. Gist was that the original net-attach-def client library branch point was at the wrong point for where the fix was made, so the net-attach-def client library release-v1.1 has been updated to the proper point.

PR #851, opened 2022-05-25
Added hyperlink for logging

nicklesimba:
This may make it easier for users to find info on logging parameters since there isn't a dedicated docs page for it.


dougbtv (approved):
LGTM! (I'll squash it)

PR #856, opened 2022-06-03
config, logging: correct the logging parameters

maiqueb:
The logging parameters were listed using uppercase, which is wrong. 

According to multus configuration, they should be in camelCase:
- [logFile]( https://github.com/k8snetworkplumbingwg/multus-cni/blob/779170a48ea767fe367f58ca472a8c403c203c6f/pkg/types/types.go#L45)
- [logLevel](https://github.com/k8snetworkplumbingwg/multus-cni/blob/779170a48ea767fe367f58ca472a8c403c203c6f/pkg/types/types.go#L46)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -110,7 +110,7 @@ Optionally, you may have Multus log to a file on the filesystem. This file will 
 For example in your CNI configuration, you may set:
 
 ```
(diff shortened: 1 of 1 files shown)


s1061123 (approved):
LGTM.

dougbtv (approved):

PR #857, opened 2022-06-03
Add metrics support

s1061123:
This change introduces prometheus metric exporter support for multus-daemon, thick plugin.


dougbtv (approved):
LGTM!

PR #859, opened 2022-06-07
Fixed that in.Delegates may remain in the CmdDel

lujinda:
Fixed the possibility that in.Delegates may remain in the CmdDel method when the Pod does not use the cluster network.
For example:
If the cluster network is of type `conflist`, the network specified by `v1.multus-cni.io/default-network` in the Pod is of type `conf`.  The following code will have a problem：
```Go
		if err := json.Unmarshal(netconfBytes, &in.Delegates); err != nil {
			return cmdErr(k8sArgs, "failed to load netconf: %v", err)
		}
```

`Umarshal` will not clear the existing data in Delegates first. CmdDel will report an error:
2022-06-06T17:03:58+08:00 [error] conflistDel: error converting the raw bytes into a conflist: error parsing configuration list: no 'plugins' key
2022-06-06T17:03:58+08:00 [error] delegateDel: error invoking ConflistDel - "xxx": conflistDel: error converting the raw bytes into a conflist: error parsing configuration list: no 'plugins' key

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -34,14 +34,15 @@ import (
 	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	nadutils "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils"
 	"github.com/vishvananda/netlink"
-	k8s "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/k8sclient"
-	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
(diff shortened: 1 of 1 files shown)


lujinda:
Add a Demo that describes this problem：
```Go
type Delegate struct {
	A string
	B string
}

func main() {
	objs := []*Delegate{
		&Delegate{A: "aaaa"},
		&Delegate{A: "aaaaaaaa"},
	} // cluster network NetCon
	body := `[{"B": "bbbb"}, {"B": "bbbbbbbbb"}]` // scratched NetCon
	if err := json.Unmarshal([]byte(body), &objs); err != nil {
		panic(err)
	}
	fmt.Printf("%+v, %+v\n", objs[0], objs[1])
}
```

Output:
> &{A:aaaa B:bbbb}, &{A:aaaaaaaa B:bbbbbbbbb}

s1061123:
Thank you submit the PR, @lujinda 

LGTM, merged.

PR #863, opened 2022-06-14
Added Multus Conf Dir to AutoConfig

naeemarsalan:
Feature:
- Added cli arg to pass "confDir" to autoconfig

This is required when using distro like RKE2/K3s which has different path for cni like multus. These distro use path: /etc/cni/net.d/ and currently with autoconfig mode uses "/etc/cni/multus/". 

The current fix for different is to use a sym link with autoconfig mode. 
ln -s /etc/cni/net.d/ /etc/cni/multus/

Signed-off-by: arsalan.naeem@suse.com

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -15,6 +15,7 @@ CNI_CONF_DIR="/host/etc/cni/net.d"
 CNI_BIN_DIR="/host/opt/cni/bin"
 ADDITIONAL_BIN_DIR=""
 MULTUS_CONF_FILE="/usr/src/multus-cni/images/70-multus.conf"
+MULTUS_CONF_DIR="/etc/cni/multus/net.d"
 MULTUS_AUTOCONF_DIR="/host/etc/cni/net.d"
 MULTUS_BIN_FILE="/usr/src/multus-cni/bin/multus"
 MULTUS_KUBECONFIG_FILE_HOST="/etc/cni/net.d/multus.d/multus.kubeconfig"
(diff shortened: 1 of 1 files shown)


images/entrypoint.sh
```diff
@@ -403,6 +408,7 @@ EOF
         {
           $CNI_VERSION_STRING
           "name": "$MASTER_PLUGIN_NET_NAME",
+          "confDir": "$MULTUS_CONF_DIR",
```
dougbtv, line 411: I'd prefer that we have it like below for `$NESTED_CAPABILITIES_STRING` or `$ISOLATION_STRING` where it doesn't populate the `confDir` always, but only populates it if the `--multus-conf-dir` is used.

images/entrypoint.sh
```diff
@@ -130,6 +132,9 @@ while [ "$1" != "" ]; do
         --multus-conf-file)
             MULTUS_CONF_FILE=$VALUE
             ;;
+        --multus-conf-dir)
```
dougbtv, line 135: Let's also add a paragraph in the how-to-use doc in this section for the parameter: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#entrypoint-script-parameters

s1061123:
First, thank you for the PR, @naeemarsalan !

> The current fix for different is to use a sym link with autoconfig mode.
> ln -s /etc/cni/net.d/ /etc/cni/multus/

I'm wondering why you mentioned such symbolic link because multus-cni (entrypoint.sh) does not do such things. 
Is there some deployments that does `ln -s /etc/cni/net.d/ /etc/cni/multus/`?

images/entrypoint.sh
naeemarsalan: That makes more sense. I would agree.

images/entrypoint.sh
naeemarsalan: Sure, will update the PR.

naeemarsalan:
> First, thank you for the PR, @naeemarsalan !
> 
> > The current fix for different is to use a sym link with autoconfig mode.
> > ln -s /etc/cni/net.d/ /etc/cni/multus/
> 
> I'm wondering why you mentioned such symbolic link because multus-cni (entrypoint.sh) does not do such things. Is there some deployments that does `ln -s /etc/cni/net.d/ /etc/cni/multus/`?

This was a suggested fix if someone else came across the same problem. When using Multus with different K8s (k3s/rke2) this can be quite frustrating, as you are not sure where "/etc/cni/multus/" is coming from.

PR #864, opened 2022-06-16
Use *[]net.IP for 'default-route' network selection element.

s1061123:
Allows for empty default route for IPv6, e.g. to select the default-route for a given interface but to pass an empty list.


maiqueb:
> Can you add a description to this PR so people can understand **what** this is fixing, and **why** that is important ?

@s1061123 / @dougbtv ^

dougbtv (approved):

PR #867, opened 2022-06-21
Fix missing device-info in networks-status annotation for chained plugins

mmirecki:
When chained plugins are used, the networks-status annotations is missing the device-info section.
This PR aims to fix this.


dougbtv:
Tests don't appear to be updated on re-running (we patched a test in master). I'm attempting to change the target branch, and then change it back.

dougbtv (approved):
LGTM!

PR #869, opened 2022-06-22
supports all capabilities for cluster networks

lujinda:
Our container runtime supports annotation capability, can pass Pod's annotation to CNI. But Multus only supports portMappings/Bandwidth capability for cluster networks. So I submitted this PR, hope it can be adopted.
I save the original runtimeConfig passed from the container runtime in CapabilityArgs. CNI can decide whether a certain value is required through the capabilities rule

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -452,7 +452,7 @@ func delPlugins(exec invoke.Exec, pod *v1.Pod, args *skel.CmdArgs, k8sArgs *type
 	var errorstrings []string
 	for idx := lastIdx; idx >= 0; idx-- {
 		ifName := getIfname(delegates[idx], args.IfName, idx)
-		rt, cniDeviceInfoPath := types.CreateCNIRuntimeConf(args, k8sArgs, ifName, netRt, delegates[idx])
+		rt, cniDeviceInfoPath := types.CreateCNIRuntimeConf(args, k8sArgs, ifName, netRt, multusNetconf.WholeRuntimeConfig, delegates[idx])
 		// Attempt to delete all but do not error out, instead, collect all errors.
 		if err := delegateDel(exec, pod, delegates[idx], rt, multusNetconf); err != nil {
 			errorstrings = append(errorstrings, err.Error())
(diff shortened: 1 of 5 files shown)


s1061123:
Hi @lujinda , thank you for the PR.

I suppose we should not implement to bring all capabilities to multus interface because some runtimeConfig/capabilities should be only for cluster network (or unique to pod network). Currently container runtime cannot define which runtimeConfig goes to which network interface, hence one runtimeConfig is used for ALL network interface (i.e. eth0, net1, net2 and so on).
So, for example, interface specific runtimeConfig parameters, such as 'mac', 'ips' and 'device id', cannot be used if we delivers all runtimeConfig.

So first, we may need to discuss

- how to config runtimeConfig per interface (i.e. cluster network interface and multus secondary network interface)
- how to deliver it and so on
(this may need to change CRI and Kubernetes, so it seems to be big changes and require discussion to Kubernetes communitiy)

So if you don't mind it, let me close this PR and let's discuss about it in [NPWG meeting](https://github.com/k8snetworkplumbingwg/community), will be held at June 30th.

s1061123:
> It is also all Cluster Networks that pass the same capabilityArgs. Which RuntimeConfig is finally passed to the cluster network is actually affected by the capabilities rules, and not every CNI receives all the RuntimeConfigs.

So if cluster network and net-attach-def has same capabilities (e.g. 'ips', 'mac' and 'bandwidth'), let's imagine what's happen. Then cluster network interface and net-attach-def interface try to use same parameters. It is not accepted. So we may need to define how to supply these information per interface (cluster network and net-attach-def).

lujinda:
> > It is also all Cluster Networks that pass the same capabilityArgs. Which RuntimeConfig is finally passed to the cluster network is actually affected by the capabilities rules, and not every CNI receives all the RuntimeConfigs.
> 
> So if cluster network and net-attach-def has same capabilities (e.g. 'ips', 'mac' and 'bandwidth'), let's imagine what's happen. Then cluster network interface and net-attach-def interface try to use same parameters. It is not accepted. So we may need to define how to supply these information per interface (cluster network and net-attach-def).

Thx. I understand what you mean. Or is it possible to skip Well-known Capabilities? what do you think?

s1061123:
I believe we don't do ad-hoc implement. We should discuss how we solve in NPWG community because that's why NPWG works for and that's why NPWG have multi-net-spec and multus (as a reference implementation).

lujinda:
> I believe we don't do ad-hoc implement. We should discuss how we solve in NPWG community because that's why NPWG works for and that's why NPWG have multi-net-spec and multus (as a reference implementation).

OK. Thank you very much for your reply :)

dougbtv:
@lujinda if you'd like to bring this to the community, we have connection information for our meetings @ https://github.com/k8snetworkplumbingwg/community

thank you for the contribution and conversation!

PR #876, opened 2022-07-06
Update golang version to 1.18.x

### LINKED ISSUE #826: update golang to 1.17 or 1.18
update golang to 1.17
**Why is this needed**:
Since golang 1.18 is released, We should update golang to 1.17 or 1.18?

s1061123:
fix #826


dougbtv (approved):
LGTM!

PR #877, opened 2022-07-06
Refine build-go.sh and update 'version' output

s1061123:


hack/build-go.sh
```diff
@@ -7,54 +7,55 @@ if [ ! -d ${DEST_DIR} ]; then
 	mkdir ${DEST_DIR}
 fi
 
-# Add version/commit/date into binary
-# In case of TravisCI, need to check error code of 'git describe'.
-if [ -z "$VERSION" ]; then
+# version information
+hasGit=true
+git version > /dev/null 2>&1 || hasGit=false
```
dougbtv, line 12: Thanks this is the main thing I was looking for, making sure it work w/o git, also, if it's not a git repository, I wonder what happens. Maybe not critical for now, though.

dougbtv (approved):
LGTM

hack/build-go.sh
s1061123: Here is the output without git. It is not perfect, but it seems to be okey...

```
$ ./bin/multus --version
multus: version:(), commit:3deb079546629300dbfff8c4c8fbe032b388fd9d, date:2022-07-13T06:01:09+00:00
$ ./bin/multus-shim --version
multus-shim: version:(), commit:3deb079546629300dbfff8c4c8fbe032b388fd9d, date:2022-07-13T06:01:09+00:00
$ ./bin/multus-daemon -version
multus-daemon: version:(), commit:3deb079546629300dbfff8c4c8fbe032b388fd9d, date:2022-07-13T06:01:09+00:00
```

PR #878, opened 2022-07-08
write output file for generated kubeconfig

tomkukral:
This change will make debugging easier in cases when non-default
location is used and user forgot to mount new directory.


dougbtv (approved):
Nice one Tom! This is a great enhancement, way better to have the path here. Thanks -- LGTM

Issue #880, opened 2022-07-15
python3 is not properly detected


chrisplo:
**What happend**:
python3 is not discovered properly

**What you expected to happen**:
python3 is recognized

**How to reproduce it (as minimally and precisely as possible)**:
enter container and run the entrypoint.sh python3 check directly

```
[root@614d47a87ac9 /]# if ! type python3 &> /dev/null; then echo python3; else echo oh python2; fi
oh python2
```

**Anything else we need to know?**:

**Environment**:

- Multus version 3.8
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`):
- Primary CNI for Kubernetes cluster:
- OS (e.g. from /etc/os-release):
- File of '/etc/cni/net.d/'
- File of '/etc/cni/multus/net.d'
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)
- Target pod yaml info (with annotation, use `kubectl get pod <podname> -o yaml`)
- Other log outputs (if you use multus logging)

chrisplo:
i think also the alias isn't being read from the tempfile

dougbtv:
Currently, the default image that we're using in the community build has only python3 and not python2. So this would be expected as it stands now.

You can see we're using `centos:centos7` @ https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/Dockerfile#L10

So this should be expected. However, if you would like to build using another base image, it would detect python3 instead.

Is there a reason you require python3 or are you using another image?

PR #881, opened 2022-07-15
Detect python3

chrisplo:
fixes https://github.com/k8snetworkplumbingwg/multus-cni/issues/880

Signed-off-by: <chrisplo@cisco.com>

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -83,7 +83,7 @@ function warn()
     log "WARN: {$1}"
 }
 
-if ! type python3 &> /dev/null; then
+if type python3 &> /dev/null; then
 	alias python=python3
 fi
 


chrisplo:
need to also deal with the temp file that doesn't eval the alias

dougbtv (approved):
Looks like the correct logic to me, Chris. Thanks for the contribution!

```
$ which python3
/usr/bin/python3

$ cat /tmp/foo.sh 
#!/bin/bash
if ! type python3 &> /dev/null; then
  echo "is python 3"
fi

$ /tmp/foo.sh 

$ vi /tmp/foo.sh 

$ cat /tmp/foo.sh 
#!/bin/bash
if type python3 &> /dev/null; then
  echo "is python 3"
fi

$ /tmp/foo.sh 
is python 3
```

Let us know about the temp file you're mentioning? Or file an issue or what not. Thanks!

PR #882, opened 2022-07-22
Support modifying the CNI configuration file name

lujinda:


s1061123:
I'm not clear that why we need the option. Could you please let me know the usecase?

lujinda:
> I'm not clear that why we need the option. Could you please let me know the usecase?

Thank you for your reply. In our scenario, we cannot assume that `00-multus.conf` is the first configuration. In some of our scenarios, the first configuration can be named `00-cni.conf`

maiqueb (commented):
Overall the changes look good, I would just like to say this only does it for the thick-plugin architecture, not to the standalone alone.

You'd need to check the [entrypoint script](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/entrypoint.sh) for the standalone deployment afaiu.

PR #890, opened 2022-08-08
Add delegate API in multus-daemon

s1061123:
This changes introduce delegate API function in multus-daemon.
This API will be consumed from other programs for hot-plug interface into running pod. This change also cleanups server code to split into client code and server code to easy to import from other golang code.


dougbtv (approved):
LGTM! Tomo and I had a working session to review prior to the merge. Thanks Tomo.

PR #893, opened 2022-08-18
Feature/multus 4.0

s1061123:


dougbtv:
note: we don't agree with lgtm-com on this one. we need to do what we're doing there.

dougbtv (approved):
LGTM. Of note, these changes were reviewed while being pulled into a feature branch.

Also, manually verified on my local machine, thanks Tomo!

PR #897, opened 2022-08-22
Update deployment yaml

s1061123:
This change adds note section in deployment yaml to explicitly mention the scope of this yaml.

### DIFF
--- a/deployments/multus-daemonset-crio.yml
+++ b/deployments/multus-daemonset-crio.yml
@@ -1,3 +1,11 @@
+# Note:
+#   This deployment file is designed for 'quickstart' of multus, easy installation to test it,
+#   hence this deployment yaml does not care about following things intentionally.
+#     - various configuration options
+#     - minor deployment scenario
+#     - upgrade/update/uninstall scenario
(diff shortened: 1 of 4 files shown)


deployments/multus-daemonset-crio.yml
```diff
@@ -1,3 +1,11 @@
+# Note:
+#   This deployment file is designed for 'quickstart' of multus, easy installation to test it,
+#   hence this deployment yaml does not care about following things as intentionally.
```
dougbtv, line 3: Remove `as`

deployments/multus-daemonset-crio.yml
```diff
@@ -1,3 +1,11 @@
...
+#     - various configuration options
+#     - minor deployment scenario
+#     - upgrade/update/uninstall scenario
+#   Multus team understand users deployment scenarios are diverse, hence we does not cover
+#   comprehensive deployment scenario. We expects that it is covered by each platform deployment.
```
dougbtv, line 8: `s/expects/expect`

deployments/multus-daemonset-crio.yml
s1061123: fixed.

PR #904, opened 2022-08-26
[backport v3] Change container image to fedora from centos

dougbtv:
This change introduce fedora container for multus image.


s1061123:
LGTM, merged!

PR #906, opened 2022-08-30
fix multiple default gw

TimFroidcoeur:
when the configuration specifies both an IPv4 and IPv6 default route,
the IsFilterV4Gateway and IsFilterV6Gateway flags should both be false,
to allow the gateway configuration.
The logic in CheckGatewayConfig would do the inverse, setting both to
true in case of both IPv4 and IPv6 gateway.

Fixes: d52f2b6a ("Update libcni cache when default-route net selection
is used")
Signed-off-by: Tim Froidcoeur <tim.froidcoeur@tessares.net>

### DIFF
--- a/pkg/types/conf_test.go
+++ b/pkg/types/conf_test.go
@@ -928,7 +928,12 @@ var _ = Describe("config operations", func() {
 
 		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
 		Expect(err).NotTo(HaveOccurred())
+
+		Expect(CheckGatewayConfig([]*DelegateNetConf{netconf})).To(Succeed())
+
 		Expect(netconf.GatewayRequest).To(BeNil())
+		Expect(netconf.IsFilterV4Gateway).To(BeTrue())
+		Expect(netconf.IsFilterV6Gateway).To(BeTrue())
 	})
 
 	It("test LoadDelegateNetConf keeps empty GatewayRequest", func() {
@@ -950,8 +955,13 @@ var _ = Describe("config operations", func() {
 
 		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
 		Expect(err).NotTo(HaveOccurred())
+
+		Expect(CheckGatewayConfig([]*DelegateNetConf{netconf})).To(Succeed())
+
 		Expect(netconf.GatewayRequest).NotTo(BeNil())
-		Expect(len(*netconf.GatewayRequest)).To(BeEquivalentTo(0))
+		Expect(len(*netconf.GatewayRequest)).To(Equal(0))
+		Expect(netconf.IsFilterV4Gateway).To(BeTrue())
+		Expect(netconf.IsFilterV6Gateway).To(BeTrue())
 	})
 
 	It("test LoadDelegateNetConf keeps GatewayRequest", func() {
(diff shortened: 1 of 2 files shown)


s1061123:
> when the configuration specifies both an IPv4 and IPv6 default route,
 
Could you please share the sample pod/net-attach-def config in your case, to understand it clearly?

dougbtv:
Thanks Tim. Looks reasonable at a glance for sure, thanks for providing a fix.

Also, maybe we should get some test coverage for this? Maybe in conf_test.go @ https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/types/conf_test.go

TimFroidcoeur:
hi, thnx for looking at it.

@s1061123 in our setup, this annotation triggered the issue when upgrading multus from v3.8.1:
```
k8s.v1.cni.cncf.io/networks:
        [{"default-route": ["10.0.4.2", "2001:2:4::2"], "interface": "main", "ips": ["10.0.4.1/24", "2001:2:4::1/64"], "name": "main-net"}]
```

@dougbtv I'll soon update the PR with extra tests.

pkg/types/conf_test.go
```diff
@@ -928,7 +928,13 @@ var _ = Describe("config operations", func() {
 
 		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
 		Expect(err).NotTo(HaveOccurred())
+
+		err = CheckGatewayConfig([]*DelegateNetConf{netconf})
+		Expect(err).NotTo(HaveOccurred())
+
 		Expect(netconf.GatewayRequest).To(BeNil())
+		Expect(netconf.IsFilterV4Gateway).To(BeEquivalentTo(true))
+		Expect(netconf.IsFilterV6Gateway).To(BeEquivalentTo(true))
```
maiqueb, line 937: IMO you should use the `BeTrue()` matcher instead. More idiomatic.

The same comment applies below to all other tests.

pkg/types/conf_test.go
```diff
@@ -973,8 +985,42 @@ var _ = Describe("config operations", func() {
 
 		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
 		Expect(err).NotTo(HaveOccurred())
+
+		err = CheckGatewayConfig([]*DelegateNetConf{netconf})
+		Expect(err).NotTo(HaveOccurred())
+
 		Expect(netconf.GatewayRequest).NotTo(BeNil())
 		Expect(len(*netconf.GatewayRequest)).To(BeEquivalentTo(1))
+		Expect(netconf.IsFilterV4Gateway).To(BeEquivalentTo(false))
```
maiqueb, line 994: I think you should use the `BeFalse()` matcher.

pkg/types/conf_test.go
TimFroidcoeur: thnx, good point

pkg/types/conf_test.go
```diff
@@ -928,7 +928,13 @@ var _ = Describe("config operations", func() {
 
 		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
 		Expect(err).NotTo(HaveOccurred())
+
+		err = CheckGatewayConfig([]*DelegateNetConf{netconf})
+		Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 933: nit: You could also inline these two - I think it makes more sense to use positive logic, unless it is impossible.
```golang
Expect(CheckGatewayConfig([]*DelegateNetConf{netconf})).To(Succeed())
```

pkg/types/conf_test.go
```diff
@@ -950,8 +956,14 @@ var _ = Describe("config operations", func() {
...
```
maiqueb, line 961: ditto.

pkg/types/conf_test.go
```diff
@@ -973,8 +985,42 @@ var _ = Describe("config operations", func() {
...
```
maiqueb, line 990: ditto.

pkg/types/conf_test.go
```diff
@@ -973,8 +985,42 @@ var _ = Describe("config operations", func() {
...
+				"name": "weave-list",
+				"plugins": [ {"type" :"weave"} ]
+			}]
+		}`
+
+		nsJSON := `{ "name": "foobar", "default-route": [ "10.1.1.1", "fc00::1" ] }`
+		ns := &NetworkSelectionElement{}
+
+		err := json.Unmarshal([]byte(nsJSON), ns)
+		Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 1013: ditto

maiqueb (changes requested):
You can ignore the nits below, but there's one error you forgot to check (and and `err` that is checked twice - I think your code just has to be moved aroud a bit).

Sorry for only having noticed it now.

pkg/types/conf_test.go
```diff
@@ -973,8 +985,42 @@ var _ = Describe("config operations", func() {
...
+
+		netconf, err := LoadDelegateNetConf([]byte(conf), ns, "", "")
+		err = CheckGatewayConfig([]*DelegateNetConf{netconf})
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(netconf.GatewayRequest).NotTo(BeNil())
+		Expect(len(*netconf.GatewayRequest)).To(BeEquivalentTo(2))
```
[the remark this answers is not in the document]
maiqueb: Oh, better yet, use the `HaveLen` matcher:
```golang
Expect(*netconf.GatewayRequest).To(HaveLen(2))
```
The error msg it throws is a lot more precise.

dougbtv (approved):
Before applying this change...

```
[fedora@kube-master1 ~]$ cat pod.yml 
apiVersion: v1
kind: Pod
metadata:
  name: samplepod2
  annotations:
    # k8s.v1.cni.cncf.io/networks: whereabouts-conf
    k8s.v1.cni.cncf.io/networks: >
      [{"default-route": ["10.0.4.2", "2001:2:4::2"], "interface": "main", "name": "macvlan-conf"}]
spec:
  containers:
  - name: samplepod2
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: alpine
```

Pod launches, but, no default routes take place.

```
[fedora@kube-master1 ~]$ kubectl exec -it samplepod2 -- ip route
10.244.0.0/16 via 10.244.1.1 dev eth0 
10.244.1.0/24 dev eth0 scope link  src 10.244.1.22 
192.168.1.0/24 dev main scope link  src 192.168.1.202 
```

After, pod does not launch, but it's because it can't reach the network ;) So... I think this is correct behavior.

```
ERRORED: error configuring pod [default/samplepod2] networking: Multus: [default/samplepod2/075a3947-e384-492e-9955-9002c960138e]: error setting default gateway: no route to host
```

Thanks for taking into consideration Miguel's feedback! And for the tests, and the contribution :+1:

PR #908, opened 2022-09-01
Idiomatic config unit tests

maiqueb:
Replaces the existing matchers by a more idiomatic approach, which return better error msgs when the test fails.

The following matchers are used:
- `BeTrue`
- `BeFalse`
- `HaveLen`
- `BeEmpty`

Furthermore, they make the test easier to read.


dougbtv (approved):
LGTM!

PR #915, opened 2022-09-20
Fix deleate API to load conflist

### LINKED ISSUE #914: delegate add endpoint: does not handle `NetConfList` CNI configurations
**What happend**:
When trying to invoke the `delegate` endpoint with a CNI configuration featuring `NetConfList` throws the following error:
```
error configuring pod [default/macvlan1-worker1] networking: error in converting the raw bytes to conf: error parsing configuration: missing 'type'
```

The requested delegate should be added.

Try the [sample delegate project](https://github.com/s1061123/multus-delegate-sample) to add a delegate with the following configuration:
```yaml
metadata:
  name: macvlan1-config
spec:
  config: '{
            "cniVersion": "0.4.0",
            "plugins": [
                {
                    "type": "macvlan",
                    "capabilities": { "ips": true },
                    "master": "eth1",
                    "mode": "bridge",
                    "ipam": {
                        "type": "static"
                    }
                }, {
                    "type": "tuning"
                } ]
        }'
```

The pod spec I've used looks like:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: macvlan1-worker1
  annotations:
    k8s.v1.cni.cncf.io/networks: '[
            {
                "name": "macvlan1-config",
                "ips": [ "10.1.1.11/24" ]
            }
    ]'
  labels:
    app: macvlan
spec:
  containers:
    - name: macvlan-worker1
      image: docker.io/library/alpine:latest
      command: ["/bin/sleep", "10000"]
```

When invoking the same endpoint with a configuration using a `NetConf` CNI configuration, it **does** work as expected.

This happens because the server code **always** [unmarshalls the configuration](https://github.com/k8snetworkplumbingwg/multus-cni/blob/505ab4567c743043cba534c61981c508180fbfd3/pkg/server/server.go#L95) as `NetConf`. 

It should check if the configuration is a `NetConfList` or `NetConf` then unmarshal into the corresponding struct.

**Environment**:
Multus in thick plugin mode

s1061123:
fix #914

### DIFF
--- a/pkg/server/api/api.go
+++ b/pkg/server/api/api.go
@@ -69,7 +69,7 @@ func DoCNI(url string, req interface{}, socketPath string) ([]byte, error) {
 
 // GetAPIEndpoint returns endpoint URL for multus-daemon
 func GetAPIEndpoint(endpoint string) string {
-	return fmt.Sprintf("http://dummy/%s", endpoint)
+	return fmt.Sprintf("http://dummy%s", endpoint)
 }
 
 // CreateDelegateRequest creates Request for delegate API request
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -446,11 +446,10 @@ func cmdDelegateAdd(cmdArgs *skel.CmdArgs, k8sArgs *types.K8sArgs, exec invoke.E
 		return nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


dougbtv (approved):
LGTM!

Issue #891, opened 2022-08-10
Multus does not send some deletes to cni plugin after a reboot of the node


saranshan2603:
**What happend**:

The node was rebooted and the pods existing in that node went it to terminating and moved to another node. When the node came back up, deletes are sent for the containers that moved to another node to the cni-plugin from multus. I see deletes of some containers are missing.

Following is a snapshot from multus-lumberjack logs

 [error] Multus: failed to get the cached delegates file: open /var/lib/cni/multus/80c8f5d1c4727793b987b5592bd0a3228cb2cdde8f292804b404fb439b379a99: no such file or directory, cannot properly delete

**What you expected to happen**:
All stale containers multus should send a delete to the cni plugin

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Multus version 
  latest
- Kubernetes version (use `kubectl version`):
 openshift 4.10
- Primary CNI for Kubernetes cluster:
calico

dougbtv:
@saranshan2603 this I believe is a functionality of CNI itself, and not of Multus. So, Multus can only react to what's been sent by CNI.

The CNI DEL for Multus is done as a kind of best effort, Multus tries to use what's available, and if resources are missing -- it quietly moves on without erroring (errr, well at least exiting zero)

See also: https://github.com/containernetworking/cni/blob/main/SPEC.md#del-remove-container-from-network-or-un-apply-modifications

> Plugins should generally complete a DEL action without error even if some resources are missing. 

In this case, the cache is missing, so Multus has missing resources.

PR #922, opened 2022-10-09
delegate plugin delete success, delete cache file

gaopeiliang:
this case should try max to cleanup 

1.  sandBox container stop times, but delegated plugin failed (CNI server maybe in starting .... ) , but cache file has been deleted
2.  container gc manager clean up forever even though POD has been fully deleted from ETCD

POD not find && cache file not find;  never clean up, we should try max ...

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -37,7 +37,8 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
     }, {
         "type": "macvlan",
         ... (snip)
-    }]
+    }],
+    allowTryDeleteOnErr: false
 }
 ```
 
@@ -62,6 +63,7 @@ User should chose following parameters combination (`clusterNetwork`+`defaultNet
 * `systemNamespaces` ([]string, optional): list of namespaces for Kubernetes system (namespaces listed here will not have `defaultNetworks` added)
 * `multusNamespace` (string, optional): namespace for `clusterNetwork`/`defaultNetworks`
 * `delegates` ([]map,required): number of delegate details in the Multus
+* `allowTryDeleteOnErr` (bool, optional): Enable or disable delegate DEL message to next when some missing error. Defaults to false.
 
 ### Network selection flow of clusterNetwork/defaultNetworks
 
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -825,6 +825,7 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 
 	// Read the cache to get delegates json for the pod
 	netconfBytes, path, err := consumeScratchNetConf(args.ContainerID, in.CNIDir)
+	removeCacheConf := false
 	if err != nil {
 		// Fetch delegates again if cache is not exist and pod info can be read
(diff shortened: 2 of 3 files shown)


s1061123:
hmm... It may seems reasonable, but need to check the CNI spec.

In CNI spec, https://github.com/containernetworking/cni/blob/main/SPEC.md#del-remove-container-from-network-or-un-apply-modifications, there is following sentense.

> Plugins should generally complete a DEL action without error even if some resources are missing. 

So CNI spec does not assume retry several times from CNI plugin to cleanup. But you faces such issue hence you file a PR, I guess.

So how about to

- add optional config to `multus.conf` to change the behavior? (e.g. `allowRetryDelete` or some fancy name)
- If you don't mind, could you please let me know which CNI plugin does retry remove?

s1061123:
> "Plugins should generally complete a DEL action without error even if some resources are missing" should do in delegated CNI , try release resource and return null even some error, not in middle of multus , so multus should wait at least one nill from delegated CNI to delete cache file; others when POD yaml also has been deleted , multus can not find enough info about delegated CNI, the delegated CNI has no chance to release resource ..

Multus is not CNI runtime, multus is just CNI plugin. So multus is just aligned with CNI spec. So I keep current simple behavior as default. But it does not means to reject your proposal. Please change to add some option in multus.conf to change the behavior.

s1061123:
Let's imagine that one worker node is shutdown suddenly (without pod migration) and deleted. In such case, IP addresses which is used in destroyed node cannot be freed. In such case, CNI cannot recover that because CNI runtime just cares about 'node', not 'cluster'.

gaopeiliang:
> Let's imagine that one worker node is shutdown suddenly (without pod migration) and deleted. In such case, IP addresses which is used in destroyed node cannot be freed. In such case, CNI cannot recover that because CNI runtime just cares about 'node', not 'cluster'.

yes ...  it maybe an gc case in IPAM or SDN project ....    it is not safe to delete an unhealthy node,   should always repair to ready and graceful evict , delete node ,   it controlled by OP process ... 

most use case  is better reboot or to  start node  when node crash shutdown or kernel bug panic,  too many load or hang ,  some machine health check worker will do it auto ...    so kubelet should clean pod that migration,  try max to release resource when new starting ....

gaopeiliang:
#https://github.com/k8snetworkplumbingwg/multus-cni/issues/891

s1061123:
As @dougbtv mentioned in #891, this issue is not multus issue. This is due to CNI/kubelet design. Hence to collect unused IP could be a part of cluster task, not CNI and kubelet because CNI and kubelet does not care about other node resources.

s1061123:
Your change is in your forked `master` branch but github does not recommend it (commit to master) as 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request

> These changes are proposed in a branch, which ensures that the default branch only contains finished and approved work.

So if you don't mind, please close the PR and create a new PR for that, based on above github document?

gaopeiliang:
https://github.com/k8snetworkplumbingwg/multus-cni/pull/926

PR #923, opened 2022-10-11
runtime config: read CNI args from cni lib

maiqueb:
Read the CNI arguments from the CNI lib API, instead of relying on the environment variable.

This enables the `Delegate` endpoint to retrieve the runtime configuration via CNI args, which was not possible when using environment variables.

Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>


maiqueb:
On second thought, I think this is useless since there's no way for the client to pass the network selection element to the server.

PR #924, opened 2022-10-11
Fix release 3.x thick image tag to isolate from 4.0

s1061123:
Due to #900, image tag, ':thick', which is used in 4.0 and 3.x, gets user confused because thick image is not compatible between 4.0 and 3.x. This fix provides strict version name in deployment yaml to specify install image tag.

Note: This requires to release v3.9.2 release, after the PR merge.


dougbtv (approved):
LGTM!

PR #925, opened 2022-10-12
Support IPs, MAC and cni-args in delegate API

s1061123:


pkg/server/server.go
```diff
@@ -446,7 +446,22 @@ func cmdDelegateAdd(cmdArgs *skel.CmdArgs, k8sArgs *types.K8sArgs, exec invoke.E
...
+		}
+		if annotation.IPRequest != nil {
+			selectionElement.IPRequest = annotation.IPRequest
+		}
+		if annotation.CNIArgs != nil {
+			selectionElement.CNIArgs = annotation.CNIArgs
+		}
+	}
+
+	delegateCNIConf, err := types.LoadDelegateNetConf(cmdArgs.StdinData, selectionElement, "", "")
```
maiqueb, line 464: nit: I would define an help function here, something like:
```golang
delegateCNIConf, err := types.LoadDelegateNetConf(cmdArgs.StdinData, delegateNetworkSelectionElement(annotation), "", "")
...

func delegateNetworkSelectionElement(interfaceAttributes *api.DelegateAnnotatio) *types.NetworkSelectionElement {
    if annotation != nil {
		selectionElement = &types.NetworkSelectionElement{}
		if annotation.MacRequest != "" {
			selectionElement.MacRequest = annotation.MacRequest
		}
		if annotation.IPRequest != nil {
			selectionElement.IPRequest = annotation.IPRequest
		}
		if annotation.CNIArgs != nil {
			selectionElement.CNIArgs = annotation.CNIArgs
		}
                return selectionElement
	}
    return nil
}
```

pkg/server/server.go
```diff
@@ -86,7 +86,7 @@ func (s *Server) HandleCNIRequest(cmd string, k8sArgs *types.K8sArgs, cniCmdArgs
 
 // HandleDelegateRequest is the CNI server handler function; it is invoked whenever
 // a CNI request is processed as delegate CNI request.
-func (s *Server) HandleDelegateRequest(cmd string, k8sArgs *types.K8sArgs, cniCmdArgs *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) ([]byte, error) {
+func (s *Server) HandleDelegateRequest(cmd string, k8sArgs *types.K8sArgs, cniCmdArgs *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo, annotation *api.DelegateAnnotation) ([]byte, error) {
```
maiqueb, line 89: nit: I would name the attribute something other than `annotation`, which is too generic - it can be pretty much anything.

I would prefer something more descriptive - not sure `interfaceAttributes` would be a good fit, because this also has the `cni args`. 

But, since I can't provide an accurate suggestion, I guess yours is good enough.

maiqueb (approved):
@s1061123 thanks for the PR - especially for how fast you pushed it out there.

I've tested this locally with a modified client and it **works as expected** for both IP and MAC addresses.

pkg/server/api/types.go
```diff
@@ -24,6 +24,20 @@ type Request struct {
 	Env map[string]string `json:"env,omitempty"`
 	// CNI configuration passed via stdin to the CNI plugin
 	Config []byte `json:"config,omitempty"`
+	// Annotation for Delegate request
+	Annotation *DelegateAnnotation `json:"delegateAnnotation,omitempty"`
+}
+
+// DelegateAnnotation annotates delegate request for additional config
+type DelegateAnnotation struct {
```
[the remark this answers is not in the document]
s1061123: Changed to `DelegateInterfaceAttributes`.

pkg/server/server.go
s1061123: changed.

pkg/server/server.go
s1061123: Currently I don't come up with to reuse this code yet, so let me keep it. Once we reuse it then let's make helper function.

pkg/server/api/types.go
```diff
@@ -24,6 +24,20 @@ type Request struct {
 	Env map[string]string `json:"env,omitempty"`
 	// CNI configuration passed via stdin to the CNI plugin
 	Config []byte `json:"config,omitempty"`
+	// Annotation for Delegate request
+	InterfaceAttributes *DelegateInterfaceAttributes `json:"delegateAnnotation,omitempty"`
```
maiqueb, line 28: Do we want to align the attribute name w/ the JSON parameter name ?

s1061123: Oh, sorry to miss that. changed.

dougbtv (approved):

PR #926, opened 2022-10-13
delegate plugin delete success, delete cache file

gaopeiliang:
this case should try max to cleanup

sandBox container stop times, but delegated plugin failed (CNI server maybe in starting .... ) , but cache file has been deleted
container gc manager clean up forever even though POD has been fully deleted from ETCD
POD not find && cache file not find; never clean up, we should try max ...

Add an option allowTryDeleteOnErr in multus conf,   allow to try delegate message on such error  scene

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -37,7 +37,8 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
     }, {
         "type": "macvlan",
         ... (snip)
-    }]
+    }],
+    allowTryDeleteOnErr: false
 }
 ```
 
@@ -62,6 +63,7 @@ User should chose following parameters combination (`clusterNetwork`+`defaultNet
 * `systemNamespaces` ([]string, optional): list of namespaces for Kubernetes system (namespaces listed here will not have `defaultNetworks` added)
 * `multusNamespace` (string, optional): namespace for `clusterNetwork`/`defaultNetworks`
 * `delegates` ([]map,required): number of delegate details in the Multus
+* `retryDeleteOnError` (bool, optional): Enable or disable delegate DEL message to next when some missing error. Defaults to false.
 
 ### Network selection flow of clusterNetwork/defaultNetworks
 
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -825,6 +825,7 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 
 	// Read the cache to get delegates json for the pod
 	netconfBytes, path, err := consumeScratchNetConf(args.ContainerID, in.CNIDir)
+	removeCacheConf := false
 	if err != nil {
 		// Fetch delegates again if cache is not exist and pod info can be read
 		if os.IsNotExist(err) && pod != nil {
@@ -854,7 +855,7 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 			return nil
 		}
 	} else {
-		defer os.Remove(path)
+		removeCacheConf = true
 		in.Delegates = []*types.DelegateNetConf{}
 		if err := json.Unmarshal(netconfBytes, &in.Delegates); err != nil {
 			return cmdErr(k8sArgs, "failed to load netconf: %v", err)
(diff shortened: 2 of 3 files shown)


pkg/multus/multus.go
```diff
@@ -896,5 +897,22 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
...
+	// this case POD not find && cache file not find; never clean up
+	if removeCacheConf {
+		if in.AllowTryDeleteOnErr {
+			if e == nil {
+				_ = os.Remove(path)
+			}
+		} else {
+			_ = os.Remove(path) // lgtm[go/path-injection]
+		}
+	}
```
s1061123, line 915: This conditional could be simplified as follows:

```
        if in.AllowTryDeleteOnErr {
                if removeCacheConf && e == nil {
                        _ = os.Remove(path.Clean(cacheConfPath))
                }
        } else {
                _ = os.Remove(path.Clean(cacheConfPath))
        }
```

pkg/types/types.go
```diff
@@ -41,7 +41,10 @@ type NetConf struct {
 	// These parameters are exclusive in one config file:
 	//  - Delegates (directly add delegate CNI config into multus CNI config)
 	//  - ClusterNetwork+DefaultNetworks  (add CNI config through CRD, directory or file)
-	Delegates       []*DelegateNetConf  `json:"-"`
+	Delegates []*DelegateNetConf `json:"-"`
+	// Allow delegate DEL message to next when some missing error
+	AllowTryDeleteOnErr bool `json:"allowTryDeleteOnErr"`
```
s1061123, line 46: Could you please move it to later, after MultusNamespace because we put optional variable there.

s1061123 (changes requested):

docs/configuration.md
```diff
@@ -62,6 +63,7 @@ User should chose following parameters combination (`clusterNetwork`+`defaultNet
 * `systemNamespaces` ([]string, optional): list of namespaces for Kubernetes system (namespaces listed here will not have `defaultNetworks` added)
 * `multusNamespace` (string, optional): namespace for `clusterNetwork`/`defaultNetworks`
 * `delegates` ([]map,required): number of delegate details in the Multus
+* `allowTryDeleteOnErr` (bool, optional): Enable or disable delegate DEL message to next when some missing error. Defaults to false.
```
dougbtv, line 66: we'd like to consider a different name for this functionality, one idea is: `retryDeleteOnError` but maybe we should get some other opinions as well.

pkg/multus/multus.go
```diff
@@ -896,5 +897,22 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
...
 	}
 
-	return delPlugins(exec, pod, args, k8sArgs, in.Delegates, len(in.Delegates)-1, in.RuntimeConfig, in)
+	e := delPlugins(exec, pod, args, k8sArgs, in.Delegates, len(in.Delegates)-1, in.RuntimeConfig, in)
+
+	// delegate plugin delete success, delete cache file
+	// others
+	// 1.  sandBox container stop times, but delegated plugin failed (cache file has been deleted)
+	// 2.  container gc manager clean up forever even though POD has been fully deleted from ETCD
+	// this case POD not find && cache file not find; never clean up
```
dougbtv, line 906: I'll make a suggestion for a comment update as well (todo)

dougbtv (changes requested):

pkg/multus/multus.go
gaopeiliang:
```
        if in.AllowTryDeleteOnErr {
                if removeCacheConf && e == nil {
                        _ = os.Remove(path)
                }
        } else {
                if removeCacheConf {
                        _ = os.Remove(path)   // path file maybe not existed ,  even though has ignore error ...
                }
        }
```

should like this ?  path file maybe not existed that else ;  add check before remove ?  so I start this conditional with  removeCacheConf  like this   if removeCacheConf {.......}

docs/configuration.md
gaopeiliang: good idea !

pkg/types/types.go
gaopeiliang: done

pkg/multus/multus.go
gaopeiliang: update something ....

pkg/multus/multus.go
```diff
@@ -896,5 +897,23 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 		}
 	}
 
-	return delPlugins(exec, pod, args, k8sArgs, in.Delegates, len(in.Delegates)-1, in.RuntimeConfig, in)
+	e := delPlugins(exec, pod, args, k8sArgs, in.Delegates, len(in.Delegates)-1, in.RuntimeConfig, in)
+
+	if removeCacheConf {
+		// Enable Option only delegate plugin delete success to delete cache file
+		// CNI Runtime maybe return an error to block sandbox cleanup a while initiative,
+		// like starting ,prepare something,  it will be OK when retry later
```
s1061123, line 905: "like starting ," -> "like starting, "

pkg/multus/multus.go
s1061123: Thank you for the reply. Yeah, my suggested code was wrong and thank you for the pointing it out.

My intention is to 'check condition of `AllowTryDeleteError` first' to clearly show the behavior with/without the option.
So how about to update such like that?

```
        if in.RetryDeleteOnError {
                if removeCacheConf {
                        if e == nil || strings.Contains(e.Error(), "no such file or directory") {
                                _ = os.Remove(path) // lgtm[go/path-injection]
                        }
                }
        } else {
                if removeCacheConf {
                        _ = os.Remove(path) // lgtm[go/path-injection]
                }
        }
```

gaopeiliang: done

pkg/multus/multus.go
gaopeiliang: ok

s1061123 (approved):
LGTM, merged.

s1061123:
Thank you for the review process! @gaopeiliang

PR #932, opened 2022-10-21
Add "EOF" to the Cat command, otherwise the operation will fail

yulng:
Add EOF to the Cat command, otherwise the operation will fail


s1061123 (approved):
LGTM

Issue #933, opened 2022-10-21
According to the e2e/README.md document, the installation error is reported


yulng:
According to the e2e/README.md document, the installation error is reported
<img width="1033" alt="image" src="https://user-images.githubusercontent.com/31728060/197195036-02a27931-f2e6-40aa-a6aa-ca4f31bbb9b1.png">

Would like to explain what needs to be installed in the environment

yulng:
https://github.com/k8snetworkplumbingwg/multus-cni/pull/934

dougbtv:
Fixed in #934 (thank you)

PR #934, opened 2022-10-21
modify e2e/readme.md

yulng:
modify e2e/readme.md for https://github.com/k8snetworkplumbingwg/multus-cni/issues/933


s1061123 (changes requested):
Thank you for the PR!

By the way, how about to add 'Prerequisite' section and mention required packages in one place, to solve it in one-shot?

yulng:
> Thank you for the PR!
> 
> By the way, how about to add 'Prerequisite' section and mention required packages in one place, to solve it in one-shot?

It's OK to make a description in the README.md document to
prompt the user what to install.
Everyone uses different OS of node, so it is unnecessary to write out how to install the required tools ,such as j2cil ,docker,kubectl

s1061123:
E2e first user just want to know which components are required to run e2e, not 'which script requires which components'.
User just want to invoke 'dnf/apt install ....' in one shot, not for each script runs.

So I propose the change might be following and it might be easy to understand.

```
### Prerequisite

To run the e2e test, you need the following components:

- curl
- j2cli
- docker
- kind
- kubectl

```

Please let me know if you have another intention not to make a such list.

s1061123 (changes requested):
BTW, we don't need kubectl because it is retrieved by gettools.sh

yulng:
> E2e first user just want to know which components are required to run e2e, not 'which script requires which components'. User just want to invoke 'dnf/apt install ....' in one shot, not for each script runs.
> 
> So I propose the change might be following and it might be easy to understand.
> 
> ```
> ### Prerequisite
> 
> To run the e2e test, you need the following components:
> 
> - curl
> - j2cli
> - docker
> ```
> 
> Please let me know if you have another intention not to make a such list.

ok ,modified

s1061123 (approved):
LGTM, merged.

PR #936, opened 2022-10-23
Modify the judgment of test results

yulng:
modify test-simple-macvlan1.sh

### DIFF
--- a/e2e/test-simple-macvlan1.sh
+++ b/e2e/test-simple-macvlan1.sh
@@ -7,7 +7,13 @@ kubectl create -f yamls/simple-macvlan1.yml
 kubectl wait --for=condition=ready -l app=macvlan --timeout=300s pod
 
 echo "check macvlan1-worker1 interface: net1"
-kubectl exec macvlan1-worker1 -- ip a show dev net1
+net=$(kubectl exec macvlan1-worker1 -- ip a show dev net1)
+if [ $? -eq 0 ];then
+	echo "macvlan1-worker1 pod has net1 card"
+else
+	echo "macvlan1-worker1 pod has no net1 card"
+	exit 1
+fi
 
 echo "check macvlan1-worker1 interface address: net1"
 ipaddr=$(kubectl exec macvlan1-worker1 -- ip -j a show  | jq -r \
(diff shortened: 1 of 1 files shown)


s1061123:
Hi @yulng , thank you for your PR, again! Could you please explain why you need the changes? I understand what the code does, but I'd like to know why you need to add it.

yulng:
> Hi @yulng , thank you for your PR, again! Could you please explain why you need the changes? I understand what the code does, but I'd like to know why you need to add it.

Confirm whether the network card name of the pod is net1, and return the result through "echo"

yulng:
> > Hi @yulng , thank you for your PR, again! Could you please explain why you need the changes? I understand what the code does, but I'd like to know why you need to add it.
> 
> Confirm whether the network card name of the pod is net1, and return the result through "echo"

@s1061123

s1061123 (commented):
I got it. If you don't mind it, please add 'exit 1' in case of 'no net1' interface, because 'no net1 interface' case, we just send error.

yulng:
> I got it. If you don't mind it, please add 'exit 1' in case of 'no net1' interface, because 'no net1 interface' case, we just send error.

ok,if no net1, return the execution result of the command
modified

yulng:
> > I got it. If you don't mind it, please add 'exit 1' in case of 'no net1' interface, because 'no net1 interface' case, we just send error.
> 
> ok,if no net1, return the execution result of the command modified

@s1061123

s1061123:
@yulng sorry, but I mean that 'if no net1 case, just `exit 1`'. Because if there is no net1, then it must be error hence CI should be stopped with 'explicitly error'. In github action, non 'exit 0' cases are recognized as error.

So please add `exit 1` in case of 'no net1' case, for two places.

yulng:
> @yulng sorry, but I mean that 'if no net1 case, just `exit 1`'. Because if there is no net1, then it must be error hence CI should be stopped with 'explicitly error'. In github action, non 'exit 0' cases are recognized as error.
> 
> So please add `exit 1` in case of 'no net1' case, for two places.

Okay, it's added











































































ok, added

Issue #931, opened 2022-10-20
Multus fails to copy user provided configuration


reversecipher:
**What happend**:

Multus daemon fails to copy user provided configuration:

`[error] failed to copy the user provided configuration /tmp/multus-conf/00-multus.conf: error copying file - copied only 0 bytes out of 1703`

Multus Daemonset (relevant parts):

```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: kube-multus-ds
  namespace: kube-system
  labels:
    tier: node
    app: multus
    name: multus
spec:
  template:
    spec:
      containers:
        - name: kube-multus
          image: ghcr.io/k8snetworkplumbingwg/multus-cni:snapshot-thick
          command: [ "/usr/src/multus-cni/bin/multus-daemon" ]
          args:
            - "-cni-version=0.3.1"
            - "-cni-config-dir=/host/etc/cni/net.d"
            - "-multus-autoconfig-dir=/host/etc/cni/net.d"
            - "-multus-log-to-stderr=true"
            - "-multus-log-level=verbose"
            - "-multus-conf-file=/tmp/multus-conf/00-multus.conf"
          volumeMounts:
            - name: custom-config
              mountPath: /tmp/multus-conf
      volumes:
        - name: custom-config
          configMap:
            name: multus-custom-config
            items:
            - key: 00-multus.conf
              path: 00-multus.conf
```

ConfigMap:

```
Name:         multus-custom-config
Namespace:    kube-system
Labels:       <none>
Annotations:  <none>

Data
====
00-multus.conf:
----
{
    "capabilities": {
        "bandwidth": true,
        "portMappings": true
    },
    "cniVersion": "0.3.1",
    "logLevel": "verbose",
    "logToStderr": true,
    "name": "multus-cni-network",
    "type": "multus-shim",
    "delegates": [
        {
            "name": "k8s-pod-network",
            "cniVersion": "0.3.1",
            "plugins": [
                {
                    "type": "calico",
                    "datastore_type": "kubernetes",
                    "mtu": 0,
                    "nodename_file_optional": false,
                    "log_level": "Info",
                    "log_file_path": "/var/log/calico/cni/cni.log",
                    "ipam": {
                        "type": "calico-ipam",
                        "assign_ipv4": "true",
                        "assign_ipv6": "false"
                    },
                    "container_settings": {
                        "allow_ip_forwarding": false
                    },
                    "policy": {
                        "type": "k8s"
                    },
[...]

PR #939, opened 2022-10-24
Fix multus-daemon in case of user given config

s1061123:
This fix the issue #931 to refining server startup code including file copy and channel processing, with simplification.


dougbtv (approved):
LGTM!

PR #941, opened 2022-10-26
fix kubeletSocket scope

moshe010:
The kubeletSocket in the function is scoping the const kubeletSocket which causing resource api to fail with

"stat /var/lib/kubelet/pod-resources/.sock: no such file or directory"

Signed-off-by: Moshe Levi <moshele@nvidia.com>


s1061123 (approved):
LGTM, thank you for the fix!

Issue #944, opened 2022-10-31
[BUG] cannot create pod requiring resources in the thick plugin arch


maiqueb:
**What happend**:
Fails to create a pod requiring resources; it hangs in `ContainerCreating` state.

**What you expected to happen**:
Pod should be created. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Have [macvtap-cni](https://github.com/kubevirt/macvtap-cni/) installed in the cluster.
2. Provision the following NAD
```yaml
apiVersion: v1
items:
- apiVersion: k8s.cni.cncf.io/v1
  kind: NetworkAttachmentDefinition
  metadata:
    annotations:
      k8s.v1.cni.cncf.io/resourceName: macvtap.network.kubevirt.io/eth0
    creationTimestamp: "2022-10-31T09:03:27Z"
    generation: 1
    name: macvtap0
    namespace: cluster-network-addons
    resourceVersion: "2443"
    uid: af896e29-101b-48d5-8e89-8f3391c1e1fa
  spec:
    config: '{ "cniVersion": "0.3.1", "name": "macvtap0", "type": "macvtap"}'
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
3. Provision the following pod; it'll hang in `ContainerCreating`
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    k8s.v1.cni.cncf.io/networks: macvtap0
  name: megapod
  namespace: cluster-network-addons
spec:
  containers:
  - command:
    - /bin/sh
    - -c
    - sleep 999999
    image: alpine
    imagePullPolicy: Always
    name: tinycontainer
    resources:
      limits:
        macvtap.network.kubevirt.io/eth0: "1"
      requests:
        macvtap.network.kubevirt.io/eth0: "1"
```

**Anything else we need to know?**:
The following event can be seen:
```
[...]

PR #946, opened 2022-10-31
fix resource api  grpc failed to connect to unix socket

moshe010:
After commit https://github.com/k8snetworkplumbingwg/multus-cni/commit/c6fa047212efc3a410fb73aa83d2594a4f914320 resource api got broken in 2 places. first place handled by https://github.com/k8snetworkplumbingwg/multus-cni/commit/f530d3eb84608fa25a59737bcf49991d4a037d82 The second break was in replacing podresources.GetV1Client with getKubeletResourceClient. The GetV1Client remove the protocol (e.g unix:) by calling GetAddressAndDialer, but the getKubeletResourceClient is using the full endpoint (with the portocol) causing failed to connect to unix socket. This patch remove the unix: prefix before the grpc connect.

Fixes: #944
Signed-off-by: Moshe Levi <moshele@nvidia.com>

### DIFF
--- a/pkg/kubeletclient/kubeletclient.go
+++ b/pkg/kubeletclient/kubeletclient.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	defaultkubeletSocket              = "kubelet" // which is defined in k8s.io/kubernetes/pkg/kubelet/apis/podresources
+	defaultKubeletSocket       = "kubelet" // which is defined in k8s.io/kubernetes/pkg/kubelet/apis/podresources
 	kubeletConnectionTimeout   = 10 * time.Second
 	defaultKubeletSocketFile   = "kubelet.sock"
 	defaultPodResourcesMaxSize = 1024 * 1024 * 16 // 16 Mb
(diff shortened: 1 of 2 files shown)


moshe010:
@maiqueb  this should fix you issue. (tested on my local env)

pkg/kubeletclient/kubeletclient.go
```diff
@@ -38,10 +39,14 @@ func LocalEndpoint(path, file string) (string, error) {
 	return filepath.Join(u.String(), file+".sock"), nil
 }
 
+func removeUnixProtocol(endpoint string) (string) {
+	return strings.TrimPrefix(endpoint, (unixProtocol + ":"))
+}
+
```
s1061123, line 45: How about to use `u, err := url.Parse(endpoint)` as `hasKubeletAPIEndpoint()` does?

```
func removeUnixProtocol(endpoint string) (string, error) {
    u, err := url.Parse(endpoint)
    if err != nil {
        return "", err
    }
   return u.Path, nil
}
```

s1061123:
@maiqueb Hi, did you try the fix with your environment?

maiqueb:
> @maiqueb Hi, did you try the fix with your environment?

Yes - it works as expected. Thank you @s1061123 / @moshe010 .

s1061123 (approved):
LGTM, merged.

PR #952, opened 2022-11-06
clean code for golang-lint

yanggangtony:
Hi , i test the code in `golangci-lint run`.
there are several errors to found .
Should we according the result to fix it?


pkg/server/api/api.go
```diff
@@ -65,23 +64,3 @@ func DoCNI(url string, req interface{}, socketPath string) ([]byte, error) {
...
-			"CNI_COMMAND":     strings.ToUpper(cniCommand),
-			"CNI_CONTAINERID": cniContainerID,
-			"CNI_NETNS":       cniNetNS,
-			"CNI_IFNAME":      cniIFName,
-			"CNI_ARGS":        fmt.Sprintf("K8S_POD_NAMESPACE=%s;K8S_POD_NAME=%s;K8S_POD_UID=%s", podNamespace, podName, podUID),
-		},
-		Config:              cniConfig,
-		InterfaceAttributes: interfaceAttributes,
-	}
-}
```
s1061123, line 87: I don't run golang-lint, however, I guess that it is shown as 'unused code' in lint, however these function is API and consumed by other repository, so please keep these code (including 'string' import).

s1061123 (changes requested):

yanggangtony:
@s1061123 
thanks for review the pr.
According to your opinion, I had updated it ..

s1061123:
Merged. Thank you for the PR!

PR #954, opened 2022-11-07
Upgrade fsnotify

yanggangtony:
Maybe we should upgrade the fsnotify version: 
https://github.com/fsnotify/fsnotify/releases

This makes checking events a lot easier , and other code advanced..


s1061123 (approved):
LGTM, merged.

PR #955, opened 2022-11-07
for miss delete system files in daemon-config

yanggangtony:

The daemon-config.json is a location where user can specify , when he modify the contents. and use it do a evil thing , we should prevent it.. like set  "socketDir": "xxxxx" ..
daemon-config.json: |
    {
        "confDir": "/host/etc/cni/net.d",
        "logToStderr": true,
        "logLevel": "debug",
        "logFile": "/tmp/multus.log",
        "binDir": "/host/opt/cni/bin",
        "cniDir": "/var/lib/cni/multus",
        "socketDir": "xxxxx"
    }

According the code 

`
multus-cni/cmd/multus-daemon/main.go
if err := srv.FilesystemPreRequirements(daemonConfig.MultusSocketDir); err != nil {

multus-cni/pkg/server/server.go
func FilesystemPreRequirements(rundir string) error {
	if err := os.RemoveAll(rundir); err != nil && !os.IsNotExist(err) {
`

### DIFF
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -458,7 +458,13 @@ func LoadDaemonNetConf(configPath string) (*ControllerNetConf, []byte, error) {
 		daemonNetConf.BinDir = defaultBinDir
 	}
 
-	if daemonNetConf.MultusSocketDir == "" {
+	if daemonNetConf.MultusSocketDir == "" ||
+		daemonNetConf.MultusSocketDir == "/" ||
+		daemonNetConf.MultusSocketDir == "/run" ||
+		daemonNetConf.MultusSocketDir == "/host" ||
+		daemonNetConf.MultusSocketDir == "/tmp" ||
+		daemonNetConf.MultusSocketDir == "/var/lib" ||
+		daemonNetConf.MultusSocketDir == "/etc" {
 		daemonNetConf.MultusSocketDir = defaultMultusRunDir
 	}
 


s1061123:
Thank you for the PR, @yanggangtony , however I suppose it is not valid use cases.
Multus CNI is of course for Kubernetes cluster network configuration, hence multus-daemonset.yaml assumes to be used by someone Kubernetes administrator, not for Kubernetes pod users. If someone (pod users) try to do evil thing, it should be prevent in [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) or some security feature, not argument parsing.

I believe that these multus configs are like linux administrator operation, such as modify '/etc' or some. So user should understand what is happen. In addition, we cannot list up 'the path we prevent', actually (your list is not sufficient, I suppose. how about '/host/boot', for example?)

yanggangtony:
> Thank you for the PR, @yanggangtony , however I suppose it is not valid use cases. Multus CNI is of course for Kubernetes cluster network configuration, hence multus-daemonset.yaml assumes to be used by someone Kubernetes administrator, not for Kubernetes pod users. If someone (pod users) try to do evil thing, it should be prevent in [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) or some security feature, not argument parsing.
> 
> I believe that these multus configs are like linux administrator operation, such as modify '/etc' or some. So user should understand what is happen. In addition, we cannot list up 'the path we prevent', actually (your list is not sufficient, I suppose. how about '/host/boot', for example?)

@s1061123 
you are right , we just close this pr .:stuck_out_tongue_winking_eye:

Issue #949, opened 2022-11-03
support pod bandwidth limit when clusterNetwork is calico


cyclinder:
**What would you like to be added**:

At present, the `kubernetes.io/ingress-bandwidth: 1M` annotation isn't recognized by Multus. So we can't use bandwidth limit when clusterNetwork is calico.

**Why is this needed**:

support pod bandwidth limit when clusterNetwork is calico

cyclinder:
I think there are two options:

- support  `kubernetes.io/ingress-bandwidth` annotation can be recognized by Multus.
- extende `v1.multus-cni.io/default-network` annotations, we can fill in the configuration parameters for the bandwidth, like this:

```
     v1.multus-cni.io/default-network: 
        {
          "name":"calico",
          "namespace": "kube-system",
          "bandwidth: {
            "ingressRate": 2048, // Limit ingress to 2Kbps
            "ingressBurst": 300, // Allow small egress bursts of 300b
            "egressRate": 8000,  // Limit egress to 8000bps
            "egressBurst": 200   // Allow small egress bursts of 200b
          }
        }
     

```

cyclinder:
I think option1 is better, if you guys think this make a sense, I can work on this for it.

s1061123:
Thank you for the issue, @cyclinder , however, I suppose it is not an issue to be solved because CNI/K8s already provide another way for bandwidth annotation.

Usual Kubernetes CNI, `kubernetes.io/ingress-bandwidth` annotation is not recognized, actually. As far as I know, these bandwidth information will be passed to CNI as a part of capability/runtimeConfig and it is consumed by each CNI plugin as described in [CNI CONVENTION document](https://github.com/containernetworking/cni/blob/main/CONVENTIONS.md).

Usually `bandwidth` plugin will take care of bandwidth annotation and do some configuration.

I'm not sure how calico works for bandwidth, however, calico community also mentions to use of 'bandwidth' CNI plugin with calico as https://github.com/projectcalico/calico/issues/5007

cyclinder:
@s1061123 Thanks for your reply! Maybe I didn't express myself clearly. 

I made the following tests for bandwidth(inject annotations(kubernetes.io/ingress-bandwidth) to pod): 

| situation|only calico | calico with multus |
|---|---|---|
|result |works | no works |

[...]

PR #958, opened 2022-11-11
enable bandwidth by default

cyclinder:
enable bandwidth by default

It's would be better if it enables bandwidth in multus-cni config by default, How do you think?

Refer to: https://github.com/k8snetworkplumbingwg/multus-cni/issues/949

### DIFF
--- a/deployments/multus-daemonset-crio.yml
+++ b/deployments/multus-daemonset-crio.yml
@@ -121,7 +121,8 @@ data:
       "name": "multus-cni-network",
       "type": "multus",
       "capabilities": {
-        "portMappings": true
+        "portMappings": true,
+        "bandwidth": true
       },
       "delegates": [
         {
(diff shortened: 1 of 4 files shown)


s1061123:
As default daemonset yaml case, we dynamically load required capability from given CNI config, hence I think we don't need that.

See:
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/server/config/generator.go#L232-L262 (for thick image)
- https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/entrypoint.sh#L382-L396 (for thin image)

s1061123:
So if you don't mind, I will close this PR.

s1061123:
In addition, as we told in comment for each yaml file, these deployment yaml does not cover all use-cases. So we expect that user customize as you want to real deployment scenario based on user's use-case. Just FYI.

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/multus-daemonset.yml#L1-L8

cyclinder:
Well, I close this.

PR #959, opened 2022-11-11
Remove the deprecated ioutil function

cyclinder:
Remove the deprecated ioutil function


s1061123:
package `ioutil ` is already removed by https://github.com/k8snetworkplumbingwg/multus-cni/pull/951 hence I suppose it is not valid PR. Please double check that.

cyclinder:
Oh sorry, I'm don't notice it's has been fixed, I will close it.

PR #960, opened 2022-11-13
add check test result for interface net1

yulng:
add check test result for interface net1

@s1061123

### DIFF
--- a/e2e/test-default-route1.sh
+++ b/e2e/test-default-route1.sh
@@ -8,6 +8,9 @@ kubectl wait --for=condition=ready -l app=default-route1 --timeout=300s pod
 
 echo "check default-route-worker1 interface: net1"
 kubectl exec default-route-worker1 -- ip a show dev net1
+if [ $? -ne 0 ];then
+       exit 1
+fi
 
 echo "check default-route-worker1 interface address: net1"
 ipaddr=$(kubectl exec default-route-worker1 -- ip -j a show  | jq -r \
@@ -25,6 +28,9 @@ fi
 
 echo "check default-route-worker2 interface: net1"
 kubectl exec default-route-worker2 -- ip a show dev net1
+if [ $? -ne 0 ];then
+       exit 1
+fi
 
 echo "check default-route-worker2 interface address: net1"
 ipaddr=$(kubectl exec default-route-worker2 -- ip -j a show  | jq -r \


e2e/test-default-route1.sh
```diff
@@ -7,7 +7,13 @@ kubectl create -f yamls/default-route1.yml
 kubectl wait --for=condition=ready -l app=default-route1 --timeout=300s pod
 
 echo "check default-route-worker1 interface: net1"
-kubectl exec default-route-worker1 -- ip a show dev net1
+net=$(kubectl exec default-route-worker1 -- ip a show dev net1)
```
s1061123, line 10: the variable `net` seems not to used anywhere. could you please let me know why `net` is required?

e2e/test-default-route1.sh
```diff
@@ -24,7 +30,13 @@ if [ $ipaddr != "10.1.1.254" ]; then
 fi
 
 echo "check default-route-worker2 interface: net1"
-kubectl exec default-route-worker2 -- ip a show dev net1
+net2=$(kubectl exec default-route-worker2 -- ip a show dev net1)
```
s1061123, line 33: the variable `net2` seems not to used anywhere. could you please let me know why `net` is required?

s1061123 (changes requested):

e2e/test-default-route1.sh
yulng: ok, echo $net

e2e/test-default-route1.sh
yulng: ok, echo $net2

yulng: @s1061123  😅

e2e/test-default-route1.sh
s1061123: We need kubectl output even though error code is non 0, for troubleshooting. How about to keep `kubectl exec default-route-worker1 -- ip a show dev net1`?

e2e/test-default-route1.sh
s1061123: We need kubectl output even though error code is non 0, for troubleshooting. How about to keep `kubectl exec default-route-worker1 -- ip a show dev net1`?

s1061123 (changes requested):

e2e/test-default-route1.sh
yulng: > We need kubectl output even though error code is non 0, for troubleshooting. How about to keep `kubectl exec default-route-worker1 -- ip a show dev net1`?
ok,updated it , kubectl output will be printed
image

s1061123: This code still seems to be redundant. how about that?

```
kubectl exec default-route-worker1 -- ip a show dev net1
if [ $? -ne 0 ];then
       exit 1
fi
```

e2e/test-default-route1.sh
s1061123: This code still seems to be redundant. how about that?

```
kubectl exec default-route-worker2 -- ip a show dev net1
if [ $? -ne 0 ];then
       exit 1
fi
```

s1061123 (changes requested):

e2e/test-default-route1.sh
yulng: ok,updated

s1061123 (approved):
LGTM. merged.

PR #961, opened 2022-11-14
Optimize the format of the return valueReturnlogging

yulng:
Optimize the format of the return value
thanks
@s1061123

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -361,8 +361,7 @@ func DelegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, dele
 	ips := []string{}
 	res, err := cni100.NewResultFromResult(result)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


s1061123:
Will take a look into it next week.

s1061123:
Thank you for the PR, @yulng 

These changes seem NOT just optimization, because it will change the behavior because both return contains error even though current code does not return without error.
In both return condition case, multus-cni successfully created interface so we should not return error (as written in comment).

Could you please let me know the detail reason why you change it and why the change is valid (including backward compatibility)?

yulng:
> Thank you for the PR, @yulng
> 
> These changes seem NOT just optimization, because it will change the behavior because both return contains error even though current code does not return without error. In both return condition case, multus-cni successfully created interface so we should not return error (as written in comment).
> 
> Could you please let me know the detail reason why you change it and why the change is valid (including backward compatibility)?

ok, i closed it

PR #988, opened 2022-12-13
Do 'go mod tidy' to cleanup go.mod/go.sum

s1061123:


s1061123:
Once CI is passed, I will merge it because it is trivial and nothing is changed from code. thanks.

PR #992, opened 2022-12-13
Create an openshift origin (OKD) golang builder image in build pipeline

dougbtv:
This has a build against openshift origin, since the original image was out of date, this builds a golang builder image prior to the origin Multus build using the dockerfile provided by OKD.


s1061123 (approved):
LGTM, merged!

PR #1006, opened 2022-12-20
Removes command line arguments in favor of configMap/JSON config

dougbtv:
Work in progress for making so that users only have to specify configurations in a single place -- the configMap.

Keeps only the bare minimum of command line options (in particular: where the config file lives, and the version flag)

Currently just attempts to remove command line arguments and see what needs to change from there (before further changes)


dougbtv:
This should be essentially ready to review and ready to roll

PR #1012, opened 2023-01-05
fix:extensions/v1beta1 is changed to apps/v1

yulng:
Signed-off-by: yulng <wei.yang@daocloud.io>

The apiversion extensions/v1beta1 for DaemonSet has been changed to apps/v1 from k8s v1.9 version

Thanks


s1061123:
@yulng thank you for the PR, however, this 1.16 and 1.19 are pretty old and we may deprecate (i.e. remove) these files because we don't support old Kubernetes version. Let me remove these files in repository. Thanks for the report that!

yulng:
ok,i closed it

PR #1027, opened 2023-02-05
fix multi-architecture for thick image

f2calv:
updates to Dockerfile.thick to allow for elegant multi-architecture builds in a single Dockerfile.

```bash
#set target platforms
PLATFORM="linux/amd64,linux/arm64,linux/arm/v7"

#Create a new builder instance
docker buildx create --name multus-cni --use

#Start the build
docker buildx build \
    -f images/Dockerfile.thick \
    -t ghcr.io/k8snetworkplumbingwg/multus-cni:snapshot-thick \
    --platform $PLATFORM \
    --pull \
    .
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+* text=auto eol=lf
+*.{cmd,[cC][mM][dD]} text eol=crlf
+*.{bat,[bB][aA][tT]} text eol=crlf
--- a/.github/workflows/image-build.yml
+++ b/.github/workflows/image-build.yml
@@ -19,8 +19,8 @@ jobs:
           tags: ghcr.io/${{ github.repository }}:latest-amd64
           file: images/Dockerfile
 
-  build-amd64-thick:
-    name: Image build/amd64 thick plugin
+  build-xplatform-thick:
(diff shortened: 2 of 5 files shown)


s1061123:
Thank you for the PR. I have question and confirmation about that.


## Are you going to help support for the multi-architecture?

I suppose you did test multus in arm64 and arm/v7 platform as you submit this multi-architecture code. Currently multus team does not have any arm platform (arm64 and arm32), so we are hesitate to make multi-arch images for that because we cannot test/reproduce issue once we get issue, that is only happen in non amd64. So we really need someone who test/reproduce issue for multi-arch related issue. Could you please help us to support multi-architecture?

## Could you support other container runtime build, for example 'podman build'?

Your PR utilizes several features only in 'docker buildx' command and that gives simplified Dockerfile for multi-arch. Currently our Dockerfile is designed for common use, not only for CI build.
So someone may use Dockerfile for private build, with podman, for example. Could you improve this Dockerfile to work with other container runtime platform?

nvollmar:
@s1061123 At least v4.0.0-alpha was published for arm64 and so far the thick plugin has been running fine on my k3s arm cluster.

PR #1028, opened 2023-02-06
Remove deprecated 'restart-crio' option

s1061123:
'restart-crio' option was used for workaround crio issues. This issue no longer exists, hence make it obsolate and removed.


dougbtv (approved):
LGTM

PR #1030, opened 2023-02-06
Clean up unused structures

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

Clean up unused structures and adjust the order of imported packages

### DIFF
--- a/pkg/server/config/generator_test.go
+++ b/pkg/server/config/generator_test.go
@@ -18,7 +18,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"testing"
 
 	testutils "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/testing"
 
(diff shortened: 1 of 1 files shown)


pkg/types/types.go
```diff
@@ -16,11 +16,11 @@
 package types
 
 import (
-	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/logging"
```
s1061123, line 19: Could you please let me know why you change the import list? If you don't have any clear reason, we'd prefer to keep as it is.

s1061123 (changes requested):

pkg/types/types.go
Fish-pro: @s1061123 The package import order is confusing. For example, `golangci-lint` has an explicit requirement for package import order. Also, the package import order here is different from the other code files. Typically, golang's built-in packages are imported first, and then third-party packages are imported

s1061123: Currently, as far as I know of, we don't have strict import rule in this repository yet. So this change should be done once we have the one. Please revert this change for now.

Fish-pro: Order has been preserved

PR #1032, opened 2023-02-09
delete GO111MODULE=off case in test-go.sh

yanggangtony:

Go 1.17 was released on August 16, 2021. As for 1.16, GO111MODULE=on is the default behavior, and GO111MODULE=auto is equivalent to GO111MODULE=on

### DIFF
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -1,23 +1,4 @@
 #!/usr/bin/env bash
 set -e
 
-# this if... will be removed when gomodules goes default
-if [ "$GO111MODULE" == "off" ]; then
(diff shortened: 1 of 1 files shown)


s1061123:
Thank you for the PR. As for 1.16, GO111MODULE=on is the default, however, some user still want to disable GO111MODULE because some user want to build/test offline (i.e. disconnect from network). For example, package build system (i.e. Fedora) does package build under offline environment. Your change just remove GO111MODULE=off case, hence offline user cannot test with your change.

Could you please change your PR to support offline for the offline users?

yanggangtony:
@s1061123
Thanks for review.
I will handle it later
Current is busy with works..
😬😬

PR #1033, opened 2023-02-09
align case line .

yanggangtony:

Keep the case code align..

### DIFF
--- a/images/entrypoint.sh
+++ b/images/entrypoint.sh
@@ -90,8 +90,8 @@ function warn()
 }
 
 function checkCniVersion {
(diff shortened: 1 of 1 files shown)


s1061123:
Thank you for the PR, @yanggangtony . This indent issues are happen by 'tab'/'space' and there are several similar case in same file, 'image/entrypoint.sh'. If you don't mind it, could you please take care about it? Thank you!

yanggangtony:
@s1061123 
Thanks for review and the suggestion..
I make 'tab'/'space' visible good in my dev-idea vscode .
And update the not right places.

PR #1034, opened 2023-02-09
avoid panic when array length =0.

yanggangtony:

Make code more robust for panic array.


s1061123:
Thank you for the PR. The fix seems to be okey, but could you please add unit-test cases to protect it from further regression issues?

yanggangtony:
@s1061123 
Thanks for review.
I will handle it later
Current is busy with works..
:grimacing::grimacing:

PR #1035, opened 2023-02-10
Cleanup : github.com/pkg/errors has been archived.

yanggangtony:
[https://github.com/pkg/errors](https://github.com/pkg/errors)  says : This repository has been archived by the owner on Dec 1, 2021. It is now read-only.

I suggest we should delete that , replaced by the go's error package.


s1061123:
Looks good to me, merged. Thanks!

PR #1037, opened 2023-02-13
code clean for if condtion

yanggangtony:

if bool variable is true , then we can omit it in ``` if xxx```


s1061123:
It looks good to me.

PR #1044, opened 2023-03-02
Remove old/obsolated config from repo

s1061123:
As of #1012 discussion, several conf file seems to be obsolated and not updated to track the changes. This change removes these obsolated files.


dougbtv (approved):
lgtm

PR #1045, opened 2023-03-04
Update getNetDelegate to first check if NAD exist before getting it.

mskrocki:
The getKubernetesDelegate is used in getNetDelegate which has failover mechanic to first look for NAD and later search CNI json file. If NAD was not present, we would report error event even if the CNI json is present and correct.

### DIFF
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
@@ -536,9 +537,18 @@ func getNetDelegate(client *ClientInfo, pod *v1.Pod, netname, confdir, namespace
 			Name:      netname,
 			Namespace: namespace,
 		}
-		delegate, resourceMap, err := getKubernetesDelegate(client, net, confdir, pod, resourceMap)
+
+		// check NAD exists first
+		_, err := client.NetClient.NetworkAttachmentDefinitions(net.Namespace).Get(context.TODO(), net.Name, metav1.GetOptions{})
 		if err == nil {
-			return delegate, resourceMap, nil
+			delegate, resourceMap, err := getKubernetesDelegate(client, net, confdir, pod, resourceMap)
+			if err == nil {
+				return delegate, resourceMap, nil
+			}
+		} else if apierrors.IsNotFound(err) {
+			logging.Verbosef("network-attachment-definition (%s) in namespace (%s) does not exist: %v", netname, namespace, err)
+		} else {
+			logging.Errorf("error getting network-attachment-definition (%s) in namespace (%s): %v", netname, namespace, err)
 		}
 
 		// option2) search CNI json config file, which has <netname> as CNI name, from confDir


pkg/k8sclient/k8sclient.go
```diff
@@ -251,12 +251,17 @@ func parsePodNetworkAnnotation(podNetworks, defaultNamespace string) ([]*types.N
...
-	logging.Debugf("getKubernetesDelegate: %v, %v, %s, %v, %v", client, net, confdir, pod, resourceMap)
+	logging.Debugf("getKubernetesDelegate: %v, %v, %s, %v, %v, %v", client, net, confdir, pod, resourceMap, try)
 	customResource, err := client.NetClient.NetworkAttachmentDefinitions(net.Namespace).Get(context.TODO(), net.Name, metav1.GetOptions{})
 	if err != nil {
 		errMsg := fmt.Sprintf("cannot find a network-attachment-definition (%s) in namespace (%s): %v", net.Name, net.Namespace, err)
+		if try {
+			// just an attempt to find NAD in getNetDelegate
+			// do not generate event
+			return nil, resourceMap, fmt.Errorf("getKubernetesDelegate: " + errMsg)
+		}
```
s1061123, line 264: I suppose we should not introduce new arugment `try` in this case because this argument may change the behavior of the function.

pkg/k8sclient/k8sclient.go
```diff
@@ -251,12 +251,17 @@ func parsePodNetworkAnnotation(podNetworks, defaultNamespace string) ([]*types.N
 	return networks, nil
 }
 
-func getKubernetesDelegate(client *ClientInfo, net *types.NetworkSelectionElement, confdir string, pod *v1.Pod, resourceMap map[string]*types.ResourceInfo) (*types.DelegateNetConf, map[string]*types.ResourceInfo, error) {
+func getKubernetesDelegate(client *ClientInfo, net *types.NetworkSelectionElement, confdir string, pod *v1.Pod, resourceMap map[string]*types.ResourceInfo, try bool) (*types.DelegateNetConf, map[string]*types.ResourceInfo, error) {
 
-	logging.Debugf("getKubernetesDelegate: %v, %v, %s, %v, %v", client, net, confdir, pod, resourceMap)
+	logging.Debugf("getKubernetesDelegate: %v, %v, %s, %v, %v, %v", client, net, confdir, pod, resourceMap, try)
 	customResource, err := client.NetClient.NetworkAttachmentDefinitions(net.Namespace).Get(context.TODO(), net.Name, metav1.GetOptions{})
 	if err != nil {
```
s1061123, line 258: This handles not only NotFound also serious error case, but in this case, we can change the behavior for NotFound error case and other error case as follows:

- suppress error to output kube events, print log with `logging.Verbose` for internal use and go further in NotFound case
- print error message to kube events (as current code does) in other error case

What do you think about that?

s1061123 (changes requested):

pkg/k8sclient/k8sclient.go
mskrocki: Good catch, I missed the other errors situation, but I think for the other cases we should report the event when nad not found,won't that be a bigger change in behavior?

s1061123: As I noted, other cases (not NotFound case), we should report the error in event. Let me share my idea which is in above comment:

- If NotFound error, we don't report the error as Kubernetes event, however, we should report it in multus logging, with  `logging.Verbosef()` for further troubleshooting (we need to report this case we don't use net-attach-def)
- If other error, we should report in Kubernetes event as currently we did

s1061123: Almost OK to me. Just a double check that this new proposal will remove `try` argument from function, is that correct?

mskrocki: Yes this would not have the 'try', but what about the behavior change in the other functions?

s1061123: Please let me know what is the behavior change without try. As far as I know of, this 'try' is tntroduced in your PR.

s1061123: We support CNI json file in pod network annotation, as well as defaultNetwork/clusterNetwork. Your PR seems to only support CNI file in defaultNetwork/clusterNetwork with `try`, so your PR changes the behavior.

We want to support both, so please remove `try` and keep current behavior.

s1061123 (changes requested):

pkg/k8sclient/k8sclient.go
mskrocki: But in 511 and 673 we return right away when `getKubernetesDelegate` fails, in contrast to line 544 where we do not fail on err != nil and do check the CNI config as a failback. My `try` ensures we do not report error when this failback happens.

s1061123: I read your code again and I make sure that we need to report the error in both case, 511, 673. 

So please let me know the reason why you do not report the error?

s1061123 (changes requested):

pkg/k8sclient/k8sclient.go
s1061123: I read your code again and I make sure that we need to report the error in both case, 511, 673. 

The difference between 511 and 673 is retrieve CNI config for network annotation in pod, or 'clusterNetwork'/'defaultNetwork' config in multus config. Both case, we should report the error to users otherwise user may not recognize error. However, your PR intentionally skip to report the error in 'clusterNetwork'/'defaultNetworks' case, by adding `try` clause.
For example, if one of 'defaultNetworks' is not invalid (i.e. net-attach-def is not exist), your code does not show any error for that. Current multus code, of course, shows the error and let the user to know users' config has something wrong. Both case, we should report the error to Kubernetes events to easily understand what is happen. So I still believe we do not need your 'try' clause.

So please let me know the reason why you do not report the error in ?

s1061123: In addition, I am not clear why `try` does skipping its error report. Could you please explain why it names `try`?

mskrocki: Because in 544 (where I set `try=true` and change the behavior) we failback to the CNI json check, and only if that fails report the error.

I probably could add event report at the end of that function (line 607) to ensure we eventually report the event on failure.

s1061123: > Because in 544 (where I set try=true and change the behavior) we failback to the CNI json check, and only if that fails report the error.

As Doug noted, net-attach-def is required even if CNI config name is in Pod annotation. In this case, net-attach-def does not have any config in spec, like following (that's why config is optional):

```
---
metadata:
  name: etc-cni-config1
spec:
```

mskrocki: To put more light to my use case, my `00-multus.conf` is:
```
{
  "name": "multus-cni-network",
  "type": "multus",
  "confDir": "/etc/cni/net.d",
...
  "clusterNetwork": "cilium"
}
```

where I am using the `clusterNetwork`, but I do not create a NAD for it, and just rely on the existing cni conf for `cilium` inside the `/etc/cni/net.d` path.
In such setup, whenever I create default Pod, (w/o any additional networks) I will get an error event that cilium NAD not found, where in reality all works.

What if we keep the `try`, and I expand the `if` with `apierrors.IsNotFound(err)` and add logging, like this:
```
		errMsg := fmt.Sprintf("cannot find a network-attachment-definition (%s) in namespace (%s): %v", net.Name, net.Namespace, err)
		if try && apierrors.IsNotFound(err) {
			logging.Verbosef(errMsg)
			return nil, resourceMap, fmt.Errorf("getKubernetesDelegate: " + errMsg)
		}
```

s1061123: In such case, we expect that net-attach-def is created with empty. In your case, you need to create this net-attach-def.

```
---
metadata:
  name: cilium
spec:
```

s1061123: In addition, if you proceed to introduce this argument, I recommend to change the name `try` to something because `try` is not clear what it does. Anyway, I still prefer not to add new arguments for the purpose.

s1061123: Or you can just specify CNI path in `clusterNetwork`, your requirement might be fullfilled.

If you put cilium CNI config in '/etc/cni/net.d/10-cilium.conflist`, then 

```
{
  "name": "multus-cni-network",
  "type": "multus",
  "confDir": "/etc/cni/net.d",
...
  "clusterNetwork": "/etc/cni/net.d/10-cilium.conflist"
}
```

mskrocki: I dont want to hard code the path `/etc/cni/net.d/10-cilium.conflist`. Any recommendation for the name?

s1061123: First, let's find out how to archieve that without change.

In your case, you can add net-attach-def with the name 'cilim' in namespace default, then it works, I suppose. Did you try that?

s1061123:
So let me summarize what you mentioned. You wan to use some CNI config file for clusterNetwork. 

Currently clusterNetwork have several way to specify cluster network CNI config and you can use 2 or 4 in your case, I suppose.

1. Specify net-attach-def with config 
```
$ kubectl get net-attach-def cbr0 -o yaml | kubectl neat
apiVersion: k8s.cni.cncf.io/v1
metadata:
  name: cbr0
  namespace: default
spec:
  config: '{ "name": "cbr0", "cniVersion": "0.3.1", "plugins": [ { "type": "flannel",
    "delegate": { "hairpinMode": true, "isDefaultGateway": true } }, { "type": "portmap",
    "capabilities": { "portMappings": true } } ] }'
[root@kube-node-1 net.d]# cat /etc/cni/net.d/00-multus.conf
{
  "cniVersion": "0.3.1",
  "name": "multus-cni-network",
  "type": "multus",
  "capabilities": {
    "portMappings": true
  },
  "logLevel": "debug",
  "logFile": "/tmp/multus.log",
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
  "clusterNetwork": "cbr0"
}
```

2. Put file in /etc/cni/multus/net.d with net-attach-def (with empty body)

```
[root@kube-node-1 net.d]# cat /etc/cni/multus/net.d/10-flannel.conflist
{
  "name": "cbr0",
  "cniVersion": "0.3.1",
  "plugins": [
    {
      "type": "flannel",
      "delegate": {
        "hairpinMode": true,
        "isDefaultGateway": true
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    }
  ]
}
[root@kube-node-1 net.d]# cat /etc/cni/net.d/00-multus.conf
{
  "cniVersion": "0.3.1",
  "name": "multus-cni-network",
  "type": "multus",
  "capabilities": {
    "portMappings": true
  },
  "logLevel": "debug",
  "logFile": "/tmp/multus.log",
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
  "clusterNetwork": "cbr0"
}
$ kubectl get net-attach-def cbr0 -o yaml
apiVersion: k8s.cni.cncf.io/v1
metadata:
  creationTimestamp: "2023-03-09T15:45:59Z"
  generation: 1
  name: cbr0
  namespace: default
  resourceVersion: "5578"
  uid: a8d3e8e0-e7dd-40fe-9709-71baab43e50d
```

3. Specify full path of CNI config JSON

```
[root@kube-node-1 ~]# cat /etc/cni/flannel/10-flannel.conflist
{
  "name": "cbr0",
  "cniVersion": "0.3.1",
  "plugins": [
    {
      "type": "flannel",
      "delegate": {
        "hairpinMode": true,
        "isDefaultGateway": true
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    }
  ]
}
[root@kube-node-1 ~]# cat /etc/cni/net.d/00-multus.conf
{
  "cniVersion": "0.3.1",
  "name": "multus-cni-network",
  "type": "multus",
  "capabilities": {
    "portMappings": true
  },
  "logLevel": "debug",
  "logFile": "/tmp/multus.log",
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
  "clusterNetwork": "/etc/cni/flannel/10-flannel.conflist"
}
```

4. Specify directory path for CNI config JSON

```
[root@kube-node-1 ~]# cat /etc/cni/flannel/10-flannel.conflist
{
  "name": "cbr0",
  "cniVersion": "0.3.1",
  "plugins": [
    {
      "type": "flannel",
      "delegate": {
        "hairpinMode": true,
        "isDefaultGateway": true
      }
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      }
    }
  ]
}
[root@kube-node-1 ~]# cat /etc/cni/net.d/00-multus.conf
{
  "cniVersion": "0.3.1",
  "name": "multus-cni-network",
  "type": "multus",
  "capabilities": {
    "portMappings": true
  },
  "logLevel": "debug",
  "logFile": "/tmp/multus.log",
  "kubeconfig": "/etc/cni/net.d/multus.d/multus.kubeconfig",
  "clusterNetwork": "/etc/cni/flannel"
}
```

I suppose these option could be feasible in your case. Could you please try that?

pkg/k8sclient/k8sclient.go
mskrocki: I do not want to maintain the "cilium" NAD, which means I have to create it, handle updates/upgrades and ensure it is never deleted, so that option does not work for me.

mskrocki:
As mentioned above, options 1 and 2 are not feasible, 3 require a hardcoded path and 4 will require changes to my default configuration, when multus in not in the picture. I have changed my approach in this PR, PTAL.

dougbtv:
hold up guys -- I feel like there's some misunderstanding here. Can we chat about this at the Multus maintainer's call next week? I want to make sure we can move forward to help your situation, but keep the current behavior intact as well.

https://github.com/k8snetworkplumbingwg/community#topic-specific-meetings

mskrocki:
Sounds good, added agenda item.

pkg/k8sclient/k8sclient.go
```diff
@@ -536,9 +536,16 @@ func getNetDelegate(client *ClientInfo, pod *v1.Pod, netname, confdir, namespace
 			Name:      netname,
 			Namespace: namespace,
 		}
-		delegate, resourceMap, err := getKubernetesDelegate(client, net, confdir, pod, resourceMap)
+
+		// check NAD exists first
+		_, err := client.NetClient.NetworkAttachmentDefinitions(net.Namespace).Get(context.TODO(), net.Name, metav1.GetOptions{})
 		if err == nil {
```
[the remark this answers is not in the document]
mskrocki: done

mskrocki:
Ping, I have addressed all the comments, PTAL.

mskrocki:
bump

PR #1048, opened 2023-03-08
upgrade containernetworking plugins to v1.2.0

panguicai008:


s1061123:
Is there any clear reason why you need v1.2.0?

panguicai008:
> Is there any clear reason why you need v1.2.0?

At first, I was missing this package for the multus-cni project, and then went to download it use go get command . I found that v1.2.0 is latest. Then I found that v1.1.1 version is patch release that fixes some bugs in v1.1.0, and v1.2.0 also fix some bugs and  enhanced some functions 

https://github.com/containernetworking/plugins/releases/tag/v1.1.1
https://github.com/containernetworking/plugins/releases/tag/v1.2.0

PR #1053, opened 2023-03-27
Improve v4 config

maiqueb:
Adjust the user configuration to take place from a single place: the `configMap`.

Keeps only the bare minimum of command line options (in particular: where the config file lives, and the version flag).

Supersedes: #1006

### DIFF
--- a/pkg/server/api/shim.go
+++ b/pkg/server/api/shim.go
@@ -34,7 +34,7 @@ type ShimNetConf struct {
 	// types.NetConf
 
 	CNIVersion      string `json:"cniVersion,omitempty"`
-	MultusSocketDir string `json:"socketDir"`
+	MultusSocketDir string `json:"shimSocketDir"`
 	LogFile         string `json:"logFile,omitempty"`
 	LogLevel        string `json:"logLevel,omitempty"`
 	LogToStderr     bool   `json:"logToStderr,omitempty"`
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -143,7 +143,7 @@ func NewCNIServer(daemonConfig *types.ControllerNetConf, serverConfig []byte) (*
 		logging.Verbosef("server configured with chroot: %s", daemonConfig.ChrootDir)
 	}
 
-	return newCNIServer(daemonConfig.MultusSocketDir, kubeClient, exec, serverConfig)
+	return newCNIServer(daemonConfig.DaemonSocketDir, kubeClient, exec, serverConfig)
 }
 
 func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec, servConfig []byte) (*Server, error) {
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -62,12 +62,21 @@ const (
 func main() {
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
+	//XXX: one section (multus CNI config)
 	cniConfigDir := flag.String("cni-config-dir", defaultCniConfigDir, "CNI config dir")
 	multusConfigFile := flag.String("multus-conf-file", "auto", "The multus configuration file to use. By default, a new configuration is generated.")
+	cniVersion := flag.String("cni-version", "", "Allows you to specify CNI spec version. Used only with --multus-conf-file=auto.")
 	multusMasterCni := flag.String("multus-master-cni-file", "", "The relative name of the configuration file of the cluster primary CNI.")
 	multusAutoconfigDir := flag.String("multus-autoconfig-dir", *cniConfigDir, "The directory path for the generated multus configuration.")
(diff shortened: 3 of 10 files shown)


pkg/server/api/shim.go
```diff
@@ -120,6 +120,8 @@ func shimConfig(cniConfig []byte) (*ShimNetConf, error) {
 	}
 	if multusConfig.MultusSocketDir == "" {
 		multusConfig.MultusSocketDir = defaultMultusRunDir
+	} else {
+		multusConfig.MultusSocketDir = strings.ReplaceAll(multusConfig.MultusSocketDir, "/host", "")
```
maiqueb, line 124: @dougbtv / @s1061123 this is the most fishy part in this PR IMO ...

Seems we've always had a bug when specifying the socket path - from the daemon's perspective, the socket will be in `/host/<whatever>` while from the shim's perspective it'll be in `<whatever>` . 

... since we were relying on the defaults (and the shim / daemon used different defaults) we were not seeing it.

I don't have a smarter idea right now on how to achieve this to be honest. Providing a new know for the socket path from the shim's perspective seems weird. And keep in mind this must be configured via the volume / hostPath mapping as well, so, the configuration must be aligned.

maiqueb:
The e2e tests work for multus config CNI version 0.3.1 but fail for multus config CNI version 0.4.0, with the following error:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/c129e72779ce2d97f848f5409f6b289846e4b7d8/pkg/server/config/generator.go#L110

Not sure how to proceed.

pkg/server/api/shim.go
s1061123: Your code change seems to make consistency between shim's socket parameter and multus-daemon's socket parameter, however, I believe that we should explicitly have shim's socket directory option and multus-daemon's socket directory as separate item.

The reason is user may change multus-daemonset.yml manifest and change '/host' prefix. At present, multus-daemon   could be still run as host process (just unix-daemon). So your code implicitly assumes that multus-daemon must runs in a pod always hence multus-daemon no longer run as host process. In another case, if user wants to change host prefix from '/host' to something different, then it may be broken.

For now, we should expect that if some user want to change these parameter, then they should understand what the thick plugin and shim plugin are and they should make consistency between multus-daemon's socket path and multus-shim's socket path manually. I suppose it should make sense (even though it is not good UI).

From multus-shim and multus-daemon component point of view, we should not have such verification I believe.
If we need to do that, this could be done in manifest orchestration or some manifest verification , I suppose.

maiqueb: Makes sense; I understand my current proposal is pretty bad, for the reasons you state.

... but it's good at showing something's wrong :) 

How should I call the parameters ? serverSocketDir / daemonSocketDir and clientSocketDir / shimSocketDir ? 

I would go with daemonSocketDir and shimSocketDir. Any objections @s1061123 ?

s1061123: daemonSocketDir and shimSocketDir are good name, thanks!

maiqueb:
> The e2e tests work for multus config CNI version 0.3.1 but fail for multus config CNI version 0.4.0, with the following error:
> 
> https://github.com/k8snetworkplumbingwg/multus-cni/blob/c129e72779ce2d97f848f5409f6b289846e4b7d8/pkg/server/config/generator.go#L110
> 
> Not sure how to proceed.

Using latest master branch this is the cluster CNI configuration on kind:
```
kubectl exec -nkube-system kube-multus-ds-amd64-j77dn -- cat /host/etc/cni/net.d/10-kindnet.conflist | jq
Defaulted container "kube-multus" out of: kube-multus, install-multus-shim (init)
{
  "cniVersion": "0.4.0",
  "name": "kindnet",
  "plugins": [
    {
      "ipMasq": false,
      "ipam": {
        "dataDir": "/run/cni-ipam-state",
        "ranges": [
          [
            {
              "subnet": "10.244.1.0/24"
            }
          ]
        ],
        "routes": [
          {
            "dst": "0.0.0.0/0"
          }
        ],
        "type": "host-local"
      },
      "mtu": 1500,
      "type": "ptp"
    },
    {
      "capabilities": {
        "portMappings": true
      },
      "type": "portmap"
    }
  ]
}
```
Seems I need somehow to ensure the provided CNI_VERSION trickles down to the delegate configuration. Any idea on how to do that @s1061123 ?

pkg/server/api/shim.go
dougbtv: +1 on the naming

pkg/server/server.go
```diff
@@ -89,6 +89,7 @@ func (s *Server) HandleDelegateRequest(cmd string, k8sArgs *types.K8sArgs, cniCm
 	var err error
 	var multusConfByte []byte
 
+	// !bang need to not have both of these?
```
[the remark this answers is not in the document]
dougbtv: You can remove this rem. This was notes I was stubbing in when I was finding what to change, thanks

pkg/server/api/shim.go
maiqueb: Done.

maiqueb:
Woop Woop! 

The tests work. Please review @dougbtv / @s1061123 .

PR #1055, opened 2023-03-31
docs, thick plugin: align docs with new configuration reference

maiqueb:
PR #1053 changed the thick plugin configuration to happen exclusively via the user provided config map.

This PR aligns the multus documentation with the existing code.

It also renames the multus socket dir configuration to something that does not stutter (daemon.DaemonSocketDir stutters).

### DIFF
--- a/docs/thick-plugin.md
+++ b/docs/thick-plugin.md
@@ -53,37 +53,58 @@ kubectl apply -f deployments/multus-daemonset-thick.yml
 
 ### Command line parameters
 
-Multus thick plugin variant accepts the same
-[entrypoint arguments](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#entrypoint-script-parameters)
-its thin counterpart allows - with the following exceptions:
-
-- `additional-bin-dir`
-- `binDir`
-- `cleanup-config-on-exit`
-- `cniDir`
-- `multus-kubeconfig-file-host`
-- `rename-conf-file`
-- `skip-multus-binary-copy`
-
-It is important to refer that these are command line parameters to the golang
-binary; as such, they should be passed using a single dash ("-") e.g.
-`-additional-bin-dir=/opt/multus/bin`, `-multus-log-level=debug`, etc.
-
-Furthermore, it also accepts a new command line parameter, where the user
-specifies the path to the server configuration:
+The available command line parameters are:
 
 - `config`: Defaults to `"/etc/cni/net.d/multus.d/daemon-config.json"`
-- `metricsPort`: Metrics port (of multus' metric exporter), default is disable
+- `version`: Prints the daemon config version and exits
 
-### Server configuration
+### Server / Daemon configuration
 
 The server configuration is encoded in JSON, and allows the following keys:
 
 - `"chrootDir"`: Specify the directory which points to host root from the pod. See 'Chroot configuration' section for the details.
-- `"socketDir"`: Specify the location where the unix domain socket used for
-client/server communication will be located. Defaults to `"/run/multus"`.
+- `"socketDir"`: Specify the location where the unix domain socket used
+for client/server communication will be located. This is the location where the
+**Daemon** will read the configuration from. Defaults to `"/run/multus"`.
+- `"metricsPort"`: Metrics port (of multus' metric exporter); by default, no port
+is provided.
+- `"logFile"`: the path to where the daemon logs will be persisted.
+- `"logLevel"`: the logging level for the multus daemon logs.
+- `"logToStderr"`: enable this to have the daemon multus logs echoed to stderr
+as well. By default, it is disabled.
 
 In addition, you can add any configuration which is in [configuration reference](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md#multus-cni-configuration-reference). Server configuration override multus CNI configuration (e.g. `/etc/cni/net.d/00-multus.conf`)
+
+Below you can see an example of the daemon configuration:
+```json
+{
+        "chrootDir": "/hostroot",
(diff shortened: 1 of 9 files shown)


docs/thick-plugin.md
```diff
@@ -53,18 +53,6 @@ kubectl apply -f deployments/multus-daemonset-thick.yml
...
-- `binDir`
-- `cleanup-config-on-exit`
-- `cniDir`
-- `multus-kubeconfig-file-host`
-- `rename-conf-file`
-- `skip-multus-binary-copy`
-
 It is important to refer that these are command line parameters to the golang
 binary; as such, they should be passed using a single dash ("-") e.g.
 `-additional-bin-dir=/opt/multus/bin`, `-multus-log-level=debug`, etc.
```
s1061123, line 58: `-additional-bin-dir` and `-multus-log-level` are no longer exist, so we may change.

For your info, now we only have `-config` and `-version`.
```
$ ./bin/multus-daemon --help
Usage of ./bin/multus-daemon:
  -config string
        Specify the path to the multus-daemon configuration (default "/etc/cni/net.d/multus.d/daemon-config.json")
  -version
        Show version
$
```

docs/thick-plugin.md
```diff
@@ -73,18 +61,38 @@ Furthermore, it also accepts a new command line parameter, where the user
...
+- `"logFile"`: the path to where the daemon logs will be persisted.
+- `"logLevel"`: the logging level for the multus daemon logs.
+- `"logToStderr"`: enable this to have the daemon multus logs echoed to stderr
+as well. By default, it is disabled.
 
 In addition, you can add any configuration which is in [configuration reference](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/configuration.md#multus-cni-configuration-reference). Server configuration override multus CNI configuration (e.g. `/etc/cni/net.d/00-multus.conf`)
 
+### Client / Shim configuration
+
+The multus shim configuration is encoded in JSON, and allows the following keys:
```
s1061123, line 85: First reader may be confused about what the 'Shim configuration'. How about to mention that shim consumes usual CNI config as shim config and it might be in `/etc/cni/net.d/00-multus.conf`?

s1061123 (changes requested):

docs/thick-plugin.md
maiqueb: Done.

s1061123 (approved):
LGTM, thanks!

docs/thick-plugin.md
```diff
@@ -53,38 +53,62 @@ kubectl apply -f deployments/multus-daemonset-thick.yml
...
+regular CNI configuration, usually available in `/etc/cni/net.d/00-multus.conf`.
+
+It allows the following keys:
+
+- `"cniVersion"`: the CNI version for the Multus CNI plugin.
+- `"logFile"`:  the path to where the daemon logs will be persisted.
+- `"logLevel"`: the logging level for the multus daemon logs.
+- `"logToStderr"`: enable this to have the daemon multus logs echoed to stderr
+  as well. By default, it is disabled.
+- `"shimSocketDir"`: Specify the location where the unix domain socket used for
```
s1061123, line 108: I'm curious that "shimScoketDir" option name in multus-shim. This option specifies server's socket dir, so we may change that to 'daemonSocketDir' here? "shimSocketDir" looks like the different soket from above.

We may change 'daemonSocketDir' to 'socketDir' in multus-daemon side as well. What do  you think about that?

maiqueb: I don't follow. This was intentional - it's the same socket, but located at a different path. 

Since afaiu we de-serialize the same object to both the shim config and to the daemon config, IMO, it makes sense to use different attribute names. Having the same attribute name, is, imo, confusing.

s1061123, line 108: "shimSocketDir" is defined in ControllerNetConf , so "shimSocketDir" definition could go to daemon configuration

https://github.com/k8snetworkplumbingwg/multus-cni/blob/7cf1c2f0c2af0cdabb47e00d9a7fd8290981e197/pkg/types/types.go#L185-L204

maiqueb: Ah! I see. 

You mean to say I should document the shim attribute in the daemon section ? 

I personally wouldn't do that, since it isn't actually **used** from the daemon - it's there just so the configurations between the shim / daemon align. I now realize that might not be needed: I think it's possible to actually remove it from [the code](https://github.com/k8snetworkplumbingwg/multus-cni/blob/6c2e09529bcfdf2f5b7b7999df6ee43080d8bd57/pkg/types/types.go#L201), since it is not being used. 

Thoughts @s1061123 ?

s1061123: Not everything. "logFile"/"logLevel" and others are actually in shim side config. I'm just curiouos that ShimSocketDir is in daemonConfig but you mentioned in shim config. I'm okey if you remove 'ShimSocketDir' code from daemon side and update document (remove 'ShimSocketDir').

maiqueb: Will do! Thank you.

Just for clarification - the `ShimSocketDir` is still required on the shim side. We need to instruct multus CNI where the socket is found (and it is in a different path than the one exposed by the `daemonSocketDir`).

s1061123: I suppose we just name 'socketDir'. If you don't want to do that, we could name
- 'socketDir' in daemon config and 
- 'daemonSocketDir' in shim config

because from daemon config side, 'daemon socket' seems redundant (as usual httpd.conf uses 'Port', not 'httpPort').

maiqueb: Makes sense; now that there aren't 2 socket dirs here, having the prefix stopped making sense.

s1061123 (approved):
LGTM. thanks!

PR #1066, opened 2023-04-07
Refactoring thick daemon config processing

### LINKED ISSUE #1058: Thick daemonset deployment fills unexpected config in '/etc/cni/net.d/00-multus.conf'
Default thick deployment file, `deployments/multus-daemonset-thick.yml`, generates following CNI config in '/etc/cni/net.d'

```
{
  "capabilities": {
    "portMappings": true
  },
  "cniVersion": "0.3.1",
  "logFile": "/tmp/multus.log",
  "logLevel": "verbose",
  "logToStderr": true,
  "logOptions": {
    "maxAge": 0,
    "maxSize": 0,
    "maxBackups": 0,
    "compress": false
  },
  "name": "multus-cni-network",
  "clusterNetwork": "/host/etc/cni/net.d/10-flannel.conflist",
  "type": "multus-shim",
  "cniConfigDir": "/host/etc/cni/net.d",
  "socketDir": "/host/run/multus/"
}
```

There are several unexpected value here as following: 
- `socketDir` should be `shimSocketDir` as #1053 
- `cniConfigDir` should be `/etc/cni/net.d` (because this CNI config is for container host), but it is `/host/etc/cni/net.d`
- `logOptions` should be removed (`logOptions` should be omitted here to leave it as default value)

s1061123:
to damonset config file, hence command line option parsing is no longer used. This change removes these parts.

Fix #1058.

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -34,26 +34,8 @@ import (
 	srv "gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/server"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/server/api"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/server/config"
-	"gopkg.in/k8snetworkplumbingwg/multus-cni.v3/pkg/types"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-)
-
-const (
-	multusPluginName                    = "multus-shim"
-	defaultCniConfigDir                 = "/etc/cni/net.d"
-	defaultMultusGlobalNamespaces       = ""
-	defaultMultusLogFile                = ""
-	defaultMultusLogMaxSize             = 100 // megabytes
-	defaultMultusLogMaxAge              = 5   // days
-	defaultMultusLogMaxBackups          = 5
-	defaultMultusLogCompress            = true
-	defaultMultusLogLevel               = ""
-	defaultMultusLogToStdErr            = false
-	defaultMultusMasterCNIFile          = ""
-	defaultMultusNamespaceIsolation     = false
-	defaultMultusReadinessIndicatorFile = ""
-	defaultSocketDir                    = "/run/multus/"
 )
 
 func main() {
@@ -62,7 +44,7 @@ func main() {
 	// keep in command line option
 	version := flag.Bool("version", false, "Show version")
 
-	configFilePath := flag.String("config", types.DefaultMultusDaemonConfigFile, "Specify the path to the multus-daemon configuration")
+	configFilePath := flag.String("config", srv.DefaultMultusDaemonConfigFile, "Specify the path to the multus-daemon configuration")
 
 	flag.Parse()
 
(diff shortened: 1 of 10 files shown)


cmd/multus-daemon/main.go
```diff
@@ -34,35 +34,17 @@ import (
...
-	defaultMultusLogMaxBackups          = 5
-	defaultMultusLogCompress            = true
-	defaultMultusLogLevel               = ""
-	defaultMultusLogToStdErr            = false
-	defaultMultusMasterCNIFile          = ""
-	defaultMultusNamespaceIsolation     = false
-	defaultMultusReadinessIndicatorFile = ""
-	defaultSocketDir                    = "/run/multus/"
-)
-
```
dougbtv, line 58: I take it these weren't used anymore? The only advantage is/was that you could see them at a glance...

deployments/multus-daemonset-thick.yml
```diff
@@ -112,11 +112,7 @@ data:
     {
         "chrootDir": "/hostroot",
         "confDir": "/host/etc/cni/net.d",
-        "logToStderr": true,
         "logLevel": "verbose",
-        "logFile": "/tmp/multus.log",
-        "binDir": "/opt/cni/bin",
-        "cniDir": "/var/lib/cni/multus",
```
dougbtv, line 119: I take it these are defaults now so they don't need to be specified? 

The only advantage is/was that you'd know where those locations are

pkg/types/conf.go
```diff
@@ -418,36 +418,6 @@ func LoadNetConf(bytes []byte) (*NetConf, error) {
...
-	// const block for multus-daemon configs
-
-	// DefaultMultusDaemonConfigFile is the Default path of the config file
-	DefaultMultusDaemonConfigFile = "/etc/cni/net.d/multus.d/daemon-config.json"
-	// DefaultMultusRunDir specifies default RunDir for multus
-	DefaultMultusRunDir           = "/run/multus/"
-)
-
-// LoadDaemonNetConf loads the configuration for the multus daemon
-func LoadDaemonNetConf(config []byte) (*ControllerNetConf, error) {
```
dougbtv, line 431: My only concern with moving this from here is that this could be more difficult to find when you're making other configuration changes

cmd/multus-daemon/main.go
dougbtv: Ahhhh ha, I see it, these are just for the command line options. Got it. +1.

s1061123: Yeah, that's right. Now command line options are removed so it is removed. But default value is still in `pkg/server`.

deployments/multus-daemonset-thick.yml
s1061123: Yeah, these are removed because 

- "binDir" and "cniDir" are same as default value so we don't have to add it, and
- "logFile" is mainly for debug and we already have logs from daemon log (i.e. kubectl log <multus server pod>)

If some user want to change any config, then he/she should read configration guide and know which configuration parameter is the one. Otherwise configuration file should be simple. Otherwise, we should have all configuration parameter in the deployment file (with default value), but this approach seems to be redundant (default parameter is in config file and go code). I believe that the approach is same as various projects (such as kubeadm, apache httpd and so on). They do not have all params in the default config file.

pkg/types/conf.go
s1061123: I move this function to 'pkg/server' because of the function scope. I would like to specify that this function is only for 'thick' plugin.

Currently we have 'thick' and 'thin' plugin, so put it in same place as 'thin' pluign, it seems to be easy, however, it may confuse to change the code or to troubleshoot the code because thin/thick both codes are in same packages.

This might be a trade-off, so there is not definitive answer for that...

deployments/multus-daemonset-thick.yml
dougbtv: thanks for explanation, agreed on removals, and agree we don't have to have all the options filled out.

pkg/types/conf.go
dougbtv: I think I'm good with this -- it's ok that it's server specific.

dougbtv (approved):

PR #1085, opened 2023-05-02
Fix potential issue, slice bounds out of range error

s1061123:


s1061123:
There is no open issue. This is only happen in my enrionment (i.e. super experiment config and believe none test such thing).

zshi-redhat:
@s1061123 is this ready to merge?
cc @dougbtv

dougbtv (approved):

PR #1090, opened 2023-05-09
Fix for CRIO Thin Client Installations

rwlove:
1. Use the correct entrypoint

2. Use /opt/cni/bin for CNI bin dir

### DIFF
--- a/deployments/multus-daemonset-crio.yml
+++ b/deployments/multus-daemonset-crio.yml
@@ -181,7 +181,7 @@ spec:
       - name: kube-multus
         # crio support requires multus:latest for now. support 3.3 or later.
         image: ghcr.io/k8snetworkplumbingwg/multus-cni:stable
-        command: ["/entrypoint.sh"]
+        command: ["/thin_entrypoint"]
         args:
         - "--cni-version=0.3.1"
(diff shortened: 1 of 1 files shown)


maiqueb:
@s1061123 does this separate crio daemonset spec even makes sense these days ?

s1061123 (commented):
Thank you for the PR, @rwlove !

I make sense the entrypoint fix, however,  I'd like to keep '/usr/libexec/cni' for CNI path. There are two reason:

- '/usr/libexec/cni' is the default path of Fedora's CNI plugin path. In Fedora38, for example, `dnf install  containernetworking-plugins` installs CNI plugins into '/usr/libexec/cni'.
- If user uses '/opt/cni/bin' for CNI plugin path, user can use `multus-daemonset.yml` instead of `multus-daemonset-crio.yml` because now the difference between '....yml' and '...-crio.yml' (previously we have more difference but now only the CNI path).

So could you please revert the change of CNI path? Thank you again and in advance.

rwlove:
I understand your feedback and I'll happily update the patch and send a new PR.

rwlove:
did I update this correctly?

PR #1095, opened 2023-05-17
Fix multus to support CNI plugin which does not create interface

s1061123:
This change fixes multus to support CNI plugin which does not create any interface and return empty result. Some CNI plugin may do network configuration change only and does not create any interface and return empty CNI result. Current multus assumes that CNI config always creates some interfaces hence above CNI plugin is out of assumption and multus may not work with such plugins.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -369,11 +369,14 @@ func DelegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, dele
 	}
 
 	if pod != nil {
-		// send kubernetes events
-		if delegate.Name != "" {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v from %s", rt.IfName, ips, delegate.Name)
-		} else {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v", rt.IfName, ips)
+		// check Interfaces and IPs because some CNI plugin just return empty result
+		if res.Interfaces != nil || res.IPs != nil {
+			// send kubernetes events
+			if delegate.Name != "" {
+				kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v from %s", rt.IfName, ips, delegate.Name)
+			} else {
+				kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v", rt.IfName, ips)
+			}
 		}
 	} else {
 		// for further debug https://github.com/k8snetworkplumbingwg/multus-cni/issues/481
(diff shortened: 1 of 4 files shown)


pkg/multus/multus.go
```diff
@@ -369,11 +369,14 @@ func DelegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, dele
 	}
 
 	if pod != nil {
-		// send kubernetes events
-		if delegate.Name != "" {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v from %s", rt.IfName, ips, delegate.Name)
-		} else {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v", rt.IfName, ips)
+		// check Interfaces and IPs because some CNI plugin just return empty result
+		if res.Interfaces != nil || res.IPs != nil {
```
mlguerrero12, line 373: checking interfaces should be enough? In which case the interfaces is equal to nil and IPs not nil?

s1061123: In case of CNI 0.2.0, there is no 'interface' in result object, so we also need to double check IPs as well.

pkg/multus/multus.go
```diff
@@ -636,6 +639,17 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
 
+		res, err := cni100.NewResultFromResult(tmpResult)
+		if err != nil {
+			logging.Errorf("CmdAdd: failed to read result: %v, but proceed", err)
+		}
+
+		// check Interfaces and IPs because some CNI plugin does not create any interface
+		// and just returns empty result
+		if res.Interfaces == nil && res.IPs == nil {
+			continue
```
mlguerrero12, line 650: it will skip the CreateNetworkStatus of those type of plugins

s1061123: I expect that CNI plugin without interfaces and IPs, so I suppose we don't need to create network status for that because this CNI plugin does not create any network interface.

pkg/multus/multus.go
```diff
@@ -636,6 +639,17 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			return nil, cmdPluginErr(k8sArgs, netName, "error adding container to network %q: %v", netName, err)
 		}
 
+		res, err := cni100.NewResultFromResult(tmpResult)
+		if err != nil {
+			logging.Errorf("CmdAdd: failed to read result: %v, but proceed", err)
```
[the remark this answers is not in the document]
mlguerrero12: I don't understand, if there is an error converting, that function returns
	if err != nil {
		return nil, err
	}

and res will be equal to nil. Below you use res.Interfaces and res could be nil

What am I missing?

pkg/multus/multus.go
mlguerrero12: but what happens if you run a regular cni like bridge version 0.2.0 without IPAM?

pkg/multus/multus.go
s1061123: changed.

pkg/multus/multus.go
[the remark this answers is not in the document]
s1061123: `cni100.NewResultFromResult(tmpResult)` is just convert result to cni100. Let's imagine that tmpResult has valid CNI result, but failed to convert cni100. In such case, multus need to return tmpResult as valid result because NewResultFromResult() failure is just failed to convert 1.0.0.

Do you guarantee that all CNI version result object can be converted 1.0.0 forever?

pkg/multus/multus.go
s1061123: 0.2.0 does require IPAM. Please see CNI 0.2.0 SPEC.

pkg/multus/multus.go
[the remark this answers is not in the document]
mlguerrero12: I understand that.

But let me ask you something, what is the value of "res" if the function cni100.NewResultFromResult(tmpResult) returns an error?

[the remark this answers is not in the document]
mlguerrero12: perhaps we need another pair of eyes, @dougbtv, I'm claiming that the function cni100.NewResultFromResult will return (nil, err) if there is an error with convert. Therefore, with this "res, err := cni100.NewResultFromResult(tmpResult)", res will be equal to "nil" and then the code below will panic "if res.Interfaces != nil || res.IPs != nil"

func NewResultFromResult(result types.Result) (*Result, error) {
	newResult, err := convert.Convert(result, ImplementedSpecVersion)
	if err != nil {
		return nil, err
	}
	return newResult.(*Result), nil
}

pkg/multus/multus.go
```diff
@@ -369,11 +369,14 @@ func DelegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, dele
...
 	if pod != nil {
-		// send kubernetes events
-		if delegate.Name != "" {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v from %s", rt.IfName, ips, delegate.Name)
-		} else {
-			kubeClient.Eventf(pod, v1.EventTypeNormal, "AddedInterface", "Add %s %v", rt.IfName, ips)
+		// check Interfaces and IPs because some CNI plugin just return empty result
+		if res.Interfaces != nil || res.IPs != nil {
+			// send kubernetes events
+			if delegate.Name != "" {
```
dougbtv, line 375: thanks for covering this erroneous output in the current version

andreaskaris (commented):
For the sake of it, can we still add proper error handling to netutils.DeleteDefaultGw?

Because the DeleteDefaultGW method is still in itself erroneous. As in, I can write a unit test for that method alone and the unit test will fail:
~~~
// DeleteDefaultGW removes the default gateway from marked interfaces.
func DeleteDefaultGW(netnsPath string, ifName string) error {
	netns, err := ns.GetNS(netnsPath)
	if err != nil {
		return logging.Errorf("DeleteDefaultGW: Error getting namespace %v", err)
	}
	defer netns.Close()

	err = netns.Do(func(_ ns.NetNS) error {
		var err error
		link, _ := netlink.LinkByName(ifName)
		routes, _ := netlink.RouteList(link, netlink.FAMILY_ALL)
		for _, nlroute := range routes {
			if nlroute.Dst == nil {
				err = netlink.RouteDel(&nlroute)
			}
		}
		return err
	})
	return err
}
~~~

I don't understand why there's no error handling here:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/netutils/netutils.go#L41

But in the SetDefaultGW method, there's error handling for exactly the same call:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/netutils/netutils.go#L67

I don't understand the logic that's higher up the callstack, but if you must ignore the error, then I'd throw it away further up the stack, but I wouldn't do the following unless that's really intended to delete whatever default route if the link doesn't exists (and even if it's intended and behaves as designed, I think you should add a comment why):
~~~
		link, _ := netlink.LinkByName(ifName)
		routes, _ := netlink.RouteList(link, netlink.FAMILY_ALL)
~~~

I hope that makes sense

pkg/multus/multus.go
```diff
@@ -636,69 +639,78 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
...
-			err = netutils.DeleteDefaultGW(args.Netns, ifName)
-			if err != nil {
-				return nil, cmdErr(k8sArgs, "error deleting default gateway: %v", err)
-			}
-			err = netutils.DeleteDefaultGWCache(n.CNIDir, rt, netName, ifName, deleteV4gateway, deleteV6gateway)
-			if err != nil {
-				return nil, cmdErr(k8sArgs, "error deleting default gateway in cache: %v", err)
+		// check Interfaces and IPs because some CNI plugin does not create any interface
+		// and just returns empty result
+		if res != nil &&  (res.Interfaces != nil || res.IPs != nil) {
```
andreaskaris, line 654: What about moving the block's content to a helper function and doing:
~~~
if res != nil &&  (res.Interfaces != nil || res.IPs != nil) {
   if err := configureDefaultGateway(delegate) != nil {
     return nil, err
   }
}
~~~

That'd make it easier to read instead of having a big if condition

dougbtv:
Andreas -- thanks for the detailed input/review.

I think I'm OK with fixing this kind of at the handler level for now, I like how you dug a layer deeper. And agreed it sure looks like it's inconsistent error handling.

I do like how this exposes what's happening with a non-interface creating CNI plugin at the handler level, but that doesn't mean we shouldn't fix what's under the surface, so I've filed that in #1098 and I think it's worth looking at as well as a follow up.

dougbtv (approved):

PR #1096, opened 2023-05-17
[backport release-v3] Fix multus to support CNI plugin which does not create interface 

dougbtv:
This change fixes multus to support CNI plugin which does not create any interface and return empty result. Some CNI plugin may do network configuration change only and does not create any interface and return empty CNI result. Current multus assumes that CNI config always creates some interfaces hence above CNI plugin is out of assumption and multus may not work with such plugins.

Backport of #1095


s1061123 (approved):
Looks good to me, thanks!

PR #1116, opened 2023-06-30
pass through the pod annotations when multus receives them

toelke:
See https://github.com/kubernetes/kubernetes/issues/69882#issuecomment-1509077177

Passing the annotations seems not to be a standardized behaviour, but it is very useful nontheless.


toelke:
ping

PR #1124, opened 2023-07-19
Fix bug: when kube-multus-ds restart,  CNI can not load 00-multus.conf

huangzynn:
When the server restart,  and  kube-multus-ds will restart, sometimes k8s dockershim cni load conf Will load calico instead of multus，Resulting in POD not being able to obtain IP
Such as:
podLostIP1
Read this part of the code:
https://github.com/kubernetes/kubernetes/blob/bc4763cbf8b9e6adfbc07bb3194509b187b0b901/pkg/kubelet/dockershim/network/cni/cni.go#L156
When CNI reads the configuration, copying files or generating configuration files from multiple will cause the next CNI configuration to be loaded


s1061123:
I'm not clear what you mentioned.

- k8s dockershim is deprecated and it is removed hence now no dockershim in k8s repo, hence your explanation of this issue is not clear to me: https://kubernetes.io/blog/2020/12/02/dont-panic-kubernetes-and-docker/
- your modification in deployments/multus-daemonset-thick.yml and pkg/cmdutils/utils.go seems not directly related to the subject.

huangzynn:
> * cmdutils

The change is to prevent the kube-multus-ds  from copying during restart，kube-multus-ds copy files caused CNI to load an incorrect configuration.The CNI loop checks the first available configuration, and if a configuration error is found or the multi execution file is unreadable, it will skip to the next configuration read.
Read this part of the code:
https://github.com/kubernetes/kubernetes/blob/bc4763cbf8b9e6adfbc07bb3194509b187b0b901/pkg/kubelet/dockershim/network/cni/cni.go#L156

Although dockership is deprecated, we are still using it at present. I am not sure if the same situation exists for other contractors.

remram44:
The reference implementation has a similar problem. If a configuration file is incorrect, it will not silently load the next one like dockershim, but it will fail the whole process:

https://github.com/containernetworking/cni/blob/v1.1.2/libcni/conf.go#L176-L179

Writing the multus config atomically is probably desirable, especially if that happens every time the DaemonSet Pods restart. I'm not sure if the MD5 logic is required since the copy is atomic anyway.

This applies to Calico too, which [uses `os.WriteFile()`](https://github.com/projectcalico/calico/blob/60f8a9e884567ca690487849ff9df991091eced0/cni-plugin/pkg/install/install.go#L400-L405C3) in an init container.

huangzynn:
> The reference implementation has a similar problem. If a configuration file is incorrect, it will not silently load the next one like dockershim, but it will fail the whole process:
> 
> https://github.com/containernetworking/cni/blob/v1.1.2/libcni/conf.go#L176-L179
> 
> Writing the multus config atomically is probably desirable, especially if that happens every time the DaemonSet Pods restart. I'm not sure if the MD5 logic is required since the copy is atomic anyway.
> 
> This applies to Calico too, which [uses `os.WriteFile()`](https://github.com/projectcalico/calico/blob/60f8a9e884567ca690487849ff9df991091eced0/cni-plugin/pkg/install/install.go#L400-L405C3) in an init container.

Thank you.

PR #1136, opened 2023-08-09
Enables multiplatform thick plugin images

nvollmar:
v4.0.0-alpha (https://github.com/k8snetworkplumbingwg/multus-cni/pkgs/container/multus-cni/47662597?tag=v4.0.0-alpha) was published as thick plugin for arm64 and I've been running it for almost a year now without any issues.

Would be great to have it available for current versions on arm64.

### DIFF
--- a/.github/workflows/image-build.yml
+++ b/.github/workflows/image-build.yml
@@ -35,8 +35,8 @@ jobs:
           sbom: false
           provenance: false
 
-  build-amd64-thick:
-    name: Image build/amd64 thick plugin
+  build-thick:
+    name: Image build thick plugin
     runs-on: ubuntu-latest
     steps:
       - name: Check out code into the Go module directory
@@ -50,8 +50,11 @@ jobs:
         with:
           context: .
           push: false
-          tags: ghcr.io/${{ github.repository }}:latest-amd64-thick
+          tags: ghcr.io/${{ github.repository }}:latest-thick
           file: images/Dockerfile.thick
+          platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le,linux/s390x
+          sbom: false
+          provenance: false
 
   build-origin:
     name: Image build/origin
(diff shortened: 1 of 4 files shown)


nvollmar:
Refs #1027

s1061123:
See following comments. We need to have support resources who could care all platform other than amd64, for thick plugin:
https://github.com/k8snetworkplumbingwg/multus-cni/pull/1027#issuecomment-1425272290

Can you do that?

nvollmar:
@s1061123 I'd be open to test things on my arm64 k3s cluster. I'm also actively using multus for my HA setup and I tend to update frequently anyway.

s1061123:
@nvollmar I need someone who can 
- troubleshoot ARM/ppc64/s390x related issue in thick plugin and
- fix these issues, and
- verify these issue in target platform.

Could you help these items (you mentioned you have arm64, but how about other platform)? 
In addition, multus team support multus with vanilla Kubernetes (not k3s, not OpenShift and not other downstream products), to simplify. So we need to ask you to verify vanilla Kubernetes for above troubleshooting/fix/verify.

nvollmar:
I only have ARM available to test on and i don't have any spare hardware to test vanilla Kubernetes on ARM.

May I ask, you already publish the thin image for ARM/PPC64/S390, what is the difference for not publishing the thick image too?

zhlhahaha:
Hi, @nvollmar and @s1061123 , I am Howard from Arm. I also get this issue when trying to use multus on Arm64 platform.

> * troubleshoot ARM/ppc64/s390x related issue in thick plugin and

As I only have Arm64 servers, shall we build multi-arch image for x86_64 and Arm64 for now? 

> * fix these issues, and verify these issue in target platform.

Please feel free to reach out to me if any issues arise on the Arm64 platform. Additionally, I think that @nvollmar is also willing to provide assistance based on his replies.

nvollmar:
I don't understand what the difference between thick and thin images is regarding platform support (thin images are published for them already).

I'd be willing to provide help regarding ARM, but with the lack of responses here I started to build and use my own ARM images.

Issue #1130, opened 2023-07-27
Infinite recursion when setting up Multus (delegates to itself)


werdnum:
Trying to install on an arm64 k3s cluster.

Looks like maybe this part of `thin_entrypoint` should be skipping to the next entry if the first one is Multus?

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/cmd/thin_entrypoint/main.go#L302-L316

**What happened**:

Server crashes because of a Multus forkbomb.

Examining the `00-multus.conflist` file, I see that Multus is delegating to itself: https://gist.github.com/werdnum/de21d625ae7c4428ed55914e8da9ef0e

**What you expected to happen**:

Should delegate to the first non-Multus CNI instead of to itself

**How to reproduce it (as minimally and precisely as possible)**:

Install this on a k3s cluster, label one of the nodes with `network-implementation: multus` https://gist.github.com/werdnum/9f36b4084f39924b74d27824ee410966

**Anything else we need to know?**:

**Environment**:

- Multus version 

sha256:35accc0395c03a8f0706304d79d670602debf4b31228f9da76b8e415c0523387
https://github.com/k8snetworkplumbingwg/multus-cni/pkgs/container/multus-cni/112115274?tag=snapshot

- Kubernetes version (use `kubectl version`):

Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:51:05Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.3+k3s1", GitCommit:"fe9604cac119216e67f8bd5f14eb5608e3bcd3cf", GitTreeState:"clean", BuildDate:"2023-06-26T15:36:47Z", GoVersion:"go1.20.5", Compiler:"gc", Platform:"linux/arm64"}

- Primary CNI for Kubernetes cluster:

Flannel

- OS (e.g. from /etc/os-release):

Ubuntu 22.04

- File of '/etc/cni/net.d/'

```
udo ls -l /var/lib/rancher/k3s/agent/etc/cni/net.d
total 312
-rw------- 1 root root  24576 Jul 27 22:02 00-multus.conf.new
[...]

PR #1137, opened 2023-08-09
fix multus config file generation to avoid self-delegation

thomasferrandiz:
Add a check to make sure that we don't use an existing multus configuration file. This avoids creating a situation where multus delegates to itself and breaks pod networking by trying multiple times to create the same intefarce for a pod.

This should fix: https://github.com/k8snetworkplumbingwg/multus-cni/issues/1130

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -23,7 +23,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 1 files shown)


cmd/thin_entrypoint/main.go
```diff
@@ -294,7 +296,20 @@ func (o *Options) createMultusConfig() (string, error) {
 		return "", fmt.Errorf("cannot find master CNI config in %q: %v", o.MultusAutoconfigDir, err)
 	}
 
-	masterConfigPath := files[0]
+	var masterConfigPath string
+	// skip existing multus configuration file to avoid creating a situation
+	// where multus delegates to itself and breaks pod networking
+	multusRegexp, err := regexp.Compile("multus")
```
s1061123, line 302: Currently thin_entrypoint creates 00-multus.conf or 00-multus.conflist. So could you please use "00-multus.conf{,list}" to match the multus generated file?

s1061123 (changes requested):

cmd/thin_entrypoint/main.go
thomasferrandiz: done, thanks for the input.

s1061123:
I locally tested simpe go code and your code still choose 00-multus.conf, hence your code does not fix this issue. Please check your code?

```
package main

import (
        "fmt"
        "path"
        "regexp"
)

func main() {

        filePathLists := []string{
                "/host/etc/cni/net.d/00-multus.conf",
                "/host/etc/cni/net.d/10-flannel.conflist",
        }

        multusRegexp, err := regexp.Compile("00-multus.conf{,list}")
        if err != nil {
                fmt.Printf("regexp compilation failed: %v", err)
                return
        }
        masterConfigPath := ""

        for _, filename := range filePathLists {
                if !multusRegexp.MatchString(path.Base(filename)) {
                        masterConfigPath = filename
                        break
                }
        }
        fmt.Printf("masterConfigPath: %s\n", masterConfigPath)
        return
}
```

```
$ go run test.go
masterConfigPath: /host/etc/cni/net.d/00-multus.conf
```

s1061123:
File another PR for this issue: https://github.com/k8snetworkplumbingwg/multus-cni/pull/1142 so let me close this PR.

thomasferrandiz:
The original PR without the modification you requested actually worked.

After testing, the correct regexp would be `regexp.Compile("00-multus.(conf|conflist)")`.

PR #1159, opened 2023-09-18
Add per-node-certification support

s1061123:
This change introduces per-node certification for multus pods. Once multus pod is launched, then specified bootstrap kubeconfig is used for initial access, then multus sends CSR request to kube API to get original certs for kube API access. Once it is accepted then the multus pod uses generated certs for kube access.

This PR only targets the code and doc/installer manifests will be addressed in another PR.

### DIFF
--- a/pkg/k8sclient/kubeconfig.go
+++ b/pkg/k8sclient/kubeconfig.go
@@ -1,0 +1,219 @@
+// Copyright (c) 2023 Multus Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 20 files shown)


pkg/k8sclient/kubeconfig.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, err
+	}
+	if clientInfo == nil {
+		return nil, fmt.Errorf("failed to create in-cluster kube client")
+	}
+	return clientInfo, err
+}
+
+// GetK8sClient gets client info from kubeconfig
+func GetK8sClient(kubeconfig string, kubeClient *ClientInfo) (*ClientInfo, error) {
```
dougbtv, line 154: Ok, I get it, the GetK8sClient moves here, makes sense.

dougbtv (approved):
LGTM! I think we can move forward with this as-is. But let's follow up with example yaml for deployment as well as docs (I can also follow up on this)

PR #1161, opened 2023-09-21
Move chroot from multus main process to its child processes

pliurh:
We used to run chroot in multus main process when calling other CNI plugin binary. We also use a mutex to lock the access to pod files. But this causes performance issues when facing heavy CNI_ADD/CNI_DEL requests.

With this patch, we do chroot in the child processes instead. So file operations in the main process will not be affected by chroot.

This change requires the multus thick plugin pod to mount CNI bin directory to the same path in the container host.

fixes https://issues.redhat.com/browse/OCPBUGS-18995

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -167,6 +167,11 @@ spec:
           volumeMounts:
             - name: cni
               mountPath: /host/etc/cni/net.d
+            # multus-daemon expects that cni-bin path must be identical between pod and container host.
+            # e.g. if the cni bin is in '/opt/cni/bin' on the container host side, then it should be mount to '/opt/cni/bin' in multus-daemon,
+            # not to any other directory, like '/opt/bin' or '/usr/bin'.
+            - name: cni-bin
+              mountPath: /opt/cni/bin
             - name: host-run
               mountPath: /host/run
             - name: host-var-lib-cni-multus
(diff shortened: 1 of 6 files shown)


s1061123:
Thank you so much for great PR, @pliurh ! Overall desgin makes sense to me.

One minor comments but this PR implicitly requires to manifest to mount CNI bin directory to same path in container host. (e.g. if /opt/cni/bin is the CNI bin directory, multus-daemon pod has to mount it with same path /opt/cni/bin, not /opt/bin.
Could you write down in [thick-daemon manifest file](https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/multus-daemonset-thick.yml) https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/multus-daemonset-thick.yml?

deployments/multus-daemonset-thick.yml
```diff
@@ -167,6 +167,8 @@ spec:
           volumeMounts:
             - name: cni
               mountPath: /host/etc/cni/net.d
+            - name: cni-bin
```
s1061123, line 170: Could you please add explicitly mentioned in comment? as

```
# multus-daemon expects that cni-bin path must be identical between pod and container host, 
# e.g. if cni bin is '/opt/cni/bin' in container host side, then it should be mount to '/opt/cni/bin' in multus-daemon,
# not other directory, as '/opt/bin' or '/usr/bin'.
```

s1061123 (changes requested):

Deadpan110:
line 173 of deployments/multus-daemonset-thick.yml should be:

` - name: cnibin`
and not:
` - name: cni-bin`

PR #1171, opened 2023-10-03
Deployments: Add watch permission to thick example

J0n4t4n:
The ClusterRole was missing the watch permission on pods, which resulted in Multus throwing this error message every few seconds:

```
Failed to watch *v1.Pod: unknown (get pods)
```

I am unsure if this permission is also needed for the other 2 example yamls, so I've just added it to the one I tested for the moment.

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -71,6 +71,7 @@ rules:
       - get
       - list
       - update
+      - watch
   - apiGroups:
       - ""
       - events.k8s.io


sdwilsh:
Based on 50c0357467bbb68530cd77161b0530d5f492f223, I'm guessing you might also need to update `e2e/templates/multus-daemonset-thick.yml.j2`, @J0n4t4n.  Adding `list`, which is in `main` right now, and `watch` from this PR, I was able to make this log line go away from the `kube-multus-ds` pods on `v4.0.2`:
```
W1104 17:49:40.828687  188635 reflector.go:533] k8s.io/client-go/informers/factory.go:150: failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:kubevirt:multus" cannot list resource "pods" in API group "" at the cluster scope
```

maiqueb (changes requested):
Good catch. 

Please do what @sdwilsh is saying in https://github.com/k8snetworkplumbingwg/multus-cni/pull/1171#issuecomment-1793509429 and also add the watch permission to the [e2e test template manifest](https://github.com/k8snetworkplumbingwg/multus-cni/blob/e2e8cfb677e8cf5352f737b5004effed7518d71a/e2e/templates/multus-daemonset-thick.yml.j2#L44).

s1061123:
Thank you for the PR, @J0n4t4n ! Merged.

Issue #1160, opened 2023-09-19
Multus daemon gets oomkilled because of too high oom_score


kpoos:
**What happend**:
Multus gets oomkilled.
[Mon Sep 18 19:06:01 2023] Memory cgroup out of memory: Killed process 3947048 (multus-daemon) total-vm:749764kB, anon-rss:39996kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:248kB oom_score_adj:999
Pods qosClass is not Guaranteed, but Burstable
status:
  qosClass: Burstable

that is why we have too high oom_score
cat /proc/4007973/oom_score
1332

 

oom scrore 1332 == 133% memory!!

It is because the init container does not have resource limit set.
The qosClass parameter cannot be changed in runtime, and when initcontanier finishes the run, it remains on Burstable even if the normal container has limit set. Setting up limits in the initcontainer solves the issue, and sets the pod qosStatus to Guaranteed

When multus gets killed, other pods are not able to start, as multus daemon is not running on the node, and these pods will be stucked forever in the phase when they try to get the containernetwork.  they can be deleted later on, but that needs manual intervention, or an ugly cronjob, better would be, if multus did not get oomkilled. 

With Guaranteed the oomscore is nice and low as well:
cat /proc/957339/oom_score_adj
-997
cat /proc/957339/oom_score
2

**What you expected to happen**:
Network component should be killed last, and the qosClass should be Guaranteed

The yaml manifest should contain the initcontainer resource limits as well by default to eliminate the possibility of oomkilling in the first row.

**How to reproduce it (as minimally and precisely as possible)**:
Remove limits from initcontainer

**Anything else we need to know?**:

**Environment**:

- Multus version : 4.0.2
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`): 1.26.8
- Primary CNI for Kubernetes cluster: canal
- OS (e.g. from /etc/os-release): ubuntu  20.04.6 LTS
- File of '/etc/cni/net.d/'   - not relevant
- File of '/etc/cni/multus/net.d'    - not relevant
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)   - not relevant
- Target pod yaml info (with annotation, use `kubectl get pod <podname> -o yaml`)  - not relevant 
- Other log outputs (if you use multus logging)

nicklesimba:
Hi, if you have the changes you described, can you make a pull request with your contribution? Thank you!

[...]

PR #1174, opened 2023-10-08
Add missing resource limits  to have pod being guaranteed qosclass

kpoos:
Changes for github issue #1160 
https://github.com/k8snetworkplumbingwg/multus-cni/issues/1160


dougbtv:
Hey @xujihui1985 -- appreciate the contribution, but, since this deployment is meant to be an example, I'm afraid that this might wind up killing Multus on sample clusters. If you have this need, we'd recommend that you implement it in your own deployment YAML, and we'd rather not make these limitations for an example.

PR #1180, opened 2023-10-24
Refactors the configuration options document reference

dougbtv:
There's some backlog in here that needs to be addressed. 

This could probably use another pass in order to address the configmap as used in the thick plugin as well. However, at least this sets the stage for that change.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,7 +1,16 @@
-## Multus-cni Configuration Reference
+# Multus-cni Configuration Reference
+
+## Introduction
+
+Aside from setting options for Multus, one of the goals of configuration is to set the configuration for your *default network*. The default network is also sometimes referred as the "primary CNI plugin", the "primary network", or a "default CNI plugin" and is the CNI plugin that is used to implement [the Kubernetes networking model](https://kubernetes.io/docs/concepts/services-networking/#the-kubernetes-network-model) in your cluster. Common examples include Flannel, Weave, Calico, Cillium, and OVN-Kubernetes, among others.
+
+Here we will refer to this as your default CNI plugin or default network.
+
+## Example configuration
 
 Following is the example of multus config file, in `/etc/cni/net.d/`.
-(`"Note1"` and `"Note2"` are just comments, so you can remove them at your configuration)
+
+Example configuration using `clusterNetwork` (see also [using delegates](#using-delegates))
 
 ```
 {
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
+Using the `clusterNetwork` option and the `delegates` are mutually exclusive. If `clusterNetwork` is set, the `delegates` field is ignored. Therefore:
+
+* Set `clusterNetwork` and if this is set, optionally set the `defaultNetworks`.
+* OR you must set `delegates`.
+
+Options:
+
+* `clusterNetwork` (string, required if not using `delegates`): default CNI network for pod-to-pod connectivity used in kubernetes cluster (Pod IP and so on). The following values are valid: name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file.
+* `defaultNetworks` ([]string, optional): Additional / secondary network attachment that is always attached to each pod. The following values are valid: name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file
+
```
s1061123, line 74: Good description. Could you add one more to mention that the net-attach-def's namespace? (i.e. multusNamespace specifies the net-attach-def's namespace for clusterNetwork and defaultNetwork)

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
-* `readinessindicatorfile`: The path to a file whose existence denotes that the default network is ready
-
-User should chose following parameters combination (`clusterNetwork`+`defaultNetworks` or `delegates`):
-
-* `clusterNetwork` (string, required): default CNI network for pods, used in kubernetes cluster (Pod IP and so on): name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file
-* `defaultNetworks` ([]string, required): default CNI network attachment: name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file
+* [`readinessindicatorfile`](#Default-Network-Readiness-Indicator): The path to a file whose existence denotes that the default network is ready
+message to next when some missing error. Defaults to false.
 * `systemNamespaces` ([]string, optional): list of namespaces for Kubernetes system (namespaces listed here will not have `defaultNetworks` added)
 * `multusNamespace` (string, optional): namespace for `clusterNetwork`/`defaultNetworks`
```
s1061123, line 60: We may also add default value is 'kube-system'

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
+### Using clusterNetwork
+
+Using the `clusterNetwork` option and the `delegates` are mutually exclusive. If `clusterNetwork` is set, the `delegates` field is ignored. Therefore:
+
+* Set `clusterNetwork` and if this is set, optionally set the `defaultNetworks`.
+* OR you must set `delegates`.
+
+Options:
+
+* `clusterNetwork` (string, required if not using `delegates`): default CNI network for pod-to-pod connectivity used in kubernetes cluster (Pod IP and so on). The following values are valid: name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file.
```
s1061123, line 72: "CNI json file name (without extension, .conf/.conflist)" might be 'name' field in CNI config file (not file name). 

Note: It also applies to defaultNetwork.

s1061123 (changes requested):

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
+
+If for example you have `defaultNetworks` set as:
+
+```
+"defaultNetworks": ["sidecarNetwork", "exampleNetwork"],
+```
+
+This expression refers to `NetworkAttachmentDefinition` custom resource names. Optionally, you could specify paths to CNI configurations. Therefore, there must be `NetworkAttachmentDefinitions` already created with the names `sidecarNetwork` and `exampleNetwork`.
+
+This means that in addition to the cluster network, each pod would be assigned two additional networks by default, and the pod would present three interfaces, e.g. `eth0`, `net1`, and `net2`, with `net1` and `net2` being set by the above described `NetworkAttachmentDefinitions`.
```
maiqueb, line 83: But if we define a pod's network selection elements, these are not applied, right ? i.e. this is the default when network selection elements are **not provided**.

(just wanted to clarify, since I wasn't aware of this feature 0_o ) The text is OK as is IMO.

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
+
+### Using delegates
+
+If `clusterNetwork` is not set, you must use `delegates`.
```
maiqueb, line 87: nit

```suggestion
If `clusterNetwork` is not set, you **must** use `delegates`.
```

maiqueb (commented):
Looks good. 

I mostly just oppose the "pod-to-pod" term.

docs/configuration.md
dougbtv: Thanks, I re-read the logic and added a new section that has details of how to use this value

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
 
-1. CRD object for given network name, in 'kube-system' namespace
-1. CNI json config file in `confDir`. Given name should be without extension, like .conf/.conflist. (e.g. "test" for "test.conf"). The given name for `clusterNetwork` should match the value for `name` key in the config file (e.g. `"name": "test"` in "test.conf" when `"clusterNetwork": "test"`)
-1. Directory for CNI json config file. Multus will find alphabetically first file for the network
-1. File path for CNI json confile file.
-1. Multus failed to find network. Multus raise error message
+1. `NetworkAttachmentDefinition` custom resource object for a given network name, in the namespace as specified by the `multusNamespace` configuration option.
+2. CNI json config file as found in the directory specified by the `confDir` option. The value should be without extension, e.g. .conf/.conflist. For example if you have a file named `test.conf`, you would specify the basename  `test`. The given name for `clusterNetwork` should match the value for `name` key in the config file (e.g. `"name": "test"` in "test.conf" when `"clusterNetwork": "test"`)
```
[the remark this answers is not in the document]
dougbtv: have a new section to address this

docs/configuration.md
```diff
@@ -1,7 +1,14 @@
-## Multus-cni Configuration Reference
+# Multus-cni Configuration Reference
+
+## Introduction
+
+Aside from setting options for Multus, one of the goals of configuration is to set the configuration for your *default network*. The default network is also sometimes referred as the "primary CNI plugin" or "primary network" and is the CNI plugin that is used to provide pod-to-pod connectivity in your cluster. Common examples include Flannel, Weave, Calico, Cillium, and OVN-Kubernetes, among others.
```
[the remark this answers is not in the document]
dougbtv: Fair enough, it is unspecific enough here, and maybe is confusing. Thanks for the link, that's a way better way to say it.

docs/configuration.md
```diff
@@ -23,65 +30,102 @@ Following is the example of multus config file, in `/etc/cni/net.d/`.
...
-
-User should chose following parameters combination (`clusterNetwork`+`defaultNetworks` or `delegates`):
-
-* `clusterNetwork` (string, required): default CNI network for pods, used in kubernetes cluster (Pod IP and so on): name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file
-* `defaultNetworks` ([]string, required): default CNI network attachment: name of network-attachment-definition, CNI json file name (without extension, .conf/.conflist), directory for CNI config file or absolute file path for CNI config file
+* [`readinessindicatorfile`](#Default-Network-Readiness-Indicator): The path to a file whose existence denotes that the default network is ready
+message to next when some missing error. Defaults to false.
 * `systemNamespaces` ([]string, optional): list of namespaces for Kubernetes system (namespaces listed here will not have `defaultNetworks` added)
 * `multusNamespace` (string, optional): namespace for `clusterNetwork`/`defaultNetworks`
-* `delegates` ([]map,required): number of delegate details in the Multus
```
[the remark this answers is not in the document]
dougbtv: I actually just moved it to its own section called `Using delegates`

docs/configuration.md
dougbtv: I am going to have to look this up / test it :thinking: I didn't realize this either

dougbtv: actually I think it's always attached, and always in addition to the items in the selection element: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/k8sclient/k8sclient.go#L524-L565

maiqueb: All good; I was simply unaware of this.

Thanks for documenting it.

s1061123 (approved):
LGTM, merged.

PR #1189, opened 2023-11-20
fix e2e test ModuleNotFoundError: No module named 'pkg_resources'

moshe010:


maiqueb (approved):

moshe010:
@dougbtv can you please review?

s1061123:
Thank you for the PR and fix CI! I prefer to use j2cli package without replacing package with others.
As far as I test, this change works too. Could you try that?

```
diff --git a/.github/workflows/kind-e2e.yml b/.github/workflows/kind-e2e.yml
index 165070b6..585104e3 100644
--- a/.github/workflows/kind-e2e.yml
+++ b/.github/workflows/kind-e2e.yml
@@ -45,8 +45,8 @@ jobs:

       - name: Setup j2cli
         run: |
-          pip3 install --user --upgrade j2cli
-          j2 --version
+          sudo apt-get install -y j2cli
+          echo $(j2 --version)
```

s1061123:
I'll change that and merge soon.

PR #1190, opened 2023-11-26
Align the message with master plugin restoration logic

kichankwon:
thin_entrypoint prints that it allows 45 seconds for its restoration, but it waits until "masterConfigFilePath" exists.
So, adjust the message to align with this logic.

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -587,7 +587,7 @@ func main() {
 
 			// if masterConfigFilePath is no longer exists
 			if os.IsNotExist(err) {
-				fmt.Printf("Master plugin @ %q has been deleted. Allowing 45 seconds for its restoration...\n", masterConfigFilePath)
+				fmt.Printf("Master plugin @ %q has been deleted. Wait until the master configuration is found...\n", masterConfigFilePath)
 				time.Sleep(10 * time.Second)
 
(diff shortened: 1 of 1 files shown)


s1061123:
Thank you for the PR. Good catch that mismatch between comment and actual code. But I would like to know why you fix to 45sec wait, not wait forever. If we cannot find master config, then your change makes the pod quit and launch it again...

How about to change to wait forever until the master config found?

kichankwon:
> Thank you for the PR. Good catch that mismatch between comment and actual code. But I would like to know why you fix to 45sec wait, not wait forever. If we cannot find master config, then your change makes the pod quit and launch it again...
> 
> How about to change to wait forever until the master config found?

I fixed to 45 seconds wait because..
- I thought there's a mistake in the message or logic
- And I found that old shell script(images/entrypoint.sh) waits 45 seconds just before it got removed
  - https://github.com/k8snetworkplumbingwg/multus-cni/blob/c129e72779ce2d97f848f5409f6b289846e4b7d8/images/entrypoint.sh#L457
- So, I decided to fix the logic

But, I agree with you. So, I'm going to fix the message not the logic.

kichankwon:
@s1061123 Could you check this PR?

kichankwon:
@s1061123 Can I ping you for a review?

PR #1199, opened 2023-12-21
feat: propagate the err message from getNetDelegate

xujihui1985:
propagate the err message from getNetDelegate, and log the error in the error message, we don't want to lose the err here


xujihui1985:
@s1061123 Hi, would you like to have a look at this PR?

xujihui1985:
ping @micw @e0ne @tedyu

dougbtv (approved):

dougbtv:
Thanks for your contribution!

dougbtv:
Hey @xujihui1985 -- appreciate the contribution, but, since this deployment is meant to be an example, I'm afraid that this might wind up killing Multus on sample clusters. If you have this need, we'd recommend that you implement it in your own deployment YAML, and we'd rather not make these limitations for an example.

xujihui1985:
@dougbtv Hi thanks for the reply, but sorry I don't get your point, what do you mean by an example? And how can I able to access the inner error if I don't propagate the err? What can I do with yaml?

dougbtv:
@xujihui1985 sorry! I had made a comment by accident on the wrong pull request!

xujihui1985:
Hi, would anyone like to merge this PR? the failing test should have no relation with this PR

PR #1203, opened 2024-01-05
Reduce informer memory usage by informer transform

s1061123:
This fix reduces multus-daemon memory usage with k8s 0.29 informer transform to trim unnecessary Pod object information to multus.


dougbtv (approved):
LGTM!

PR #1204, opened 2024-01-08
pass through the pod annotations when multus receives them

toelke:
See https://github.com/kubernetes/kubernetes/issues/69882#issuecomment-1509077177 and https://github.com/k8snetworkplumbingwg/multus-cni/pull/1116

Passing the annotations seems not to be a standardized behaviour, but it is very useful nontheless.

### DIFF
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -268,6 +268,9 @@ func newCNIRuntimeConf(containerID, sandboxID, podName, podNamespace, podUID, ne
 		if delegateRc.CNIDeviceInfoFile != "" {
 			capabilityArgs["CNIDeviceInfoFile"] = delegateRc.CNIDeviceInfoFile
 		}
+		if delegateRc.PodAnnotations != nil {
+			capabilityArgs["io.kubernetes.cri.pod-annotations"] = delegateRc.PodAnnotations
+		}
 		rt.CapabilityArgs = capabilityArgs
 	}
 	return rt, cniDeviceInfoFile
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -69,13 +69,14 @@ type NetConf struct {
 
 // RuntimeConfig specifies CNI RuntimeConfig
 type RuntimeConfig struct {
-	PortMaps          []*PortMapEntry `json:"portMappings,omitempty"`
-	Bandwidth         *BandwidthEntry `json:"bandwidth,omitempty"`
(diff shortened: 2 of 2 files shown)


s1061123:
I'm not clear that `io.kubernetes.cri.pod-annotations` comes from and use-case yet. Could you please clarify that, including why the annotation, `io.kubernetes.cri.pod-annotations`, should be default (to all platform).

toelke:
[This comment](https://github.com/kubernetes/kubernetes/issues/69882#issuecomment-1509077177) is a good summary.

My use case is that a custom CNI plugin needs some extra information to configure the networking of a Pod that I would like to pass as an annotation on the Pod; containerd can already be configured to pass this annotation to CNI plugins, this patch enables multus to receive it.

dougbtv:
Tomo's looking into the possibility of checking this out with capability args in the meanwhile, and we'll circle back to it to make sure it's necessary.

aojea:
> Tomo's looking into the possibility of checking this out with capability args in the meanwhile, and we'll circle back to it to make sure it's necessary.

@dougbtv @s1061123 I think that this refers to the fact that you don't have to go to the apiserver to fetch the labels and the annotations of the pod if you set this option in containerd and crio https://github.com/kubernetes/kubernetes/issues/69882#issuecomment-1701125419, as those values will be available in the CNI ADD request
@mikezappa87

MikeZappa87:
I didn't think we passed all the annotations/labels from the container runtime (containerd) to the CNI ADD? This would be in sandbox_run.go. I can get the link when I get home.

AJMansfield:
This would be a useful feature for my own use-case -- to pass node affinity/taint information concerning particular network attachments onto pods that use them.

bgaussen:
Hi, this feature would help our CNI to get some pod annotations directly and avoid implementing token management for it to access api-server...

toelke:
I still think this is useful.

toelke:
I still still think this is useful.

toelke:
I still still think this is useful.

toelke:
I still still think this is useful.
/remove-lifecycle stale

toelke:
I still still think this is useful.
/remove-lifecycle stale

toelke:
I still still think this is useful.
/remove-lifecycle stale

toelke:
I still still think this is useful.
/remove-lifecycle stale

PR #1208, opened 2024-01-11
Add watch permission to thick e2e template

LionelJouin:
As described in #1171, the watch function is required in the clusterrole for the thick Multus version, otherwise "Failed to watch *v1.Pod" would be returned.

This permission is still missing in `e2e/templates/multus-daemonset-thick.yml.j2`


s1061123:
LGTM, merged.

PR #1210, opened 2024-01-15
The code repository is not cloned and the configuration is delivered directly

Fish-pro:
The code repository is not cloned and the configuration is delivered directly


s1061123:
LGTM, merged.

PR #1213, opened 2024-01-19
update deploy file: use install_multus bin to update cni file

41tair:
install cni initcontiainer can't execute sometimes 

a sample error logs looks like:

cp: cannot create regular file '/host/opt/cni/bin/multus-shim': Text file busy

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -199,9 +199,11 @@ spec:
         - name: install-multus-binary
           image: ghcr.io/k8snetworkplumbingwg/multus-cni:snapshot-thick
           command:
-            - "cp"
-            - "/usr/src/multus-cni/bin/multus-shim"
-            - "/host/opt/cni/bin/multus-shim"
+            - "/usr/src/multus-cni/bin/install_multus"
+            - "-d"
+            - "/host/opt/cni/bin"
+            - "-t"
+            - "thick"
           resources:
             requests:
               cpu: "10m"


s1061123 (changes requested):
Hi @41tair, thank you for the PR!

Once we use `install_multus` for thick pluigin, then we could change container base image to distroless from 

https://github.com/k8snetworkplumbingwg/multus-cni/blob/ba18cf5ab3fbe7f748164e18bb12317de4412341/images/Dockerfile.thick#L10

to such like that as thin Dockerfile

https://github.com/k8snetworkplumbingwg/multus-cni/blob/ba18cf5ab3fbe7f748164e18bb12317de4412341/images/Dockerfile#L11

Could you pelase take care of that?

kfox1111:
Anything left to do on this? Think mode seems broken/risky without it.

dougbtv (approved):
LGTM! Thanks for the update

kfox1111:
Reluctant to deploy a new instance of multus until this bug is fixed.

dustinrouillard:
> Reluctant to deploy a new instance of multus until this bug is fixed.

Would love to see this fixed.

e2e/templates/multus-daemonset-thick.yml.j2
```diff
@@ -168,9 +168,11 @@ spec:
       - name: install-multus-shim
         image: localhost:5000/multus:e2e
         command:
-          - "cp"
-          - "/usr/src/multus-cni/bin/multus-shim"
-          - "/host/opt/cni/bin/multus-shim"
+	        - "/usr/src/multus-cni/bin/install_multus"
+	        - "-d"
+          - "/host/opt/cni/bin"
```
kfox1111, line 173: The indenting here looks off.

hansbogert:
as an end-user  I'm wondering what's blocking this? I'm kinda stuck, 4.0.2 is non working for my cluster, but so is master, b/c of this: https://github.com/k8snetworkplumbingwg/multus-cni/issues/1221 

#1221  mentions that that issue should be fixed by this change

sandj1438:
@dougbtv  can we merge this and release a new image ? it is a critical bug fix right?

dustinrouillard:
Hey maintainers, can we please get some action on this?!?!?!?!?!?!?!?! People are tired of manually patching and fixing their installs.. Merge. The. Fix.

kub3let:
@s1061123 sorry to bother you but please give a statement about this, it's breaking a lot of clusters.

What has to be done to get this merged ?

kfox1111:
@dougbtv https://github.com/kubevirt/cluster-network-addons-operator/issues/1810 is the same issue too. They seem to just be following this repo's lead and running into the same issue. :/

dougbtv:
anyone willing to look into why there's a failing test? that'll help us move this along, thanks!

appreciate the frustration, but, we'd *love* to have the passing tests, and it'll be a "no question about it" to merge during the maintainer's call if there's both passing tests and an additional e2e test that validates this functionality, too.

dims:
@dougbtv looked into one of the logs and spotted this. Looks like `portmap` is mandating newer GLIBC which is not present in the centos:8 container. ` ubuntu:22.04` perhaps? as it is there in one of the other templates

```
Dec 05 15:00:10 kind-worker2 containerd[167]: 2024-12-05T15:00:10Z [error] CmdDel (shim): CNI request failed with status 400: 'ContainerID:"59a6abee1e17eac1d3f833c90fcd4ef107be97902c5c21d4d85ba262cc23b795" Netns:"/var/run/netns/cni-54886520-6d07-c76b-a9f0-4e093f2bcb5d" IfName:"eth0" Args:"K8S_POD_INFRA_CONTAINER_ID=59a6abee1e17eac1d3f833c90fcd4ef107be97902c5c21d4d85ba262cc23b795;K8S_POD_UID=c6fd5fae-26cc-437a-86b4-bf6075a38f2f;IgnoreUnknown=1;K8S_POD_NAMESPACE=default;K8S_POD_NAME=simple-centos1" Path:"" ERRORED: DelegateDel: error invoking ConflistDel - "kindnet": conflistDel: error in getting result from DelNetworkList: plugin type="portmap" failed (delete): netplugin failed: "/host/opt/cni/bin/portmap: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /host/opt/cni/bin/portmap)\n/host/opt/cni/bin/portmap: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /host/opt/cni/bin/portmap)\n"
```

evilhamsterman:
Since it looks like E2E is not an issue, and a lot of people have been manually patching the copy command for almost a year already so I think we can be pretty confident it doesn't have issues. Can we get this merged

BenTheElder:
I'm not that familiar with multus ... are you attempting to invoke the host's CNI from within your container? That's not portable, regardless of if we ship the kind change. Can you install your own CNI binaries?

Also: 
https://kind.sigs.k8s.io/docs/design/node-image/

> DO NOT DEPEND ON THE INTERNALS OF THE NODE IMAGES.

We only guarantee that these can create Kubernetes, please try not to depend on what is available on the host versus using conformant Kubernetes APIs ...

> Generally speaking, a regular K8s cluster, rather than a kind cluster, typically uses statically linked executable CNI binaries:

That seems like a rather fragile assumption. This workload probably needs to either:
a) fully install CNI itself
b) run on the host
c) stick to a very stable glibc track

kfox1111:
> I'm not that familiar with multus ... are you attempting to invoke the host's CNI from within your container? That's not portable, regardless of if we ship the kind change. Can you install your own CNI binaries?

the issue is simply about copying the file from the container onto the host so the host can use it. If its in use currently, it blocks updating. In thick mode, it tries to start and blocks until the service is up. But the service wont start until the init container starts to copy the driver onto the host. so it wont auto recover. Simply always overwriting it is one way to fix it.

BenTheElder:
Again, I'm not sure if this is relevant for multus (apologies), but many "cni" daemonset "installers" will run an init container to copy out the binaries they need to the host, which is a little more portable (because so many do this it's virtually required to not customize the usual predictable CNI host directory).

I merged @dims's PR because it makes sense anyhow, but we are probably going to delete unused binaries in the future as well, and may or may not break this setup in the future (certainly not intentionally out of malice, but this is not something we're actively trying to guarantee either, bring your own CNI is generally "best of luck, we don't have the bandwidth to test all of these").

😅

dougbtv:
Definitely bring this to the maintainer's call to have a real-time chat about it. Like I said, we'd love to discuss it.

There's two main reasons:

1. We want passing tests to merge a PR, if a PR breaks a test -- we need a fix for it. One way or another, we can't just break the rest of CI because this fixes what is currently an untested fix.
2. I don't think we test an upgrade situation as of right now. I think that's why people are hitting this issue, but we're not seeing this issue in CI.

The fastest path to a merge here is:

1. Have passing end-to-end tests.
2. Have replication of the issue in the end-to-end tests (possibly an upgrade test), and a fix for it.

That being said, this needs to be tackled from a few directions.... And I appreciate the help I've gotten there (thanks Dims for the build

1. We do already install our own copy of the reference CNI plugins, so, potentially the order by which we install these, we should check that.
2. I don't grok why this test changes and makes it so we hit that error, but we're not seeing it in other e2e tests? That is baffling to me. I'd love to know why.
3. I will try to run it with the updated kind, if that works and CI can pass, great.

The maintainer's stance is that the yaml as provided for a quick start and for end to end testing, and as a reference. We do advise maintaining your own yaml for production deployment and to customize to your systems.

kfox1111:
its not an upgrade issue, its a reboot race issue. it works the first time. driver is installed on the node with init container, and starts up fine.

Then, reboot the node:

Kubelet comes up. tries to start pods. cni is called, it runs the multus binary in client mode, tries to contact the multus server. it is not available yet, the cni driver **blocks** rather then returns like most cni drivers at this point..... The process is actively using the multus cni driver file at this point and has locked out writes to that file.

next the multus pod tries to start on the node. it runs the init container, it tries to copy the driver onto the host (its already there). The copy fails currently because it tries to overwrite the data inside the binary file rather then roll it into place. the process in memory that is already running blocks that from happening.

because the init container fails, it never starts the server, so the cni instance on the host can never start to get out of the way.

Deadlock/broken cluster.

evilhamsterman:
What @kfox1111 said this is a long running issue with regular operations not just upgrades. Take a look at the linked issue that this is supposed to fix. People do maintain their own deployment files and have been applying this exact patch for almost a year now. But most people don't figure out this issue until they go to reboot a node and hit the problem. Then they spend hours trying to figure it out while managers are breathing down their necks asking when it'll be fixed.

It's a simple change to basically do the same thing that the thin version does. It's been identified that the failing tests are due to the tests themselves not the change. If you want to skip the change they made to the docker file that's fine, I believe that was done due to a request in the issue.

But otherwise please just merge this. It's a simple bug with a simple fix that's causing a lot of headaches

evilhamsterman:
Also due to the fact it's an issue with a race condition by its nature will be difficult to replicate. But the patch resolves it

images/Dockerfile.thick
```diff
@@ -7,7 +7,7 @@ ADD . /usr/src/multus-cni
 RUN  cd /usr/src/multus-cni && \
      ./hack/build-go.sh
 
-FROM debian:stable-slim
+FROM gcr.io/distroless/base-debian11:latest
```
iiiceoo, line 10: > I don't grok why this test changes and makes it so we hit that error, but we're not seeing it in other e2e tests? That is baffling to me. I'd love to know why.

In the e2e test, multus-ds mounts the CNI binaries provided by the kind node. These binaries are dynamically linked (compiled with `CGO_ENABLED=1`). When these CNI binaries (e.g., `portmap`) are called, they rely on the GLIBC of multus-ds container, which is the GLIBC of base image `gcr.io/distroless/base-debian11:latest`.

Perhaps change to `debian:stable-slim` allows the e2e test to pass, but it does not meet @s1061123's suggestions.

The reason I've explained so much is simply to convey that the failure of this e2e test is solely an issue with the CI job, not a problem with the changes made in this PR.

lion7:
I just ran into this exact same issue. It took me a while to figure out why none of my pods would start. Please merge this fix.

kub3let:
Since this still has not been merged, to fix your cluster apply the patch manually for the time being via

```
kubectl patch daemonset kube-multus-ds -n kube-system --type='json' -p='[
  {
    "op": "replace",
    "path": "/spec/template/spec/initContainers/0/command",
    "value": [
      "/usr/src/multus-cni/bin/install_multus",
      "-d",
      "/host/opt/cni/bin",
      "-t",
      "thick"
    ]
  }
]'
```

Your cluster should come back up within a couple seconds and not fail again in the future as long as you don't redeploy multus

---

The command update from the commit is needed, however replacing the base distro image to debian11 which is outdated is not good IMHO, at least use debian12

evilhamsterman:
@dougbtv this is starting to get ridiculous. I can understand the importance of tests, but some times you also have to have some common sense.

This is an issue with the tests not the fix, all the fix is doing is making the thick version (the recommended version per the docs)  use the same command to install the multus binary as the thin version https://github.com/k8snetworkplumbingwg/multus-cni/blob/4a0b5073afa53b57eb910896e72891d6857c199b/deployments/multus-daemonset.yml#L208

The point of tests is to ensure functionality works as expected and prevent regressions. Multiple people have commented that we've been applying this patch to multiple clusters of different types to resolve the issue that is genuinely a serious **cluster breaking** problem in regular operations not just upgrades. I think that meets the needs of testing, we've proven it works, just because you don't have a green check mark on your automated tests doesn't mean the needs aren't being met.

Also despite your statements
> The maintainer's stance is that the yaml as provided for a quick start and for end to end testing, and as a reference. We do advise maintaining your own yaml for production deployment and to customize to your systems.

that is not reflected in the documentation. If you just read the documentation the canonical recommended way to install Multus is 

```
kubectl apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset-thick.yml
```

Nothing says do this to try and then copy and modify the yaml for your own purposes. The documentation implies you can just use the provided file, which aligns with many other Kubernetes applications that offer deployment files. If the intention is different then it is not stated. Because of the nature of the issue you may deploy Multus, have it working just fine, go months without issue and start relying on Multus, until you need to reboot your systems for some reason, say for instance you had a power outage. Everything else comes back up and then you hit this issue and now your entire cluster is broken and you don't know why. 

Then you dig and discover this issue has been known about for over a year, and it has a stupidly simple fix that the maintainers refuse to merge because their own tests are faulty. They admit they're faulty, but they still refuse to merge it because they can't fix their own tests. 

Now you discover that despite what the documentation says you have to make a copy of the deployment file, or use something like kustomize to apply the four line patch yourself. You do so and find it resolves all your issues, but now you also have to worry about keeping your copy or kustomize patch up to date with the source, and apply the patch. 

It is incredibly frustrating. We are pleading you please merge this, if you don't want to change the Dockerfile, that's fine we don't care, revert that line and merge the main fix. That was a change made after the initial PR as an aside. All we care about is the fix to the command so that others don't have to go through the same frustration we all have.

And if the expectation is that everyone should maintain their own copy of the deployment files then you need to update your documentation to reflect that. As it is now Multus is broken. If there were an alternative I would be switching at this point.

evilhamsterman:
@s1061123 @dougbtv Can you please for the love that all that is holy and good in this world fix this. At this point there's merge conflicts, which is why we didn't want to have to maintain our own copies of the file. But the issue still remains. Yet another person was caught by it. The patch has been applied by admins for almost two years now. I'm sorry that your TESTING environment testing is broken, but our PRODUCTION environments work fine.

It's been tested in PRODUCTION, it's a simple copy operation, there is little to no chance of side affects that could cause issues. I'll remind you again the goal of TESTING is to ensure that the changes won't affect PRODUCTION. It's working in PRODUCTION so the test is complete, yet the bug still exists and is causing issues in PRODUCTION.

I'll update my PR to add some retries or something when I get the chance, probable just changing it to `cp -f` may be all that's needed.

Untersander:
As this is blocked because of the changes to the base for the Dockerfile which break the e2e tests consider merging my new PR #1445  which only fixes the node reboot race condition issue and little flakiness i discovered in the tests.

casell:
This issue has been affecting me on almost every node reboot in PRODUCTION, and honestly I completely agree with @evilhamsterman.
I have applied ONLY this change: https://github.com/k8snetworkplumbingwg/multus-cni/pull/1445 (switched from CP to INSTALL_MULTUS) and rebooting went flawlessly.

evilhamsterman:
@kfox1111 they merged the original PR #1445 instead, this was a version that removed a change in the original PR that caused some issues with the tests due to a change in the docker image used, I omitted it that change.

The important change about using `instaull_multus` is now in the main repo https://github.com/k8snetworkplumbingwg/multus-cni/blob/f42e0bd8fa436df5e2b60b0d69cd35ac85a78005/deployments/multus-daemonset-thick.yml#L206

kfox1111:
oh, thats great news. thanks for the update.

PR #1214, opened 2024-01-19
Add net-attach-def informer for thick plugin

s1061123:
This change introduces net-attach-def informer in multus-daemon, thick pluign case. It could reduced API calls to get net-attach-def.


dougbtv (approved):
LGTM!

PR #1216, opened 2024-01-25
Update github actions

s1061123:


s1061123:
Merged because no CI failures (and it is trivial).

PR #1219, opened 2024-02-01
Opt out to mount service account token

s1061123:


s1061123:
Will be merged because it does not change actual multus code, just example yamls.

PR #1228, opened 2024-02-13
Reload bootstrap kubeconfig if cert mgr failed to load valid certs

s1061123:
When user recreate whole cluster certs, multus thick plugin's previous cert is no longer valid. In such case, we need to prevent to use cert manager's old certs and restart it from bootstrap kubeconfig. This fix reloads client config from bootstrap kubeconfig if cert mgr's cert is failed to load pod.

### DIFF
--- a/pkg/k8sclient/kubeconfig.go
+++ b/pkg/k8sclient/kubeconfig.go
@@ -22,10 +22,12 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	certificatesv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


pkg/k8sclient/kubeconfig.go
```diff
@@ -87,8 +89,23 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+		if err != nil {
+			logging.Errorf("failed to read kubeconfig from cert manager: %v", err)
+		} else {
+			_, err := tempClient.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+			// tls unknown authority error is unrecoverable error with retry
+			if err != nil && strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
+				// reload and use bootstrapKubeconfig again
+				newBootstrapKubeconfig, _ := clientcmd.BuildConfigFromFlags("", bootstrapKubeconfigFile)
+				cfg = newBootstrapKubeconfig
+			}
```
dougbtv, line 104: Maybe we should log a warning when it matches the string, and then also log an error when it doesn't match, e.g. 

`err != nil && strings.Contains(err.Error(), "x509: certificate signed by unknown authority")` == log warning
else if `err !=nil` then log the error

pkg/k8sclient/kubeconfig.go
```diff
@@ -87,8 +89,24 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+
+		// validate the kubeconfig
+		tempClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			logging.Errorf("failed to read kubeconfig from cert manager: %v", err)
+		} else {
+			_, err := tempClient.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
+			// tls unknown authority error is unrecoverable error with retry
+			if err != nil && strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
+				logging.Verbose("cert mgr gets invalid config. rebuild from bootstrap kubeconfig")
```
dougbtv, line 101: Thanks, that'll do.

dougbtv (approved):

pkg/k8sclient/kubeconfig.go
s1061123: changed.

s1061123: FYI: https://github.com/googleapis/gax-go/blob/main/v2/invoke.go#L86-L94

dougbtv (approved):
Nice Tomo, thanks!

PR #1229, opened 2024-02-14
Add filepath sanity check

s1061123:
Uses `filepath.Abs()` to check that paths passed  in are truly absolute paths, and not garbage data (for example).


dougbtv (approved):
LGTM!

PR #1231, opened 2024-02-15
Minimize temp file creations for createKubeConfig and createMultusConfig

connorkuehl:
The thin plugin invokes this function in a tight loop when using --cleanup-config-on-exit and --multus-conf-file=auto.
    
Instead of rendering directly to the temp file, render to an in-memory buffer which can simply be discarded if there's no need to replace the working copy of the file.


connorkuehl:
Looks like a different PR to accomplish this was merged https://github.com/k8snetworkplumbingwg/multus-cni/pull/1234

Closing.

PR #1232, opened 2024-02-15

connorkuehl:

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -57,6 +57,7 @@ type Options struct {
 	AdditionalBinDir         string
 	ForceCNIVersion          bool
 	SkipTLSVerify            bool
+	WatchTimer               time.Duration
 }
 
 const (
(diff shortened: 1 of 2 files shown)


s1061123:
Thank you for the PR, @connorkuehl 

Currently multus thin_entrypoint assumes that Kubeconfig is not updated by deployment because this modification might be reverted again when cert is rotated. Hence thin_entrypoint supports only 'generated Kubeconfig without mody out-of-band' and thin_entrypoint does not support 'Kubeconfig with modify out-of-band'.

connorkuehl:
@s1061123 I dropped the kubeconfig hashing patch. Would you be willing to consider the remaining patch for controlling how quickly the loop spins?

dougbtv:
can we keep the default second wait? if we'd like to we can update the sample daemonset to include the minute long wait

also let's update the docs/how-to-use.md and/or docs/configuration.md to reflect the option as well

connorkuehl:
> can we keep the default second wait? if we'd like to we can update the sample daemonset to include the minute long wait
> 
> also let's update the docs/how-to-use.md and/or docs/configuration.md to reflect the option as well

Sure thing! I've pushed a new revision setting the default value to 1 second and also added to the how-to-use.md. Thanks for the review :slightly_smiling_face:

dougbtv (approved):
LGTM! Appreciate the contribution.

connorkuehl:
Thanks, @dougbtv! It looks like the CI failure is unrelated to the PR. Would you mind retriggering it?

connorkuehl:
@dougbtv @s1061123 just checking to see if either of you had any other feedback on the patch

mingshuoqiu:
@dougbtv @s1061123 gentle ping. Do you still have any concern to merge the PR?

mingshuoqiu:
@dougbtv @s1061123 gentle ping. Do I need anyone else to review to speed up the merge progress? Thanks

PR #1234, opened 2024-02-19
Recreate configs only if base files changed (thin_entrypoint)

pmtk:
Changes how new kubeconfig is created after CA/token change. Instead of always creating new file on disk and comparing it to existing, it will compare base files (`ca.crt` and `token`) and recreate kubeconfig only if they changed.
Multus CNI config is now treated the same way - as base file the master cni config is used.

As a result `thin_entrypoint` will no longer create new kubeconfig and 00-multus on each iteration of the watch loop and only recreate them when underlying base files change.

This approach was chosen over using a fsnotify because it wasn't reliable watching `/var/run/secrets/kubernetes.io/serviceaccount/` - it would just stop noticing changes to files, possible reason is underlying filesystem being `tmpfs`.

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -21,7 +21,6 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


s1061123:
Looks good to me. Merged. Thanks @pmtk

PR #1235, opened 2024-02-20
Skips checking for readiness on CNI DEL

dougbtv:
Because deletes should favor a successful path, the readiness check should be skipped for pod removals.

This can cause an issue where there's pods pending deletes and that might impact scheduling of a pod that may be necessary in order to set the readiness indicator.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -815,8 +815,13 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 	}
 
 	if in.ReadinessIndicatorFile != "" {
-		if err := types.GetReadinessIndicatorFile(in.ReadinessIndicatorFile); err != nil {
-			return cmdErr(k8sArgs, "PollImmediate error waiting for ReadinessIndicatorFile (on del): %v", err)
+		readinessfileexists, err := types.ReadinessIndicatorExistsNow(in.ReadinessIndicatorFile)
+		if err != nil {
+			return cmdErr(k8sArgs, "error checking readinessindicatorfile on CNI DEL @ %v: %v", in.ReadinessIndicatorFile, err)
+		}
+
+		if !readinessfileexists {
+			logging.Verbosef("warning: readinessindicatorfile @ %v does not exist on CNI DEL", in.ReadinessIndicatorFile)
 		}
 	}
 
--- a/pkg/types/conf.go
+++ b/pkg/types/conf.go
@@ -20,10 +20,9 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
-
-	utilwait "k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/containernetworking/cni/libcni"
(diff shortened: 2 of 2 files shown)


pkg/multus/multus.go
```diff
@@ -814,12 +814,6 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 		return cmdErr(nil, "error getting k8s args: %v", err)
 	}
 
-	if in.ReadinessIndicatorFile != "" {
```
s1061123, line 817: Could you please add warning message (as Verbosef)?

s1061123 (changes requested):

pkg/multus/multus.go
```diff
@@ -815,8 +815,13 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 	}
 
 	if in.ReadinessIndicatorFile != "" {
-		if err := types.GetReadinessIndicatorFile(in.ReadinessIndicatorFile); err != nil {
-			return cmdErr(k8sArgs, "PollImmediate error waiting for ReadinessIndicatorFile (on del): %v", err)
+		readinessfileexists, err := types.ReadinessIndicatorExistsNow(in.ReadinessIndicatorFile)
+		if err != nil {
+			return cmdErr(k8sArgs, "error checking readinessindicatorfile on CNI DEL @ %v: %v", in.ReadinessIndicatorFile, err)
+		} else {
```
s1061123, line 821: As of test warning message in https://github.com/k8snetworkplumbingwg/multus-cni/pull/1235/files#annotation_18379130107 you can remove else block (because if clause ends with return, hence we don't need else block.

pkg/types/conf.go
```diff
@@ -610,11 +610,32 @@ func CheckSystemNamespaces(namespace string, systemNamespaces []string) bool {
...
+
 	pollDuration := 1000 * time.Millisecond
 	pollTimeout := 45 * time.Second
 	return utilwait.PollImmediate(pollDuration, pollTimeout, func() (bool, error) {
 		_, err := os.Stat(readinessIndicatorFile)
 		return err == nil, nil
 	})
 }
+
+func ReadinessIndicatorExistsNow(readinessIndicatorFileRaw string) (bool, error) {
```
s1061123, line 627: > exported function ReadinessIndicatorExistsNow should have comment or be unexported
Please take care of github action's warning message?

pkg/types/conf.go
```diff
@@ -610,11 +610,32 @@
...
+	readinessIndicatorFile, err := filepath.Abs(readinessIndicatorFileRaw)
+	if err != nil {
+		return false, fmt.Errorf("failed to get absolute path of readinessIndicatorFile: %v", err)
+	}
+
+	_, err = os.Stat(readinessIndicatorFile)
```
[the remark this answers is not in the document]
dougbtv: I tried to read up on this, and I think it wants you to limit it to a subset of directories given a  "base path", but, that doesn't work  here, and it's just a presence indicator, and not a destructive action, so it should be fine.

s1061123 (approved):
Looks good to me. So let's check e2e tests and merged.

PR #1243, opened 2024-03-14
Change Validation of interface name

adrianchiris:
interface name should not be limited to DNS-1123 label format. instead validate interface name if provided in pod network annotation in a similar manner as iproute2[1].

this will allow to request interface names such as: "uplink_p0"

[1]https://github.com/iproute2/iproute2/blob/11740815bfe69d6ee2cad7c608a8edc70147209a/lib/utils.c#L832

### DIFF
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"syscall"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


dougbtv:
Hey Adrian -- this one is bigger than a breadbox :) Would you mind joining the NPWG or Multus maintainer's call to chat us about it?

I think there's some subtlety here between "attachment name" and "interface name" and the nuance there in the spec as well, we should chat it out. Thanks!

adrianchiris:
sure, will raise it in one of the meetings. ATM on PTO (paternity leave) so will bring it up once im back. (~3 weeks )

dougbtv:
I think my readthrough was wrong, and that this is totally reasonable given teh spec:

> 4.1.2.1.5 "interface"
> This optional key with value of type string requires the implementation to use the given name for the pod interface resulting from this network attachment. This key’s value must be a valid Linux kernel interface name.

killianmuldoon:
@Eoghan1232 - wondering if I could get your eyes on this one

Eoghan1232 (approved):

Eoghan1232:
lgtm - apologies I have been away for some time and only back now.

killianmuldoon:
@adrianchiris @dougbtv WDYT about merging this one?

dougbtv (approved):

dougbtv:
thanks for the submission, after considerable discussion: we couldn't find a reason why underscores would be invalid

PR #1252, opened 2024-03-28
Configuration writes should be atomic

dougbtv:
This change introduces a method `atomicFileWrite(...)` which writes to a temporary file and then renames it to avoid partial reads, e.g.

```
Mar 29 04:13:05 labkubedualhost-node-1 crio[33593]: time="2024-03-29 04:13:05.977398607+09:00" level=info msg="CNI monitoring event CREATE        \"/etc/cni/net.d/00-multus.conf\""
Mar 29 04:13:05 labkubedualhost-node-1 crio[33593]: time="2024-03-29 04:13:05.977767189+09:00" level=error msg="Error loading CNI config file /etc/cni/net.d/00-multus.conf: error parsing configuration: unexpected end of JSON input"
```

### DIFF
--- a/pkg/server/config/manager.go
+++ b/pkg/server/config/manager.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/server/config/manager.go
```diff
@@ -64,6 +65,32 @@ func NewManager(config MultusConf) (*Manager, error) {
 	return newManager(config, defaultPluginName)
 }
 
+// atomicFileWrite writes our configs atomically to avoid partial reads
+func atomicFileWrite(filename string, data []byte, perm os.FileMode) error {
```
s1061123, line 69: As warning mentiond, perm is not used in the function hence, none changed the files permission to 0644. Could you please change the file permission?

pkg/server/config/manager.go
```diff
@@ -64,6 +65,32 @@ func NewManager(config MultusConf) (*Manager, error) {
...
+	// Create a temporary file in the same directory as the destination file
+	dir := filepath.Dir(filename)
+	tmpFile, err := ioutil.TempFile(dir, ".multus.tmp")
```
s1061123, line 72: ioutil.TempFile() is deprecated as https://pkg.go.dev/io/ioutil#TempFile so please change that.

s1061123 (changes requested):

pkg/server/config/manager.go
dougbtv: https://pkg.go.dev/os#CreateTemp

PR #1261, opened 2024-04-16
Enforce termination message policy on all platform pods

dougbtv:
Picked from suggestion @ https://github.com/openshift/multus-cni/pull/227/files

Change terminationMessagePolicy to FallbackToLogsOnError for enhanced error visibility, Kubernetes will use the last chunk of the container's log output as the termination message


s1061123:
LGTM, merged!

PR #1265, opened 2024-04-18
Update e2e tests (replace j2cli with gomplate)

### LINKED ISSUE #1264: [e2e] need to replace j2cli with something
**What would you like to be changed**:

j2cli, used in e2e, is no longer active as https://github.com/kolypto/j2cli/issues/80#issuecomment-1937052924. 
Currently github action works well, but j2cli does not work well with python (>=3.12), hence we need to replace it with something else.

Here is the requirements that I come up with:
- something familier with go
- less dependency (i.e. runtime version)

s1061123:
Fix #1264

### DIFF
--- a/.github/workflows/kind-e2e.yml
+++ b/.github/workflows/kind-e2e.yml
@@ -43,11 +43,6 @@ jobs:
         with:
           python-version: 3.x
 
-      - name: Setup j2cli
-        run: |
-          sudo apt-get install -y j2cli
-          echo $(j2 --version)
-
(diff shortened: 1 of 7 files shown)


killianmuldoon:
Am I right in seeing that this tool is only used on the `e2e/templates` folder and that it only sets the CNI_VERSION in ~4 places? 

If so maybe envsubst is a good replacement - it's installed on most systems and it's exactly for this use case.

s1061123:
I got it. May need to file another one.

PR #1270, opened 2024-04-25
thin plugin: Handle `--multus-master-cni-file-name` flag

### LINKED ISSUE #1226: Master CNI file flag option isn't working for thin plugin
**What happend**:
I am trying to use use the `--multus-master-cni-file-name` flag with the thin-plugin. It is expected that it should use the cni file name passed here everytime as default master cni and not the alphabetically first file in the cni-bin directory. But this feature does not seem to work and it always picks up the first file alphabetically instead of what is passed in that flag. 

I checked the code and I see that the file name is taken and is not being used to configure anything in this function: https://github.com/k8snetworkplumbingwg/multus-cni/blob/8e5060b9a7612044b7bf927365bbdbb8f6cde451/cmd/thin_entrypoint/main.go#L301

Tested this for thick-plugin and works as expected. 

It was expected that the master cni is always the plugin file that is passed. 

Install the helm chart on a k8s cluster. Additionally, include the flag `--multus-master-cni-file-name` under `args` in the daemonset yaml. 

**Environment**:

- Multus version : lates
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`): 1.25
- Primary CNI for Kubernetes cluster: calico
- OS (e.g. from /etc/os-release):
- File of '/etc/cni/net.d/'

ordovicia:
Multus v3.9.3 has `--multus-master-cni-file-name` flag to specify the name of a primary CNI config file.
https://github.com/k8snetworkplumbingwg/multus-cni/blob/v3.9.3/images/entrypoint.sh#L22

In Multus v4.0.2, the thin plugin has the flag defined, but it is not read and so does not have effect.

This pull request fixes the problem by making the thin plugin correctly handles `--multus-master-cni-file-name` flag.

Fixes #1226

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -310,22 +310,32 @@ const multusConfTemplate = `{
 }
 `
 
-func (o *Options) createMultusConfig(prevMasterConfigFileHash []byte) (string, []byte, error) {
-	// find master file from MultusAutoconfigDir
-	files, err := libcni.ConfFiles(o.MultusAutoconfigDir, []string{".conf", ".conflist"})
-	if err != nil {
-		return "", nil, fmt.Errorf("cannot find master CNI config in %q: %v", o.MultusAutoconfigDir, err)
+func (o *Options) getMasterConfigPath() (string, error) {
+	// Master config file is specified
+	if o.MultusMasterCNIFileName != "" {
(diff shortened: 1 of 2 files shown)


s1061123:
@ordovicia, Looks good to me, thank you for the PR!

PR #1273, opened 2024-05-02
CNI 1.1.0 support

s1061123:
This PR support CNI SPEC 1.1.0, with new CNI commands, GC and STATUS.


s1061123:
Need to discuss: May need to disable GC for now because it may happens unexpected resource removal by GC (if same CNI plugin is used for both, cluster network and net-attach-def, it mihgt be  happen)

s1061123:
> Need to discuss: May need to disable GC for now because it may happens unexpected resource removal by GC (if same CNI plugin is used for both, cluster network and net-attach-def, it mihgt be happen)

Changed.

dougbtv (approved):
Some notes from our call:

* GC and STATUS
  * Hard to invoke STATUS in a multus context, e.g. no status for macvlan as a delegated plugin, for example.
  * Maybe only for cluster networks
  * GC more complicated
    * Only stores cluster network object.
  * GC and STATUS are just delegated.

So, this delegation is what Multus is designed to do, I think this is the right place to start.

champtar:
@dougbtv there is a typo in the release notes https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v4.2.0
you say "Supports CNI spec v1.2.0", it should be "v1.1.0"

PR #1274, opened 2024-05-09
Fix CNI cache update function to prevent nil access

s1061123:
deleteDefaultGWResult() may create 'routes:null' in CNI cache file and it causes nil pointer access at addDefaultGWCacheBytes(). This code change prevents deleteDefaultGWResult() to generate 'routes:null' in cache file.


dougbtv:
results in the error:

```
wrong routes format
```

dougbtv (approved):

PR #1275, opened 2024-05-09
Enhance CNI flexibility to read and write deviceInfo by removing the deviceID restriction

jiayoukun:
Based on the Device Information Specification described in the provided links 
[https://github.com/k8snetworkplumbingwg/device-info-spec/]
[https://github.com/k8snetworkplumbingwg/device-info-spec/blob/main/SPEC.md]

The NPWG Implementation Requirements necessitate checking the deviceID for creating the "CNIDeviceInfoFile", which lacks flexibility. My understanding is that CNIs should also have read and write permissions for deviceInfo. However, in the current Multus code, if the deviceID is not configured in the NetworkAttachmentDefinition (NAD), the "CNIDeviceInfoFile" is not created and passed to the CNI. I believe this part of the code has limitations.

In the current code, I added a condition for generating the "CNIDeviceInfoFile". As long as Capabilities["CNIDeviceInfoFile"] is configured as true in the NAD, an empty file is generated. This way, the "CNIDeviceInfoFile" can be passed to the CNI, allowing CNIs without a deviceID to participate in reading and writing deviceInfo. All the deviceInfo will eventually be returned to the kubelet.

Furthermore, I plan to extend the Device Information Specification in the future to include more adaptable fields, enhancing its flexibility and accommodating a wider range of scenarios.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -640,6 +640,16 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 			if err != nil {
 				logging.Debugf("CmdAdd: CopyDeviceInfoForCNIFromDP returned an error - err=%v", err)
 			}
+		} else if delegate.Conf.Capabilities["CNIDeviceInfoFile"] {
+			devInfo := &nettypes.DeviceInfo{}
+			err = nadutils.SaveDeviceInfoForCNI(cniDeviceInfoPath, devInfo)
+			// For CNIs with Capabilities["CNIDeviceInfoFile"] set to true, a deviceinfofile is
+			// created to conveniently pass the CNIDeviceInfoFile to the CNI. This allows the CNI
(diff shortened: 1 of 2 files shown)


dougbtv:
Any chance you could join our community call to talk about this some more? The connection information can be found in [the maintainer's agenda](https://docs.google.com/document/d/17-vkuoObBFBb6feXNA7tjTaKr-rs8ea0VTnqvlEdCv8/edit)

Thanks for the contribution, we'd love to chat about it some more

jiayoukun:
@dougbtv 
I apologize for my poor spoken English. I'm afraid that attending the meeting might affect everyone's time. Could we discuss the current PR's content here instead?

The functionality modified in this PR is quite effective and flexible for extending CNI development of more deviceInfo. Wasn't the original intention of using deviceInfo to allow NPWG and CNI to share extended device information? However, the current situation is that CNIs without a deviceID are all limited within NPWG and don't have the opportunity to manipulate deviceInfo.

I hope everyone can provide some assistance! Thank you for the reply!

PR #1282, opened 2024-05-19
update cni to 1.2.0

scydas:


s1061123:
Thank you for the PR, @scydas, but it seems to be overlapped to #1273. #1273 will also introduce new CNI command, INIT and GC, so it supersedes yours. Hence let me close that.

PR #1286, opened 2024-05-24
Not exposed APIReadyCheckFunc to outside of package

s1061123:
https://github.com/k8snetworkplumbingwg/multus-cni/actions/runs/9225475657 recommends to change the code, so this PR fix the code. In addition, APIReadyCheckFunc  can be decaptalize (i.e not expose this type to outisde of package), so change it decaptalized.

```
comment on exported type APIReadyCheckFunc should be of the form "APIReadyCheckFunc ..." (with optional leading article) test (1.21.x, ubuntu-latest): pkg/server/api/shim.go#L44
type name will be used as api.APIReadyCheckFunc by other packages, and that stutters; consider calling this ReadyCheckFunc
```

### DIFF
--- a/pkg/server/api/shim.go
+++ b/pkg/server/api/shim.go
@@ -40,8 +40,8 @@ type ShimNetConf struct {
 	LogToStderr     bool   `json:"logToStderr,omitempty"`
 }
 
-// Define a type for API readiness check functions
-type APIReadyCheckFunc func(string) error
+// readyCheckFunc defines a type for API readiness check functions
+type readyCheckFunc func(string) error
 
 // CmdAdd implements the CNI spec ADD command handler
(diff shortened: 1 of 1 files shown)


pkg/server/api/shim.go
```diff
@@ -40,8 +40,8 @@ type ShimNetConf struct {
 	LogToStderr     bool   `json:"logToStderr,omitempty"`
 }
 
-// Define a type for API readiness check functions
-type APIReadyCheckFunc func(string) error
+// readyCheckFunc defines a type for API readiness check functions
+type readyCheckFunc func(string) error
```
dougbtv, line 44: can we call it `apiReadyCheckFunc`

dougbtv (approved):

PR #1290, opened 2024-05-29
Bidirectional mount propagation for hostroot

dougbtv:
Modifies the `hostroot` volume in the `multus-daemonset-thick.yml` example/quickstart deployment file from `HostToContainer` to `Bidirectional`. This change enables the volume to be accessible in both directions, which is necessary for users who need to share a mount with another container/pod.

This is motivated by the fact that in thin plugin mode, since all things were run on the host directly, CNI plugins wouldn't be limited by the mount propagation.

One example that has come up recently is userspace CNI interaction with kubevirt, and sharing the usage of the socket as mounted by kubevirt.

This does expose some level of risk (as noted in the mount propagation docs regarding changing of mounts), however, I don't believe it's significantly more than would've been the case in thin plugin mode.

References
- [Kubernetes Volume Mount Propagation Documentation](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation)

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -190,7 +190,7 @@ spec:
               readOnly: true
             - name: hostroot
               mountPath: /hostroot
-              mountPropagation: HostToContainer
+              mountPropagation: Bidirectional
           env:
             - name: MULTUS_NODE_NAME
               valueFrom:


deployments/multus-daemonset-thick.yml
```diff
@@ -190,7 +190,7 @@ spec:
               readOnly: true
             - name: hostroot
               mountPath: /hostroot
-              mountPropagation: HostToContainer
+              mountPropagation: Bidirectional
```
maiqueb, line 193: don't we actually consider these yamls to be just examples ? 

Why can't the downstreams that require it to be bi-directional use what they need ? 

I'm trying to wrap my head around the fact this is not the recommended configuration, required only for specific use cases, but still, we're advocating / instructing everyone to use it.

dougbtv: These are indeed examples, but, I think this more accurately represents a setup that's similar to what happens in a thin plugin mode on the host -- you'd see mounts in both directions. So, I think it's a valid setup and probably better for more cases at least.

s1061123 (changes requested):
How about to add same config to kind-e2e config? Others are okey!

deployments/multus-daemonset-thick.yml
maiqueb: I disagree, but don't have more arguments other than this is not the recommended configuration, and is required only for specific use cases.

Do we even know if this will work on openshift ?

PR #1297, opened 2024-06-12
Build multiarch thick images

sfackler:
I'm not totally sure why the thick image is different from the thin one here, but it doesn't seem like there should be anything stopping the thick image from being multiarch.

### DIFF
--- a/.github/workflows/image-build.yml
+++ b/.github/workflows/image-build.yml
@@ -35,8 +35,8 @@ jobs:
           sbom: false
           provenance: false
 
-  build-amd64-thick:
-    name: Image build/amd64 thick plugin
+  build-thick:
+    name: Image thick plugin
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


dougbtv (changes requested):
I think we can probably move forward as this matches the thin plugin, but, we realized that we don't really have a disclaimer anywhere to say "we don't actually test the other platforms" -- because we just run the e2e tests on x86.

Let's add some text to the development markdown:

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/development.md

```
# Multi-arch builds

Multus is currently built for a number of architectures, however, our testing and validation is only performed against x86 architectures. Our x86 architecture has end to end testing, however, for other architectures, only supported via best effort community contributions.
```

dougbtv (approved):
thanks for the updates, appreciate it

PR #1298, opened 2024-06-16
Use same base image as thick plugin

adrianchiris:
change thin plugin dockerfile to use the same
base image as the thick plugin.

this will allow users to implement simple lifecycle hooks e.g exec simple shell cmd to delete multus conf file on container exit which allows to remove multus (thin) from the cluster without breaking it.

this functionality was possible in v3.

Related issue: #592

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -8,7 +8,7 @@ ARG TARGETPLATFORM
 RUN  cd /usr/src/multus-cni && \
      ./hack/build-go.sh
 
-FROM gcr.io/distroless/base-debian11:latest
+FROM debian:stable-slim
 LABEL org.opencontainers.image.source https://github.com/k8snetworkplumbingwg/multus-cni
 COPY --from=build /usr/src/multus-cni/bin /usr/src/multus-cni/bin
 COPY --from=build /usr/src/multus-cni/LICENSE /usr/src/multus-cni/LICENSE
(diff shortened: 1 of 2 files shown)


adrianchiris:
An alternative solution is to actually clean up multus config file on exit if `cleanup-config-on-exit` flag is set in `thin_entrypoint`[1]

note: thick plugin today cleans up multus config file on exit[2]

[1] https://github.com/k8snetworkplumbingwg/multus-cni/blob/aff99fccc5fe3b21aec84bb56a6d1bbaf8ffc45b/cmd/thin_entrypoint/main.go#L607
[2] https://github.com/k8snetworkplumbingwg/multus-cni/blob/aff99fccc5fe3b21aec84bb56a6d1bbaf8ffc45b/pkg/server/config/manager.go#L163

s1061123:
We use distroless for less maintainance for CVE in container image because distroless image, `gcr.io/distroless/base-debian11:latest`, has less components than distro image, `debian:stable-slim`. So we (or I, at least), expects to change thick pluign container image to distroless in the future. Not to change thin plugin to distro image.

Currently our release container image is just for simple use cases and it does not care about whole lifecycle for all deployment scenarios (on all platforms). If you want to do more, we expect that you should have original container image for that.

So I suppose we should not change thin pluign base images.

adrianchiris:
closing this one in fafor of #1299

PR #1299, opened 2024-06-18
Cleanup multus conf on exit

adrianchiris:
- if cleanup-config-on-exit is set delete generated multus
  config file on exit.

- add an option to skip watch for master cni config and kubeconfig
  as cleanup-config-on-exit with multus-conf-file=auto also triggered
  the watch for cases when deletion of multus config is desired but watch isnt

- setup signal handling to allow config file cleanup on exit

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/spf13/pflag"
 
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/cmdutils"
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/signals"
 )
 
 // Options stores command line options
@@ -57,6 +58,7 @@ type Options struct {
 	AdditionalBinDir         string
 	ForceCNIVersion          bool
 	SkipTLSVerify            bool
+	SkipMultusConfWatch      bool
 }
 
 const (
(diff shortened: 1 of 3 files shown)


adrianchiris:
This is a proposed alternative to : #1298

as we used to clean up multus conf file in preStop hook, however since the container no longer has shell we cannot run simple sh commands for cleanup.

dougbtv:
Let's add some notes about the flags @ https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md#entrypoint-script-parameters

adrianchiris:
@dougbtv docs update. 

additional feedback/review appreciated :)

seastco:
@dougbtv @adrianchiris ty for landing this.

I take it disabling the watch functionality is necessary if CleanupConfigOnExit=true and MultusConfFile=auto because the file may be recreated before we're able to completely exit? this seems like CleanupConfigOnExit and watch are incompatible, which is unfortunate because I need both.

could we instead send a signal to the watch loop to stop before CleanupConfigOnExit is called? I'm thinking something like 
```
stopWatchChan := make(chan bool)

if opt.CleanupConfigOnExit {
        defer func() {
            stopWatchChan <- true // Send a stop signal to the watch loop
            cleanupMultusConf(&opt)
        }()
}
```

adrianchiris:
@seastco there is a select statement on signal handler context.  cleanup config on exit should work both if watching or not.

the "skip-config-watch" flag was added to preserve backward compatibility since cleanup config on exit originally triggered the watch for some reason (imo it shouldnt and is counter intuitive), but it is what it is.

PR #1300, opened 2024-06-18
Add mountPropagation to host-var-lib-kubelet

adrianchiris:
This is needed to avoid issues with volume unmounts when (other) pods exit as by default kubernetes mounts as private.


dougbtv (approved):
I tend to think this is probably a good call, and it fits the other mount propagation that we have for hostroot

PR #1311, opened 2024-07-15
[WIP] Ensure deviceID is set when a resource is requested

vasrem:
If a NAD has the resourceName annotation set, then it means that the subsequent CNI calls should provide the deviceID for that particular resource.

Previously, we were doing best effort to get the deviceID from kubelet. With this patch, we guarantee the deviceID will be passed in the CNI call and fail if we can't find any deviceID.

### DIFF
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -312,6 +312,10 @@ func getKubernetesDelegate(client *ClientInfo, net *types.NetworkSelectionElemen
 				entry.Index++ // increment Index for next delegate
 			}
 		}
+
+		if deviceID == "" {
+			return nil, resourceMap, logging.Errorf("getKubernetesDelegate: requested resource not found in resourceMap %+v", resourceMap)
+		}
 	}
 
 	configBytes, err := netutils.GetCNIConfig(customResource, confdir)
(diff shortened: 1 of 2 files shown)


dougbtv:
let's also make sure the functionality is described in: https://github.com/k8snetworkplumbingwg/multus-cni/blob/41013e7580396629a02213019baad1d0c84a6a2c/docs/how-to-use.md?plain=1#L677

dougbtv (changes requested):
Let's make sure this functionality is described in: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/docs/how-to-use.md?plain=1#L677

vasrem:
Closing this PR as this bugfix breaks the DRA integration. To support the DRA integration, we have to switch the e2e test to use `Immediate` allocation which is not supported by the dra-example https://github.com/kubernetes-sigs/dra-example-driver/blob/classic-dra/cmd/dra-example-controller/driver.go#L108-L110.

On a similar note, DRA API for 1.31 Kubernetes will change, so the plan is to:
1. Update the current e2e test to target the "classic dra" which is supported until Kubernetes 1.30
2. Adjust the multus integration for 1.31 DRA when 1.31 is released
3. Re-open this PR to ensure the bug is fixed.

PR #1314, opened 2024-07-25
Updates to use CreateNetworkStauses from net-attach-def client, bump to v1.7.1

dougbtv:
NOTE: This functionality is dependent on https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/pull/68 and will require a library update in this PR in order to be complete.

This functionality properly represents Network Status for CNI ADD results that return multiple interfaces themselves.

This approach uses a new method from the net-attach-def-client -- `CreateNetworkStatuses`, and refactors to account for the array return.

With the previous method, it's somewhat buggy. If multiple pod interfaces are present, it returns the last one -- and groups all of the IP addresses to that address.

For replication of the error yourself, and to see the fix in action, see this gist: https://gist.github.com/dougbtv/1eb8ac2d61d494b56d65a6b236a86e61

For the related suggested update to the NPWG net-attach-def specification, see this proposal @ https://docs.google.com/document/d/1DUTV-o6E6zlRTKZkxeDhAeyGrmq03qPgPbU580Rm7-g/edit

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/containernetworking/plugins v1.1.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/go-logr/logr v1.3.0 // indirect
-	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.7.0
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.7.1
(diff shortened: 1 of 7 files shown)


adrianchiris (approved):
/lgtm

SchSeba (approved):
LGTM great!

maiqueb (approved):
Thank you for this.

PR #1322, opened 2024-08-12

mingshuoqiu:
The original PR is https://github.com/k8snetworkplumbingwg/multus-cni/pull/1232. This PR rebase the commit on top of the current master.

### DIFF
--- a/docs/how-to-use.md
+++ b/docs/how-to-use.md
@@ -627,6 +627,10 @@ master CNI configuration and kubeconfig. when such change detected, the script w
 
     --skip-config-watch
 
+The poll interval of the config changes watcher is configurable with `--multus-config-watch-timer=<duration>`. Please refer to the [time.Duration](https://pkg.go.dev/time#Duration) to configure the time interval. The following example watches the config change every 5 minutes.
+
+    --multus-config-watch-timer=5m
+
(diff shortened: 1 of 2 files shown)


docs/how-to-use.md
```diff
@@ -627,6 +627,10 @@ master CNI configuration and kubeconfig. when such change detected, the script w
 
     --skip-config-watch
 
+The poll interval of the config changes watcher is configurable with `--multus-config-watch-timer=<duration>`
+
```
dougbtv, line 631: Let's also provide a link to whatever appropriate go docs about what are valid values for "duration" here. Otherwise, lgtm

dougbtv (changes requested):

docs/how-to-use.md
mingshuoqiu: @dougbtv Thanks. I've updated the description with the go doc link for the `duration`.  Please let me know if any suggestions.

mingshuoqiu:
rebase on top of the latest [adfb270](https://github.com/k8snetworkplumbingwg/multus-cni/commit/adfb270991d410417a71755d01f53445510ef2a0) of k8snetworkplumbingwg:master

docs/how-to-use.md
bk201: @dougbtv Thanks for reviewing the PR. Can we merge the PR or is there something can do to make it better?

bk201:
unstale, @dougbtv can you help review the PR? thanks!

PR #1323, opened 2024-08-28
upgrade cni plugin version to use new feature

a180285:
upgrade 
	github.com/containernetworking/plugins v1.1.0 to v1.5.1

to use new features. Like below:
https://github.com/containernetworking/plugins/commit/821982da1c9b96870b2459d1cb972d6e5dc0a1d2

This new parameter allows users to remove the default vlan

Fixes: https://github.com/containernetworking/plugins/issues/667


a180285:
Hi @dougbtv @s1061123. Can you give a review on this.

dougbtv:
I think it's probably OK to get updates to the CNI requirements, but...

...question -- does this really address the issue? like, if those changes are in the resulting binaries, does that address your issue if it's just the libraries updated in Multus?

s1061123:
Based on  your comment, you want to have fix for `bridge` CNI pluign. But `multus` repository does not have any `bridge` CNI code, hence your issue does not fixed even if this PR is merged.

> This new parameter allows users to remove the default vlan
This parameter is added in  `bridge` CNI plugin, not `multus` CNI plugin. If you want to introduce it, you just update `bridge` CNI, as get the latest plugin from https://github.com/containernetworking/plugins/releases and install.

a180285:
Thanks @dougbtv @s1061123 
Yes, I see this issue on bridge plugin. I need that fix. 

Can you help show me how to update the plugin instead of update mulus? 

Because I think multus depends on plugin repo, so the plugin code is built in multus, Is that wrong?

dougbtv:
Yeah, those binaries aren't automatically included with Multus. Multus just uses some specific functionality from the plugins repo (we think for netns utilities). You would build the binaries from the containernetworking/plugins repo, and then install them onto the hosts :+1:

dougbtv:
I think this cnilib bump might be bigger than just a bump and might require some functional changes in the tests, so, let's close this out for now. Thanks for the contribution and asking the question

a180285:
Here is my config. Does that mean I should update `kube-ovn` or update plugin on my host?

```
---
apiVersion: k8s.cni.cncf.io/v1
metadata:
  name: vlan4009
  namespace: default
spec:
  config: '{
     "cniVersion": "0.3.1",
     "name": "vlan4009",
     "type": "bridge",
     "bridge": "br0",
     "vlan": 4009,
     "ipam": {
        "type": "kube-ovn",
        "server_socket": "/run/openvswitch/kube-ovn-daemon.sock",
        "provider": "vlan4009.default"
      }
     }'
```

a180285:
@dougbtv Thanks for you info. If I need update cni-plugin on my host. Can you kindly provide more info, on how to upgrade cni to my host? Thanks very much.

PR #1332, opened 2024-09-11
always attempt a live pod get on miss to confirm its really not there

dougbtv:
From [openshift pull request](https://github.com/openshift/multus-cni/pull/246):

> This makes sure that stale caches never result in NotFound errors.

It was explained to me that informers are almost always are more efficient, and in most cases will work, but a live lookup is appropriate after a number of failures.

This happens only on the retry portion, so we're still getting the benefits of informers, but, on a retry situation, we don't get a cache miss.


s1061123 (changes requested):

bpickard22:
bump to remove stale

bpickard22:
name change looks good

PR #1336, opened 2024-09-13
Bumps net-attach-def client to v1.7.3

dougbtv:
Previous version didn't account for accounts for the sandox interfaces when reporting the interfaces in the network-status annotation when calculating the default:true interface


maiqueb (approved):
Thank you @dougbtv !!!

PR #1337, opened 2024-09-17
Don't have entrypoint.sh in multus-daemonset-crio.yml

Sagit-chu:
Error: container create failed: executable file `/entrypoint.sh` not found in $PATH: No such file or directory


dougbtv:
Thanks for the contribution! ...Maybe we should get an e2e test for this yaml, but... otherwise, looks good, someone else should give it a whirl to verify, but the e2e test would be perfect.

PR #1343, opened 2024-10-10
[DO NOT MERGE] Tests change to net-attach-def library for gateway determination of d…

dougbtv:
…efault=true in status

This is a test change to the library for determination of the default=true interface in the network-status annotation when multiple interfaces are present in the result.

### DIFF
--- a/vendor/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils/net-attach-def.go
+++ b/vendor/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils/net-attach-def.go
@@ -136,6 +136,11 @@ func CreateNetworkStatuses(r cnitypes.Result, networkName string, defaultNetwork
 		return nil, fmt.Errorf("error converting the type.Result to cni100.Result: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


vendor/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/utils/net-attach-def.go
```diff
@@ -136,6 +136,11 @@ func CreateNetworkStatuses(r cnitypes.Result, networkName string, defaultNetwork
 		return nil, fmt.Errorf("error converting the type.Result to cni100.Result: %v", err)
 	}
 
+	if len(result.Interfaces) == 1 {
+		networkStatus, err := CreateNetworkStatus(r, networkName, defaultNetwork, dev)
+		return []*v1.NetworkStatus{networkStatus}, err
+	}
```
maiqueb, line 142: I don't think this is needed; the library itself is doing it, right ? 
https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/9b218a2d42c61fbf61a36fac990a2d67befdb30a/pkg/utils/net-attach-def.go#L139

Issue #1346, opened 2024-10-16
[OOMKilled] High memory consumption


juliusmh:
**What would you like to be added**:

The option to specify the number of parallel CNI requests being processed. In a scenario where many pods come to alive at the same time, memory of the multus pods (thick-mode) increases rapidly. Hence, I suggest we add a function to limit the concurrency. 

**Why is this needed**:

This will prevent setting higher memory requests, which are really a waste as they're only required in these bursty scenarios.

Btw, I'm happy to help with the implementation.

juliusmh:
Not stale.

JensErat:
We successfully applied #1356 as mitigation. Even a limit of 500Mi did not resolve an OOM-loop in one of our rather "involved" use cases.

evilhamsterman:
@JensErat That's a much better fix, mine is just a bandaid. Agreed 200M sometimes doesn't even get me out of a oom loop, most of the time it works, but sometimes not

erickuiper:
Not stale

JensErat:
still an issue and has proposed solution

fdomain:
We're running into the same issue, any chance we could see the related PR https://github.com/k8snetworkplumbingwg/multus-cni/pull/1356 or https://github.com/k8snetworkplumbingwg/multus-cni/pull/1437 merged ?

SchSeba:
Hello I make a comment here https://github.com/k8snetworkplumbingwg/multus-cni/pull/1458#pullrequestreview-3530600812

if you are able to run a pprof so we can see where the memory is mainly used to see if we can improve that area it will be great!

if you need help providing an image what enabled pprof let me know and I can make one

attila123:
Hi, I am experiencing OOMKilled in the kube-multus container as it is currently having 50Mi memory limit:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/deployments/multus-daemonset-thick.yml#L165 .  
Sometimes this error goes away after the first run of the container (?) and it works only after the 1st container restart.
Other times it goes to CrashLoopBackOff because of that.  
-> but this is now the known behavior I suppose, just wanted to summarize that.

Is the connection limit PR https://github.com/k8snetworkplumbingwg/multus-cni/pull/1458 supposed to fix this issue?
Until this fixed I do a quick and dirty patch in the downloaded file (which also changes the `requests` value, but it's OK for now):
```
[...]

PR #1347, opened 2024-10-16
Limit HTTP server's concurrency using semaphore.

juliusmh:
Signed-off-by: Julius Hinze <juliusmh@proton.me>

Closes: https://github.com/k8snetworkplumbingwg/multus-cni/issues/1346

### DIFF
--- a/docs/thick-plugin.md
+++ b/docs/thick-plugin.md
@@ -72,6 +72,7 @@ is provided.
 - `"logLevel"`: the logging level for the multus daemon logs.
 - `"logToStderr"`: enable this to have the daemon multus logs echoed to stderr
 as well. By default, it is disabled.
+- `concurrentExecs`: integer that, if specified, defines the amount of parallel chroot plugin executions (optional).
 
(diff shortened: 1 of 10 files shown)


dougbtv:
Overall, I'm in favor of it, and I like that it's parameterized, but I'd like to see that pods do eventually come up in a situation where there's more than the concurrency limit, I believe those pods would crash loop?

I think maybe we should have some testing for it, best case scenario: e2e tests that both capture the error and the success.

dougbtv:
Tomo also suggested taking a look at the limiters on https://pkg.go.dev/golang.org/x/net/netutil

juliusmh:
> Overall, I'm in favor of it, and I like that it's parameterized, but I'd like to see that pods do eventually come up in a situation where there's more than the concurrency limit, I believe those pods would crash loop?
> 
> I think maybe we should have some testing for it, best case scenario: e2e tests that both capture the error and the success.

These pods would wait as the semaphore blocks, until the shim timeouts the requests. In this case the pod will fail and kubelet try again in the future. 

We should decide if we want to
- fail the HTTP request if concurrency is too high, retry from shim
- semaphore on listener level (`LimitListener`) or inside the handler (this PR)

juliusmh:
@dougbtv fixed unit test & rebased

s1061123:
I suppose using `LimitListener`, we don't need to introduce semaphore because `LimitListener` does the same things. Can we close it and proceed #1356?

juliusmh:
Closing in favor of https://github.com/k8snetworkplumbingwg/multus-cni/pull/1356

PR #1353, opened 2024-11-06
clean up go.mod, get rid of client-go v1.5.2

xrstf:
client-go 1.5.2 has been retracted and having it here as a dependency forces not just Multus, but everyone who imports Multus to do the same `replace` dance with tons of Kube dependencies.

From what I can see, there is simply no need to replace tons of dependencies. So I cleaned it up and `go mod tidy` and `go build` are happy.


dougbtv (approved):
Thanks for the heads up and contribution on this one!

PR #1356, opened 2024-11-12
Connection limit for thick daemon.

juliusmh:
Signed-off-by: Julius Hinze <juliusmh@proton.me>

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -29,6 +29,7 @@ import (
 	"sync"
 	"syscall"
 
+	"golang.org/x/net/netutil"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/logging"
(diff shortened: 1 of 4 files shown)


dougbtv:
Julius -- quick question: what's the difference between #1347 and this one? Could we put them together? Thanks! Went to review #1347 and then saw this one as well

juliusmh:
> Julius -- quick question: what's the difference between #1347 and this one? Could we put them together? Thanks! Went to review #1347 and then saw this one as well

Sure. Let's proceed here.

juliusmh:
@dougbtv is this PR still considered? Is there anything I can do from my end to help get this merged?

JensErat:
This resolved heavy OOM issues in a cluster running a larger 2-digit number of pods with multus enabled. Rebooting a node reproducible forced multus into an OOM loop, stopping all pods from running. I'd propose to additionally either add this to the configuration or even set a reasonable default value (I was randomly choosing 5).

@dougbtv does this have any chance of getting merged? Currently, we're considering whether we'll have to run a forked copy.

juliusmh:
Not stale

juliusmh:
Reopening in https://github.com/k8snetworkplumbingwg/multus-cni/pull/1437.

PR #1363, opened 2024-12-03
fix: md lint

zbb88888:
fix markdown lint

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,10 +1,11 @@
+#
+
 ---
 name: Bug Report
 about: Report a bug encountered
 
 ---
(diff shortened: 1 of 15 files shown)


dougbtv:
Thanks for the contrib! I appreciate it. But if we're going to lint markdown, let's keep it that way for keeps -- if you don't mind, add some automation or check for it so we can keep it tidy for good. Appreciate it! Thanks

bobz965:
> Thanks for the contrib! I appreciate it. But if we're going to lint markdown, let's keep it that way for keeps -- if you don't mind, add some automation or check for it so we can keep it tidy for good. Appreciate it! Thanks

ok, I will try

bobz965:
> > Thanks for the contrib! I appreciate it. But if we're going to lint markdown, let's keep it that way for keeps -- if you don't mind, add some automation or check for it so we can keep it tidy for good. Appreciate it! Thanks
> 
> ok, I will try

add .github/workflows/md-lint.yaml

PR #1372, opened 2024-12-19
Thin plugin: don't wait too long for an answer from API Server

pmtk:
If Multus plugin gets a DEL request, but the API Server is down (e.g. via 'crictl rmp'), the call takes so long, it actually never finishes. This prevents CRI-O from deleting the Pods.

Proposed fix sets a timeout for Pods().Get() using context.


dougbtv (approved):
nice approach with the context timeout. Potentially this can be tuned further on DEL and skip API requests where they're not necessary, but this is really clean and still efficient. Thanks

PR #1374, opened 2024-12-29
fix: dockerfile change cmd to entrypoint

buroa:
Pretty simple change. Would allow Kubernetes deployments to specify just args instead of both command + args.


buroa:
Hi @dougbtv, this one should be fairly easy to pull in. It looks like the OpenShift image already has it: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/Dockerfile.openshift

buroa:
@s1061123 Any help here? @dougbtv?

buroa:
@RamLavi Can you please help here?

buroa:
@dougbtv @bpickard22 Any help here?

bpickard22:
@buroa sorry this took so long to get to, thanks for the patch

PR #1375, opened 2025-01-06
enforce copy the multus-cni

dgsponer:
is the multus-cni already there, cp is not working and wait for a user interaction. with parameter -f the copy command will be forced.


evilhamsterman:
This is a duplicate effect of PR #1213

Issue #1387, opened 2025-02-12
Container is not getting additional interface on k8s node reboot


glebkin:
Hi! We're using Multus with Cilium to supply some of our apps with additional interfaces.

Example of CR:
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: nad-dhcp
spec:
  config: '{
             "cniVersion":"0.3.1",
             "name":"nad-dhcp",
             "plugins":[
                {
                   "type":"bridge",
                   "bridge":"br111",
                   "isGateway":true,
                   "ipam":{
                      "type":"static",
                      "addresses":[
                         {
                            "address":"169.254.169.103/24",
                            "gateway":"169.254.169.1"
                         }
                      ],
                      "routes":[
                         {
                            "dst":"169.252.252.0/24",
                            "gw":"169.254.169.1"
                         }
                      ]
                   }
                }
             ]
          }'
```

The problem is that after node reboot our container is not getting additional interface from Multus, instead we can the see the following errors in Multus logs:
```
time="2025-02-12T16:02:07Z" level=warning msg="Errors encountered while deleting endpoint" error="[DELETE /endpoint][404] deleteEndpointNotFound " subsys=cilium-cni
time="2025-02-12T16:02:07Z" level=warning msg="Unable to enter namespace \"\", will not delete interface" error="failed to Statfs \"\": no such file or directory" subsys=cilium-cni
```
To resolve this issue, we need to restart the pod that uses the Multus additional interface.
Additionally, if we simply kill the container within a pod, it will be restored with the additional interface, but there will be no log messages from Multus. It appears that Cilium is handling all the work.

When using Multus with Flannel, we didn't experience these issues.
Can you please advise on what we might be doing wrong?

dougbtv:
Have you checked the /etc/cni/net.d dir to see if cilium deleted the multus config?

glebkin:
> Have you checked the /etc/cni/net.d dir to see if cilium deleted the multus config?

Many thanks for the tip! I found that Multus automatically generates 00-multus.conf in /etc/cni/net.d dir and automatically deletes it on shutdown. 
So each time we restart node - multus deletes its config. When kubelet starts - it starts Multus and our app almost at the same time and our app gets only primary interface from Cilium, as Multus config still not exists.
So I tried to comment piece of code with config deletion - now if you restart multus and app using additional interface at the same time - it works fine.
[...]

PR #1391, opened 2025-02-14
Add multusKeepConfig option

glebkin:
On the basis of https://github.com/k8snetworkplumbingwg/multus-cni/issues/1387

I tried to make this change backwards compatible, so you need to enable multusKeepConfig option in order to persist `/etc/cni/net.d/00-multus.conf` configuration file. This change is really important for us, as we're using Cilium+Multus in our k8s environment and we need our services to be fault tolerant. Without this option our containers with additional interfaces will not survive node reboots.


glebkin:
Would you please kindly take a look into this PR?

glebkin:
It's still relevant. Please don't close it.

Issue #1394, opened 2025-02-20
DRA is broken in kube 1.31


bpickard22:
**What happend**:
DRA e2e is perma failing
` [It] should return no error with dynamic resource
/home/runner/work/multus-cni/multus-cni/pkg/kubeletclient/kubeletclient_test.go:245

  [FAILED] Expected
      <map[string]*types.ResourceInfo | len:0>: {}
  to equal
      <map[string]*types.ResourceInfo | len:1>: {
          "resource-class": {
              Index: 0,
              DeviceIDs: ["cdi-resource"],
          },
      }`

**What you expected to happen**:
e2e to pass

**How to reproduce it (as minimally and precisely as possible)**:
Upgrade to kube 1.31 / go 1.23 and run e2e

**Anything else we need to know?**:
Found this issue when bumping multus to go 1.23 which we need for some other dep bumps

PR #1396, opened 2025-02-20
go 1.23 bump

bpickard22:
Bumps to go 1.23 and bumps deps

skips broken tests and links to issues created to track them


dougbtv:
See: #1394 #1397

PR #1411, opened 2025-03-25
[WIP] Bumps k8s libs and go 1.23

bpickard22:
Bump deps (most importantly go to 1.23 and k8s to 1.32)

Skips broken tests (created issues to fix tests)

Create makefile for simpler dev experience


dougbtv:
<3 makefile

PR #1410, opened 2025-03-24
multus-daemonset-thick: Use install_multus for installation

owenthomas17:
Running the e2e tests locally seem to pass when I omit the change to the image in the original PR. I've taken the original commits and am just re-submitting them to make the checks run again since the original contributor doesn't seem to be responding.

See https://github.com/k8snetworkplumbingwg/multus-cni/pull/1213 for the full context.

Fixes: https://github.com/k8snetworkplumbingwg/multus-cni/issues/1221

Feel free to close this PR and fix up any in flight PR's if its easier.

PR #1420, opened 2025-04-10
[do not merge] e2e tests for upgrade  and node reboot

dougbtv:
If we want to fix upgrade issues, we should test upgrade issues.

This includes a test for a node reboot and an "upgrade", which just HUPs the Multus pods to see that a fresh binary is installed.

NOTE: This is done as a nicety for the community. Any lifecycle changes are optional and "best effort" for maintainers to support, as our current deployment is given as an example (see header on the quickstart deployments), and we expect there to be many environment concerns from many different users. This is provided in order to add some example lifecycle management handling, but we encourage you to build your own deployment to suit your environment given the examples and your own experience.

### DIFF
--- a/e2e/test-reboot.sh
+++ b/e2e/test-reboot.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -o errexit
+
+NODE_NAME="kind-worker"
+DAEMONSET_NAME="kube-multus-ds-amd64"
+NAMESPACE="kube-system"
+EXPECTED_BINARIES="${EXPECTED_BINARIES:-/opt/cni/bin/multus-shim}"
+INSTALL_INIT_CONTAINER="${INSTALL_INIT_CONTAINER:-install-multus-shim}"
+
+declare -A before_mtime
+
+for bin in $EXPECTED_BINARIES; do
+    before_ts=$(docker exec "${NODE_NAME}" stat -c %Y "${bin}")
+    before_mtime["${bin}"]=$before_ts
+    echo "Before reboot: ${bin} mtime = ${before_ts}"
+done
+
+echo "Rebooting node..."
+docker restart "${NODE_NAME}"
+sleep 2
+docker start "${NODE_NAME}"
+
+kubectl wait --for=condition=Ready node/${NODE_NAME} --timeout=300s
+kubectl rollout status daemonset/${DAEMONSET_NAME} -n ${NAMESPACE} --timeout=300s
+
+source ./test-check-binaries.sh
+
+echo "SUCCESS: reboot test passed"
--- a/e2e/test-upgrade.sh
+++ b/e2e/test-upgrade.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+set -o errexit
+
+NODE_NAME="kind-worker"
+DAEMONSET_NAME="kube-multus-ds-amd64"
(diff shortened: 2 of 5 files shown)


dougbtv:
we will merge this with the fix for what this demonstrates in #1410

dougbtv:
I can't seem to show that the error exists on the main branch with this test suite? Any ideas? :thinking:

e2e/test-reboot.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo "Rebooting node..."
+docker restart "${NODE_NAME}"
+sleep 2
+docker start "${NODE_NAME}"
+
+kubectl wait --for=condition=Ready node/${NODE_NAME} --timeout=300s
+kubectl rollout status daemonset/${DAEMONSET_NAME} -n ${NAMESPACE} --timeout=300s
+
+source ./test-check-binaries.sh
```
dajrivera, line 26: On reboot the issue is missing multus interfaces, this test is incorrect.

e2e/test-reboot.sh
```diff
@@ -0,0 +1,28 @@
...
+declare -A before_mtime
+
+for bin in $EXPECTED_BINARIES; do
+    before_ts=$(docker exec "${NODE_NAME}" stat -c %Y "${bin}")
+    before_mtime["${bin}"]=$before_ts
+    echo "Before reboot: ${bin} mtime = ${before_ts}"
+done
+
+echo "Rebooting node..."
+docker restart "${NODE_NAME}"
```
dajrivera, line 19: I don't know how quickly kind takes to stop in all containers, however, as I understand, it the reboot issue is with missing multus interfaces due to the multus cni conf file not being available before other cni. If the docker restart kills the multus pods due to the default timeout of 10 seconds, the multus config file will not be cleaned up and will be available on restart thereby never allowing us to hit the missing interfaces issue. To be safe, we should increase the timeout to at least 120 seconds or use -1 to not use a timeout to allow all containers to stop gracefully. 

`docker restart -t 120 "${NODE_NAME}"`

e2e/test-upgrade.sh
```diff
@@ -0,0 +1,33 @@
...
+    echo "Getting mtime for ${bin}..."
+    before_ts=$(docker exec "${NODE_NAME}" stat -c %Y "${bin}")
+    before_mtime["${bin}"]=$before_ts
+    echo "Before reboot: ${bin} mtime = ${before_ts}"
+done
+
+
+# Delete all Multus DaemonSet pods to simulate an upgrade.
+echo "Deleting all Multus DaemonSet pods to simulate upgrade..."
+kubectl delete pods -n ${NAMESPACE} -l name=multus
```
dajrivera, line 25: This is not a good test for upgrades. The issue here is that the file is in use by the daemon during an upgrade. Killing the pods and allowing the daemonset to recreate them will never hit this situation as the file is not in use.

I think it would be better to use the following to try to emulate an upgrade:
`kubectl rollout restart daemonset/${DAEMONSET_NAME} -n ${NAMESPACE} --timeout=300s`

dajrivera (commented):
Added some comments from my experience. Hopefully it helps in figuring this out.

dajrivera:
@dougbtv in case you didn't see this.

PR #1421, opened 2025-04-14
Bumps and makefile

bpickard22:
update to go 1.23 and latest k8s version (1.32.3)
add makefile w simple commands (build and test)

DRA claims are no longer associated with a class as of kube 1.31, so we need to remove the class name ref

cc @dougbtv


bpickard22:
this is too stale, going to start over / get some help upstream on this

Issue #1362, opened 2024-11-27
By default, the path /etc/cni/multus/net.d is not mounted into the Multus daemon


ding-9:
**What happend**:

The default value of the confDir configuration option (/etc/cni/multus/net.d) is not mounted into the Multus daemon when using the installation method:
kubectl apply -f https://raw.githubusercontent.com/k8snetworkplumbingwg/multus-cni/master/deployments/multus-daemonset-thick.yml

**What you expected to happen**:

The path /etc/cni/multus/net.d should be mounted by default into the Multus daemon

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:

- Multus version 
  image path and image ID (from 'docker images')
- Kubernetes version (use `kubectl version`):
- Primary CNI for Kubernetes cluster:
- OS (e.g. from /etc/os-release):
- File of '/etc/cni/net.d/'
- File of '/etc/cni/multus/net.d'
- NetworkAttachment info (use `kubectl get net-attach-def -o yaml`)
- Target pod yaml info (with annotation, use `kubectl get pod <podname> -o yaml`)
- Other log outputs (if you use multus logging)

cyclinder:
Same issue here:

```
[...]

PR #1424, opened 2025-04-27
Thick client: mount multus-conf-dir

nirdothan:
Currently, the default CNI config dir is /etc/cni/multus/net.d [1]. However, pods do not mount this hostPath.
This issue only occurs when a delegate network-attachment-definition has no spec, and it therefore needs to be loaded from disk [2]. For example, when deploying Istio-cni in multus mode, the deployment creates an empty Istio-cni NAD in default namespace, while the actual config is deployed on disk.
CmdAdd fails with the following error message:
GetNetworkDelegates: failed getting the delegate: GetCNIConfig: err in GetCNIConfigFromFile: No networks found in /etc/cni/multus/net.d

[[1]](https://github.com/k8snetworkplumbingwg/multus-cni/blob/v4.2.0/pkg/types/conf.go#L38) [[2]](https://github.com/k8snetworkplumbingwg/multus-cni/blob/v4.2.0/pkg/k8sclient/k8sclient.go#L506)

Fixes issue https://github.com/k8snetworkplumbingwg/multus-cni/issues/1362.

### DIFF
--- a/deployments/multus-daemonset-thick.yml
+++ b/deployments/multus-daemonset-thick.yml
@@ -192,6 +192,8 @@ spec:
             - name: hostroot
               mountPath: /hostroot
               mountPropagation: HostToContainer
+            - mountPath: /etc/cni/multus/net.d
+              name: multus-conf-dir
           env:
             - name: MULTUS_NODE_NAME
               valueFrom:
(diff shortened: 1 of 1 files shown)


dougbtv (approved):
LGTM! Thank you Nir!

PR #1430, opened 2025-06-04

mingshuoqiu:
The original PR is https://github.com/k8snetworkplumbingwg/multus-cni/pull/1232. This PR rebase the commit on top of the current master.


mingshuoqiu:
@dougbtv can you help review the PR? thanks!

PR #1433, opened 2025-06-19
Revert "Suppress error message in cmdDel, in thick plugin"

pliurh:
This reverts commit 5d64ec3367eaa3f066ee3613abdb8eae36bcc7b8.

Multus shall return an error to the caller if a DEL fails.

### DIFF
--- a/pkg/server/api/shim.go
+++ b/pkg/server/api/shim.go
@@ -68,8 +68,7 @@ func CmdCheck(args *skel.CmdArgs) error {
 func CmdDel(args *skel.CmdArgs) error {
 	_, _, err := postRequest(args, CheckAPIReadyNow)
 	if err != nil {
-		// No error in DEL (as of CNI spec)
-		logging.Errorf("CmdDel (shim): %v", err)
+		return logging.Errorf("CmdDel (shim): %v", err)
 	}
 	return nil
 }


s1061123:
>Multus shall return an error to the caller if a DEL fails.

In [CNI Specification](https://github.com/containernetworking/cni.dev/blob/main/content/docs/spec.md#del-remove-container-from-network-or-un-apply-modifications), here is the sentence.

```
Plugins should generally complete a DEL action without error even if some resources are missing. 
```

So CNI specification point of view, CNI plugin should execute `DEL` command without error. Multus CNI is CNI Plugin and it should be alined with specification. So I suppose we should keep it.

If you have something stale resources (e.g. something cannot be cleaned up by `DEL` command), CNI Spec provides `GC` command.
https://github.com/containernetworking/cni.dev/blob/main/content/docs/spec.md#gc-clean-up-any-stale-resources

Multus CNI already supports `GC` command, so I strongly recommend to use `GC` command to cleanup stale resources, something cannot be cleaned up by failed `DEL` command.

s1061123:
@dougbtv PTAL?

danwinship:
> > Multus shall return an error to the caller if a DEL fails.
> 
> In [CNI Specification](https://github.com/containernetworking/cni.dev/blob/main/content/docs/spec.md#del-remove-container-from-network-or-un-apply-modifications), here is the sentence.
> 
> ```
> Plugins should generally complete a DEL action without error even if some resources are missing. 
> ```
> 
> So CNI specification point of view, CNI plugin should execute `DEL` command without error.

That's not what it says. (Or at least, not what it meant.) It just means that a CNI plugin should not return a "not found" error if it is asked to delete something that already doesn't exist. If it is aware that there is something that it needs to delete, and it is unable to delete it, then it's supposed to return an error.

If CNI plugins were never supposed to return errors from DEL, then why would DEL even allow returning an error?

We discussed this in the CNI meeting on 2025-06-16 and @squeed agreed that Multus's interpretation of the spec is wrong.

SchSeba (approved):
LGTM

JensErat:
still relevant

PR #1437, opened 2025-07-15
Connection limit for thick daemon.

juliusmh:
Closes: https://github.com/k8snetworkplumbingwg/multus-cni/issues/1346


juliusmh:
@dougbtv please advise how to progress here.

Issue #1221, opened 2024-02-01
Race issue after node reboot


SchSeba:
Hi, it looks like there is an issue after a node reboot where we can have a race in multus that will prevent the pod from starting

```
kubectl -n kube-system logs -f kube-multus-ds-ml62q -c install-multus-binary
cp: cannot create regular file '/host/opt/cni/bin/multus-shim': Text file busy
```

The problem is mainly after reboot that the multus-shim gets called by crio to start pods but the multus pod is not able to start because the init container fails to cp the shim.
The reason it failed to copy is because crio called the shim who is stuck waiting for the communication with the pod

```
[root@virtual-worker-0 centos]# lsof /opt/cni/bin/multus-shim
COMMAND    PID USER  FD   TYPE DEVICE SIZE/OFF     NODE NAME
multus-sh 8682 root txt    REG  252,1 46760102 46241656 /opt/cni/bin/multus-shim
[root@virtual-worker-0 centos]# ps -ef | grep mult
root        8682     936  0 16:27 ?        00:00:00 /opt/cni/bin/multus-shim
root        9082    7247  0 16:28 pts/0    00:00:00 grep --color=auto mult
```

SchSeba:
```
[root@virtual-worker-0 centos]# ps -ef | grep 942
root         942       1  5 17:07 ?        00:00:00 /usr/bin/crio
root        1246     942  0 17:07 ?        00:00:00 /opt/cni/bin/multus-shim
root        2745    2395  0 17:08 pts/0    00:00:00 grep --color=auto 942
```

SchSeba:
just update doing -f looks like fix the issue in the copy command

rrpolanco:
Coincidentally, we also saw this error crop up yesterday with one of our edge clusters rebooting.

adrianchiris:
As an FYI i see different deployment yamls use different way to copy the cni binary in init container:

the first one will use `install_multus` which will copy files in an atomic manner. the latter will just use `cp`.
(`install_multus` support both thick and thin plugin types)

although im not sure that copying file atomically will solve the above issue.

see:
https://github.com/k8snetworkplumbingwg/multus-cni/blob/8e5060b9a7612044b7bf927365bbdbb8f6cde451/deployments/multus-daemonset.yml#L207

and 

[...]

PR #1445, opened 2025-09-28
Fix node reboot issue by using install_multus bin to update cni file

Untersander:
There is a well known race condition that breaks clusters after node reboot described in #1221.
There PR #1213 that would fix this issue but also introduces another change which breaks the e2e tests.

This PR fixes the race condition and also updates the e2e test to wait a little longer in the setup as I noticed a little flakiness in my own tests.


casell:
Successully applied this in PRODUCTION! Thanks @Untersander!

PR #1450, opened 2025-10-11
Pass delegated plugin conflist also for del/check calls

### LINKED ISSUE #1449: Delegated CNI conflist passed with add, but not check/del
**What happend**:

Kubernetes cluster with switched L2 pod network. Each host has static /24 pod network and uses Calico IPAM in host-local mode.

Newly scheduled pod stuck in ContainerCreating. Logs indicate IP addresses are exhausted. Checking `/var/lib/cni/networks/k8s-pod-network/`, I see all IPs in range seem in used. Only roughly 20 IPs should be assigned though.

Checking multus logs, I see messages like

`2025-10-11 15:25:28.892 [ERROR][38370] plugin.go 580: Final result of CNI DEL was an error. error=invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable`

This is response of the calico CNI plugin. Trying to further understand, I injected a small shell script tracing input/output of the CNI calls and some env variables. Then, I deleted a daemonset pod.

The delete output, observe the third line (input to the Calico CNI plugin) and the error message we already know:

```
CNI_ARGS=IgnoreUnknown=true;K8S_POD_NAMESPACE=backup;K8S_POD_NAME=node-agent-d97p8;K8S_POD_INFRA_CONTAINER_ID=a173930f520b46882cb2eb983ed593
bcaccb65e35a3e305c304aa1bb8f894fea;K8S_POD_UID=774f021f-d09a-4e62-ab2b-add3adb6ca82
CNI_COMMAND=DEL
{"cniVersion":"0.3.1","ipam":{"type":"host-local"},"name":"k8s-pod-network","type":"calico"}
{
    "code": 999,
    "msg": "invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable"
}
```

Checking an add output, we see multus passes the full output of `/etc/cni/net.d/10-calico.conflist`:

```
CNI_ARGS=IgnoreUnknown=true;K8S_POD_NAMESPACE=backup;K8S_POD_NAME=node-agent-d97p8;K8S_POD_INFRA_CONTAINER_ID=a173930f520b46882cb2eb983ed593
bcaccb65e35a3e305c304aa1bb8f894fea;K8S_POD_UID=774f021f-d09a-4e62-ab2b-add3adb6ca82
CNI_COMMAND=DEL
{"cniVersion":"0.3.1","datastore_type":"kubernetes","ipam":{"ranges":[[{"subnet":"usePodCidr"}]],"routes":[{"dst":"0.0.0.0/0"}],"type":"host-local"},"kubernetes":{"kubeconfig":"/etc/cni/net.d/calico-kubeconfig"},"log_file_path":"/var/log/calico/cni/cni.log","log_level":"info","mtu":1440,"name":"k8s-pod-network","nodename":"<redacted>","policy":{"type":"k8s"},"type":"calico"}
{
    "cniVersion": "0.3.1",
    "interfaces": [
        {
            "name": "cali109d890e582"
        }
    ],
    "ips": [
        {
            "version": "4",
            "address": "172.24.15.16/32"
        }
    ],
    "routes": [
        {
            "dst": "0.0.0.0/0"
        }
    ],
    "dns": {}
}
```

Full content of `/etc/cni/net.d/10-calico.conflist` also passed to the calico CNI plugin for del requests, such that these run successfully and IPs are released.

Trace delegated plugin calls.

Checking the code, I see https://github.com/k8snetworkplumbingwg/multus-cni/pull/1419 changed the behavior of the add call. Dropping to neither 4.1.4 nor 4.2.0 which should not have the new code did not resolve the issue though. Porting back the changes to the add function to the del function resolved the issue, see #1450.

I support https://github.com/k8snetworkplumbingwg/multus-cni/pull/1433 -- which was hiding the actual error until the IP pool exhausted.

The cluster ran sucecssfully for some time with a Multus release 4.1.4 based on bitnami's charts. I cannot tell for sure whether this was stable, but must be assumed (the cluster runs a bunch of cronjobs). I assume the bitnami chart was deploying the "thin" version of multus?

**Environment**:

- Multus version 4.2.2
  image path and image ID (from 'docker images')  v4.2.2-thick 
- Kubernetes version (use `kubectl version`): ghcr.io/k8snetworkplumbingwg/multus-cni:v4.2.2-thick
- Primary CNI for Kubernetes cluster: Calico
- OS (e.g. from /etc/os-release): Ubuntu 24.04
- File of '/etc/cni/net.d/'

JensErat:
A delegated plugin that requires configuration fails for del and check calls. Porting back the code added to the add call in #1419 resolves the issue.

Fixes #1449

This is validated in production, and the unit tests do not fail.

Do we need tests for this, what would be the most straight-forward way of adding one? From what I see the fake delegated plugin does not yet check for any passed configuration. Can we assume the configuration should be passed for any delegated CNI calls?

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -287,16 +287,25 @@ func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, cniConfList *lib
 	return result, nil
 }
 
-func conflistCheck(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *types.NetConf, exec invoke.Exec) error {
+func conflistCheck(rt *libcni.RuntimeConf, rawnetconflist []byte, cniConfList *libcni.NetworkConfigList, multusNetconf *types.NetConf, exec invoke.Exec) error {
 	logging.Debugf("conflistCheck: %v, %s", rt, string(rawnetconflist))
 
 	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
 	binDirs = append([]string{multusNetconf.BinDir}, binDirs...)
 	cniNet := libcni.NewCNIConfigWithCacheDir(binDirs, multusNetconf.CNIDir, exec)
 
-	confList, err := libcni.ConfListFromBytes(rawnetconflist)
-	if err != nil {
-		return logging.Errorf("conflistCheck: error converting the raw bytes into a conflist: %v", err)
+	var confList *libcni.NetworkConfigList
+	var err error
+
+	// This may wind up being set during parsing the default network config.
+	// In this case -- we'll use it as passed. Otherwise, we'll recalculate it.
+	if len(cniConfList.Plugins) > 0 {
+		confList = cniConfList
+	} else {
+		confList, err = libcni.NetworkConfFromBytes(rawnetconflist)
+		if err != nil {
+			return logging.Errorf("conflistCheck: error converting the raw bytes into a conflist: %v", err)
+		}
 	}
 
 	err = cniNet.CheckNetworkList(context.Background(), confList, rt)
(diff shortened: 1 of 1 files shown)


JensErat:
still an issue

AlexDouze:
We need this fix too!

Jc2k:
Still need this fix here too.

Whisperish:
Need this fix here too!

FlyGoat:
This is impacting us as well.

wizhaoredhat:
Hi, @JensErat , Does https://github.com/k8snetworkplumbingwg/multus-cni/pull/1521 fix your issue?  For PR 1521, unit tests were added that uses a conflist with plugins (specifically Calico). If PR 1521 suits your needs, I believe we can move this conversation there and resolve this issue.

pkg/multus/multus.go
```diff
@@ -287,16 +287,25 @@ func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, cniConfList *lib
...
 
-	confList, err := libcni.ConfListFromBytes(rawnetconflist)
-	if err != nil {
-		return logging.Errorf("conflistCheck: error converting the raw bytes into a conflist: %v", err)
+	var confList *libcni.NetworkConfigList
+	var err error
+
+	// This may wind up being set during parsing the default network config.
+	// In this case -- we'll use it as passed. Otherwise, we'll recalculate it.
+	if len(cniConfList.Plugins) > 0 {
```
wizhaoredhat, line 302: Although this definitely fixes the reported issue. By reading the logic, there is something definitely wrong. When Plugins (such as IPAMs) are present it uses the cniConfgList but if it is empty then we rebuild the confList from bytes. This makes there being 2 sources of truth which is not good at all.

I am aware this is how "ADD" does this, but let's fix it properly like here: https://github.com/k8snetworkplumbingwg/multus-cni/pull/1521

JensErat:
After updating multus, I could not observe the issue any more for some time now. I guess it's indeed resolved by #1521.

PR #1452, opened 2025-10-13
use `net.JoinHostPort` for proper IPv6 address formatting in kubeconfig

defo89:
While RFC3986 requires providing IPv6 addresses in brackets, having them for IPv4 and DNS names is not required and may sometimes require workarounds on proxy side (e.g. istio).

This will set the brackets for `kubeHost` only if it's an IPv6 address.

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -21,6 +21,7 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/thin_entrypoint/main.go
```diff
@@ -224,8 +225,16 @@ func (o *Options) createKubeConfig(prevCAHash, prevSATokenHash []byte) ([]byte,
 	if err != nil {
 		return nil, nil, fmt.Errorf("template parse error: %v", err)
 	}
+	var kubeConfigHost string
+	if ip := net.ParseIP(kubeHost); ip != nil && ip.To4() == nil {
+		// use brackets only for IPv6 address
+		kubeConfigHost = fmt.Sprintf("%s://[%s]:%s", kubeProtocol, kubeHost, kubePort)
+	} else {
+		kubeConfigHost = fmt.Sprintf("%s://%s:%s", kubeProtocol, kubeHost, kubePort)
+	}
```
brandond: no, just use `net.JoinHostPort(kubeHost, kubePort)`. It handles all this for you.

PR #1453, opened 2025-10-16
[multus] use shortPollTimeout when calling APIserver for pods

olegg-oai:
Existing implementation uses very short timeout (50 or 250ms) when falling back to GetPodAPILiveQuery().
In some environments it results in increased number of `error waiting for pod` errors.
Increasing it to `shortPollTimeout     = 2500 * time.Millisecond`


6ixfalls:
I'm seeing this issue as well, would be great to get this merged

6ixfalls:
not stale

PR #1469, opened 2026-01-01
Add ability to set `cleanupConfigOnExit` in the thick client config

### LINKED ISSUE #1468: Add ability to set `--cleanup-config-on-exit=false` on the thick client

The thin client provides the ability to configure whether the multus CNI config file is removed on exit. However,
the thick client does not provide this, but instead (indirectly) hardcodes it to `--cleanup-config-on-exit=true` here:

https://github.com/k8snetworkplumbingwg/multus-cni/blob/39d6a8ffd2fb2a6f23d6767e1d0613dfc5536d96/pkg/server/config/manager.go#L156-L164

**Why is this needed**:

I need to set `--cleanup-config-on-exit=false` because after a node reboot, my primary CNI starts before multus. This causes pods that specify a multus interface to be started before the multus config file is created, resulting in the multus-specified interfaces to not be added to the pod. To work around this, I need to have custom scripts added to each pod that check whether an interface with a certain name is present and restart (crash) the pod if not.

astro-stan:
Adds ability to set `cleanupConfigOnExit` in the thick client config.

Closes #1468

## Summary by CodeRabbit

* **New Features**
  * Added a configurable option to control whether the old Multus configuration file is removed on exit; enabled by default to keep the system clean.
* **Documentation**
  * New documentation section explaining the cleanup-on-exit option, default behavior, risks of disabling it, and when disabling may be useful.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


astro-stan:
Looking at the test log, I don't think the failing tests are related to my changes.

PR #1470, opened 2026-01-15
Adds support for CNI STATUS + other fixes for CNI Spec 1.1.0

trozet:
Adds support for 1.1.0:

- Adds STATUS handling for delegates.
- Update gateway result handling to include 1.1.0.
- Other places updated that looked for 1.0.0 instead of >=1.0.0.
- Fixes GC to work for single plugin as well.

With these things fixed. Multus should now be good to go with 1.1.0.

### DIFF
--- a/cmd/thin_entrypoint/main.go
+++ b/cmd/thin_entrypoint/main.go
@@ -29,6 +29,7 @@ import (
 	"time"
 
 	"github.com/containernetworking/cni/libcni"
+	cniversion "github.com/containernetworking/cni/pkg/version"
 	"github.com/spf13/pflag"
 
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/cmdutils"
@@ -497,14 +498,14 @@ func (o *Options) createMultusConfig(prevMasterConfigFileHash []byte) (string, [
 		return "", nil, fmt.Errorf("cannot create multus cni temp file: %v", err)
 	}
 
-	// use conflist template if cniVersionConfig == "1.0.0"
+	// use conflist template if cniVersionConfig >= "1.0.0"
 	multusConfFilePath := fmt.Sprintf("%s/00-multus.conf", o.CNIConfDir)
 	templateMultusConfig, err := template.New("multusCNIConfig").Parse(multusConfTemplate)
 	if err != nil {
 		return "", nil, fmt.Errorf("template parse error: %v", err)
 	}
 
-	if o.CNIVersion == "1.0.0" { //Check 1.0.0 or above!
+	if gt, err := cniversion.GreaterThanOrEqualTo(o.CNIVersion, "1.0.0"); err == nil && gt {
 		multusConfFilePath = fmt.Sprintf("%s/00-multus.conflist", o.CNIConfDir)
 		templateMultusConfig, err = template.New("multusCNIConfig").Parse(multusConflistTemplate)
 		if err != nil {
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -764,10 +764,10 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 
 	pod, err := GetPod(kubeClient, k8sArgs, false)
 	if err != nil {
-		if err == errPodNotFound {
-			emptyresult := emptyCNIResult(args, "1.0.0")
-			logging.Verbosef("CmdAdd: Warning: pod [%s/%s] not found, exiting with empty CNI result: %v", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, emptyresult)
-			return emptyresult, nil
+		if stderrors.Is(err, errPodNotFound) {
+			emptyResult := emptyCNIResult(args, n.CNIVersion)
+			logging.Verbosef("CmdAdd: Warning: pod [%s/%s] not found, exiting with empty CNI result: %v", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, emptyResult)
+			return emptyResult, nil
 		}
 		return nil, err
 	}
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -1215,9 +1215,28 @@ func CmdGC(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) err
 	binDirs = append([]string{n.BinDir}, binDirs...)
 	cniNet := libcni.NewCNIConfigWithCacheDir(binDirs, n.CNIDir, exec)
 
-	conf, err := libcni.ConfListFromBytes(n.Delegates[0].Bytes)
-	if err != nil {
-		return logging.Errorf("error in converting the raw bytes to conf: %v", err)
+	delegate := n.Delegates[0]
+	isConfList := delegate.ConfListPlugin
+	if !isConfList && delegate.Conf.Type == "" && delegate.ConfList.Name != "" {
+		isConfList = true
+	}
+
+	var confList *libcni.NetworkConfigList
+	if isConfList {
+		confList, err = libcni.ConfListFromBytes(delegate.Bytes)
+		if err != nil {
+			return logging.Errorf("error in converting the raw bytes to conf: %v", err)
+		}
+	} else {
+		conf, err := libcni.ConfFromBytes(delegate.Bytes)
+		if err != nil {
+			return logging.Errorf("error in converting the raw bytes to conf: %v", err)
+		}
+		confList = &libcni.NetworkConfigList{
+			Name:       conf.Network.Name,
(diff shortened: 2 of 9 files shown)


pkg/server/server.go
```diff
@@ -705,6 +746,25 @@ func (s *Server) cmdDelegateCheck(cmdArgs *skel.CmdArgs, k8sArgs *types.K8sArgs,
...
+		if err != nil {
+			return err
+		}
+		delegateCNIConf = loadedConf
+	} else if delegateCNIConf.Conf.Type == "" && delegateCNIConf.ConfList.Name != "" {
+		delegateCNIConf.ConfListPlugin = true
+	}
+	rt, _ := types.CreateCNIRuntimeConf(cmdArgs, k8sArgs, cmdArgs.IfName, nil, delegateCNIConf)
+	return multus.DelegateStatus(s.exec, delegateCNIConf, rt, multusConfig)
+}
```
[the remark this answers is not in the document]
trozet: will fix

pkg/multus/multus.go
```diff
@@ -456,6 +516,49 @@ func DelegateCheck(exec invoke.Exec, delegateConf *types.DelegateNetConf, rt *li
...
+		if err != nil {
+			var cniErr *cnitypes.Error
+			if stderrors.As(err, &cniErr) {
+				return err
+			}
+			return logging.Errorf("DelegateStatus: error invoking DelegateStatus - %q: %v", delegateConf.Conf.Type, err)
+		}
+	}
+
+	return err
```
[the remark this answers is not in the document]
trozet: will fix

trozet:
@SchSeba do you mind taking a look at this one?

pkg/server/server.go
```diff
@@ -389,6 +392,34 @@ func (s *Server) Start(ctx context.Context, l net.Listener) {
 	}()
 }
 
+func (s *Server) writeCNIErrorResponse(w http.ResponseWriter, err error) {
+	var cniErr *cnitypes.Error
+	if errors.As(err, &cniErr) {
+		w.WriteHeader(http.StatusBadRequest)
+		w.Header().Set("Content-Type", "application/json")
```
bpickard22, line 398: we need to set the header before we call writeHeader right?

bpickard22: otherwise looks good to me, thanks for adding testing too.
@SchSeba we should probably document that we support cni status at cniversion 1.1.0 and up now

pkg/multus/multus.go
```diff
@@ -327,6 +363,30 @@ func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *t
 	return err
 }
 
+func conflistStatus(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *types.NetConf, exec invoke.Exec) error {
```
pliurh: Why not also check the CNI version in this function?

pkg/multus/multus.go
```diff
@@ -258,6 +260,40 @@ func confDel(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.Net
...
+	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
+	binDirs = append([]string{multusNetconf.BinDir}, binDirs...)
+	cniNet := libcni.NewCNIConfigWithCacheDir(binDirs, multusNetconf.CNIDir, exec)
+
+	conf, err := libcni.ConfFromBytes(rawNetconf)
+	if err != nil {
+		return logging.Errorf("error in converting the raw bytes to conf: %v", err)
+	}
+
+	if gt, _ := cniversion.GreaterThanOrEqualTo(conf.Network.CNIVersion, "1.1.0"); !gt {
```
pliurh, line 275: We shouldn't ignore the error here?

pkg/multus/multus.go
pliurh, line 368: We shall also check the `cniVersion` here.

pkg/multus/multus.go
```diff
@@ -456,6 +516,46 @@ func DelegateCheck(exec invoke.Exec, delegateConf *types.DelegateNetConf, rt *li
...
+
+	if err != nil {
+		var cniErr *cnitypes.Error
+		if stderrors.As(err, &cniErr) {
+			return err
+		}
+		if isConfList {
+			return logging.Errorf("DelegateStatus: error invoking ConflistStatus - %q: %v", delegateConf.ConfList.Name, err)
+		}
+		return logging.Errorf("DelegateStatus: error invoking DelegateStatus - %q: %v", delegateConf.Conf.Type, err)
```
pliurh: Should it be `error invoking confStatus`?

pkg/multus/multus.go
```diff
@@ -258,6 +260,40 @@ func confDel(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.Net
...
+	err = cniNet.GetStatusNetworkList(context.Background(), confList)
+	if err != nil {
+		var cniErr *cnitypes.Error
+		if stderrors.As(err, &cniErr) {
+			return err
+		}
+		return logging.Errorf("error in getting result from StatusNetworkList: %v", err)
+	}
+
+	return err
```
pliurh, line 296: Can we just return 'nil'?

pliurh (changes requested):

pkg/multus/multus.go
SchSeba, line 275: maybe we can log here so we know why we skip this

pkg/multus/multus.go
```diff
@@ -327,6 +363,30 @@ func conflistDel(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *t
...
+	logging.Debugf("conflistStatus: %v, %s", rt, string(rawnetconflist))
+
+	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
+	binDirs = append([]string{multusNetconf.BinDir}, binDirs...)
+	cniNet := libcni.NewCNIConfigWithCacheDir(binDirs, multusNetconf.CNIDir, exec)
+
+	confList, err := libcni.ConfListFromBytes(rawnetconflist)
+	if err != nil {
+		return logging.Errorf("conflistStatus: error converting the raw bytes into a conflist: %v", err)
+	}
```
SchSeba, line 378: I think here we should have the same validation as for the single one no? (just in a list)

```
if gt, _ := cniversion.GreaterThanOrEqualTo(conf.Network.CNIVersion, "1.1.0"); !gt {
```

pkg/multus/multus.go
bpickard22: err will always be nil here so thats fine but go best practice is just return err, i dont think it really matters

pkg/multus/multus.go
trozet: I think we should. If the CNI plugin is less than 1.1.0, then it doesn't support status, so it shouldn't return an error.

pkg/multus/multus.go
trozet: conflistStatus calls GetStatusNetworkList, which does the check there in libcni

pkg/multus/multus.go
trozet: it's already done here https://github.com/containernetworking/cni/blob/main/libcni/api.go#L857

trozet:
would it be acceptable to change the deployments -> yaml files to reference 1.1.0? THey currently reference 0.3.1.

bpickard22:
> would it be acceptable to change the deployments -> yaml files to reference 1.1.0? THey currently reference 0.3.1.



> would it be acceptable to change the deployments -> yaml files to reference 1.1.0? THey currently reference 0.3.1.

yeah I think its a good idea. Only possible issue is if a delegate CNI doesnt have support for 1.1.0, I did a quick search and it didnt see any that didnt but one could exist i guess

another option would be to update that in a seperate commit or pr. Regardless once this merges im going to create a new release for multus so if there is an issue they could always use the older version

trozet:
@bpickard22 
OK. From what I understand the thin client derives the version from delegate, while thick plugin uses this config file from the yaml. I think it should be OK to bump the default in master, but I'm also fine deferring to another PR.

bpickard22:
lets leave the default for now and ill work on a solution that more graceful for upgrades, id rather it be an opt in

cmd/thin_entrypoint/main.go
```diff
@@ -497,14 +498,14 @@ func (o *Options) createMultusConfig(prevMasterConfigFileHash []byte) (string, [
 		return "", nil, fmt.Errorf("cannot create multus cni temp file: %v", err)
 	}
 
-	// use conflist template if cniVersionConfig == "1.0.0"
+	// use conflist template if cniVersionConfig >= "1.0.0"
```
SchSeba, line 501: Hi @trozet I think the comment is not in the right place no?
here we have the ConfTemplate and below in the If we use the ConfList no?

SchSeba (commented):
Hi @trozet no critical just a small nit and I saw the PR is already merged :P

PR #1478, opened 2026-03-04
k8sclient: use ServerVersion instead of pod list for kubeconfig validation

pliurh:
Listing all pods across all namespaces during bootstrap is expensive in large clusters and unnecessary since the result is discarded. Use the lightweight /version endpoint to validate connectivity instead.

### DIFF
--- a/pkg/k8sclient/kubeconfig.go
+++ b/pkg/k8sclient/kubeconfig.go
@@ -27,7 +27,6 @@ import (
 
 	certificatesv1 "k8s.io/api/certificates/v1"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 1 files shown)


pkg/k8sclient/kubeconfig.go
```diff
@@ -92,26 +102,24 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+				cfg = newBootstrapKubeconfig
+			} else {
+				logging.Errorf("failed to validate kubeconfig with new certs: %v", err)
 			}
+		}
 
-			if current != nil {
-				cfg = config
-			}
+		if current != nil {
```
bpickard22: this jumps where we set the config outside of where we were previously validating, so previously if we had some error other than unknown authority we would keep the bootstrap kubeconfig instead of the cert based kubeconfig, now we just override the bootstrap either way.

I think this is okay / maybe even an improvement (if we can use the cert based kubeconfig we should), just wanted to make sure this was your intention @pliurh

pliurh: The old implementation is buggy. I did some refactory, PTAL.

PR #1480, opened 2026-03-05
Fix thick plugin STATUS and GC handling for plugin-level commands

jkary:
STATUS and GC are plugin-level commands with no pod context per the
CNI 1.1.0 spec. The thick plugin daemon incorrectly required
CNI_CONTAINERID, CNI_NETNS, and K8S_POD_NAME/K8S_POD_NAMESPACE for
these commands, causing failures when invoked by kubelet.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -434,9 +434,14 @@ func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
 		return nil, fmt.Errorf("could not extract the CNI command args: %w", err)
 	}
 
-	k8sArgs, err := kubernetesRuntimeArgs(cr.Env, s.kubeclient)
-	if err != nil {
-		return nil, fmt.Errorf("could not extract the kubernetes runtime args: %w", err)
+	// STATUS and GC are plugin-level commands with no pod context,
+	// so they don't have K8S_POD_NAME/K8S_POD_NAMESPACE in CNI_ARGS.
+	var k8sArgs *types.K8sArgs
+	if cmdType != "STATUS" && cmdType != "GC" {
+		k8sArgs, err = kubernetesRuntimeArgs(cr.Env, s.kubeclient)
+		if err != nil {
+			return nil, fmt.Errorf("could not extract the kubernetes runtime args: %w", err)
+		}
 	}
 
 	result, err := s.HandleCNIRequest(cmdType, k8sArgs, cniCmdArgs)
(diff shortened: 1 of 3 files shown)


bpickard22:
@jkary thanks for the contribution! The bot has left a pretty good review, take a look and address that and then I will take a look again

jkary:
@bpickard22 thanks for taking a look! I went through the Gemini comments and it looks like the bot reviewed the full branch diff against main rather than just this PR. Most of the comments are about files that aren't changed in this PR at all, like images/entrypoint.sh, Dockerfile.openshift, and cmd/multus-daemon/main.go.

This PR only touches pkg/server/server.go and pkg/server/thick_cni_test.go, so those comments don't apply here.

One thing the bot did catch that is worth fixing is a bug in pkg/multus/multus.go where conflistStatus logs that it will skip STATUS for older CNI versions but doesn't actually return, so it falls through and calls GetStatusNetworkList anyway. I pushed a fix for that.

pkg/server/thick_cni_test.go
```diff
@@ -91,6 +91,69 @@ var _ = Describe(suiteName, func() {
...
+			K8sClient *k8s.ClientInfo
+			ctx       context.Context
+			cancel    context.CancelFunc
+		)
+
+		BeforeEach(func() {
+			var err error
+			K8sClient = fakeK8sClient()
+			// Touch the default network file.
+			os.OpenFile(configPath, os.O_RDONLY|os.O_CREATE, 0755)
```
pliurh, line 108: The return values are ignored. We shall ensure no error returned by this call.

pkg/server/server.go
```diff
@@ -434,9 +434,14 @@ func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
...
+	// so they don't have K8S_POD_NAME/K8S_POD_NAMESPACE in CNI_ARGS.
+	var k8sArgs *types.K8sArgs
+	if cmdType != "STATUS" && cmdType != "GC" {
+		k8sArgs, err = kubernetesRuntimeArgs(cr.Env, s.kubeclient)
+		if err != nil {
+			return nil, fmt.Errorf("could not extract the kubernetes runtime args: %w", err)
+		}
 	}
 
 	result, err := s.HandleCNIRequest(cmdType, k8sArgs, cniCmdArgs)
```
pliurh, line 447: Consider documenting that `k8sArgs` may be nil for STATUS/GC in the `HandleCNIRequest()` signature, or adding a nil guard in it.

pliurh (commented):
Apart from the nits, it looks good to me.

bpickard22:
handled via #1490

PR #1481, opened 2026-03-06
Sort DeviceIDs in GetPodResourceMap for deterministic ordering

cathy-zhou:
When a namespace uses a primary User-Defined Network (UDN) with a device-plugin resource (e.g. SR-IOV), OVN-Kubernetes uses the last device in the list for the primary interface while Multus assigns earlier devices to cluster-default/secondary interfaces. The kubelet and checkpoint paths build the list from map iteration, so order was non-deterministic and the "last" device could differ between callers. Sorting ensures both Multus and OVN-K8s see the same order so the last device is consistently the one reserved for the primary UDN.

### DIFF
--- a/pkg/checkpoint/checkpoint.go
+++ b/pkg/checkpoint/checkpoint.go
@@ -108,5 +108,6 @@ func (cp *checkpoint) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resource
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


girishmg:
LGTM

girishmg:
@bpickard22  can you please help review this MR? Thanks.

bpickard22:
lgtm

PR #1482, opened 2026-03-09
Cert fix

bpickard22:
Multus pods crash-loop when restarted during certificate rotation with the following error:

```
failed to initialize the certificate manager: could not convert data from
"/etc/cni/multus/certs/multus-client-current.pem" into cert/key pair:
tls: failed to find any PEM data in certificate input 
```

The per-node certificate feature (introduced in #1159) uses a certificate store that manages certificates via symlinks:

```
1. Create new cert: multus-client-<timestamp>.pem
2. Remove symlink: multus-client-current.pem          ← RACE WINDOW (~3 seconds)
3. Create new symlink → new certificate file
``` 

When a pod restarts during step 2-3:
- `certificateStore.Current()` attempts to read the empty/missing file
- Returns error: "failed to find any PEM data in certificate input"
- Error propagates to daemon startup → **immediate panic**
- No retry logic exists for transient file read failures

This pr adds exponential backoff retry logic to handle transient certificate errors during daemon startup in `pkg/k8sclient/kubeconfig.go`:

**1. New function `isTransientCertError()`** - Classifies errors:
   - **Transient** (retry): `fs.PathError`, `"failed to find any PEM data"`, `"invalid PEM"`
   - **Permanent** (fail fast): `NoCertKeyError` (expected initial state), RBAC errors

**2. Enhanced `PerNodeK8sClient()`** - Adds retry logic:
   - **Inner retry**: Exponential backoff (100ms, 200ms, 400ms, 800ms, 1.6s = ~3.1s total)
   - **Outer poll**: Existing 2-minute timeout for CSR approval (unchanged)
   - **Logging**: Verbose-level summary on successful retry, debug-level on exhaustion
   - 
- Retries are scoped to certificate loading only, not CSR approval
- Permanent errors still fail immediately (no regression)
- Minimal logging to avoid noise in production
- Total added latency: ~3 seconds worst case, only during active rotation

I tested this change in a local kind cluster and the multus pod will now restart where before it would go into crashLoop


bpickard22:
this branch is based off of downstream, closing and fixing

PR #1483, opened 2026-03-10
Fix certificate rotation race condition in per-node certificates

bpickard22:
Add retry logic to handle transient certificate errors during daemon startup. This prevents crash-loops when pods restart during the certificate rotation window (~3 seconds) where the symlink is temporarily removed or the file is empty.

We accomplish this by retrying if we get a transient error, but still fail fast on permanent errors.

Changes:
- Add isTransientCertError() to classify errors as transient vs permanent
- Transient errors (file not found, empty/invalid PEM, and the "no cert/key present yet" case that's expected during initial provisioning or mid-rotation) are retried within the existing 2-minute CSR approval timeout, using its existing 1-second poll interval
- Other, non-transient errors (e.g. RBAC issues) fail fast instead of polling for the full 2-minute timeout
- Add verbose logging on each transient retry attempt to aid debugging

Testing
- Tested with missing certificate file during pod restart
- Pods successfully recover and start normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved certificate readiness handling by retrying temporary certificate errors during connection setup.
  * Added immediate failure reporting for non-recoverable certificate-store errors.
  * Continued waiting appropriately when certificates are not yet available.

### DIFF
--- a/pkg/k8sclient/kubeconfig.go
+++ b/pkg/k8sclient/kubeconfig.go
@@ -19,7 +19,9 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 	"strings"
@@ -53,6 +55,37 @@ const (
 var (
 	certUsages = []certificatesv1.KeyUsage{certificatesv1.UsageDigitalSignature, certificatesv1.UsageClientAuth}
 )
+
+// isTransientCertError determines if a certificate error is transient (e.g., due to
+// certificate rotation) and should be retried, or if it's a permanent error that
+// requires different handling.
+func isTransientCertError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// NoCertKeyError is expected when cert doesn't exist yet, not transient
+	var noCertErr *certificate.NoCertKeyError
+	if errors.As(err, &noCertErr) {
+		return false
+	}
+
+	// PathError (file not found, permission denied) could be transient during rotation
+	var pathErr *fs.PathError
+	if errors.As(err, &pathErr) {
+		return true
+	}
+
+	// PEM parsing errors indicate empty/corrupted file during rotation
+	errMsg := err.Error()
+	if strings.Contains(errMsg, "failed to find any PEM data") ||
+		strings.Contains(errMsg, "invalid PEM") ||
+		strings.Contains(errMsg, "certificate signed by unknown authority") {
+		return true
+	}
+
+	return false
+}
 
 // getPerNodeKubeconfig creates new kubeConfig, based on bootstrap, with new certDir
 func getPerNodeKubeconfig(bootstrap *rest.Config, certDir string) *rest.Config {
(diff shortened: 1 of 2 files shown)


pkg/k8sclient/kubeconfig.go
```diff
@@ -54,6 +56,37 @@ var (
...
+
+	// NoCertKeyError is expected when cert doesn't exist yet, not transient
+	var noCertErr *certificate.NoCertKeyError
+	if errors.As(err, &noCertErr) {
+		return false
+	}
+
+	// PathError (file not found, permission denied) could be transient during rotation
+	var pathErr *fs.PathError
+	if errors.As(err, &pathErr) {
```
pliurh, line 75: the certificate store's loadFile function https://github.com/pliurh/multus-cni/blob/ddd00fe48bda0bd13902d97ef02b3a2ec4a03c92/vendor/k8s.io/client-go/util/certificate/certificate_store.go#L194 wraps errors with fmt.Errorf("could not convert data from %q into cert/key pair: %v", ...) using %v not %w, which destroys the type information.

pkg/k8sclient/kubeconfig.go
```diff
@@ -148,8 +181,47 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+					logging.Verbosef("Certificate loaded successfully after %d retries (likely cert rotation in progress)", retryCount)
+				}
+				return true, nil
+			}
+			return false, nil
+		})
+
+		if err == nil && retryCount > 0 && currentCert == nil {
+			logging.Debugf("Exhausted %d retries for transient cert error, will retry poll: %v", retryCount, firstTransientErr)
+		}
```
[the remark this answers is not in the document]
pliurh: This needs to be addressed.

pkg/k8sclient/kubeconfig.go
```diff
@@ -148,8 +181,47 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+		backoff := wait.Backoff{
+			Duration: 100 * time.Millisecond,
+			Factor:   2.0,
+			Steps:    5,
+		}
+
+		var retryCount int
+		var firstTransientErr error
+
+		err := wait.ExponentialBackoff(backoff, func() (bool, error) {
```
pliurh, line 197: Is this inner polling really necessary?

pkg/k8sclient/kubeconfig.go
```diff
@@ -148,8 +181,47 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
...
+				return true, nil
+			}
+			return false, nil
+		})
+
+		if err == nil && retryCount > 0 && currentCert == nil {
+			logging.Debugf("Exhausted %d retries for transient cert error, will retry poll: %v", retryCount, firstTransientErr)
+		}
+
+		return currentCert != nil && err == nil, nil
```
pliurh, line 224: The "permanent" errors don't actually fail fast; the outer callback always returns nil as its error.

SchSeba (commented):
@coderabbitai full review

pkg/k8sclient/kubeconfig.go
```diff
@@ -54,6 +55,26 @@ var (
...
+	if err == nil {
+		return false
+	}
+	errMsg := err.Error()
+	var noCertKeyErr *certificate.NoCertKeyError
+	if errors.As(err, &noCertKeyErr) {
+		return false
+	}
+	if strings.Contains(errMsg, "no such file or directory") ||
+		strings.Contains(errMsg, "failed to find any PEM data") ||
```
SchSeba, line 75: suggestion (non-blocking): Consider replacing `strings.Contains(errMsg, "no such file or directory")` with `errors.Is(err, os.ErrNotExist)`. This is the idiomatic Go way to check for file-not-found errors — it handles wrapped errors, is locale-independent, and won't break if the message text ever changes. The other two PEM string checks are fine as-is since there are no typed error equivalents in the standard library.

SchSeba:
@coderabbitai review

pkg/k8sclient/kubeconfig.go
```diff
@@ -54,6 +55,26 @@ var (
 	certUsages = []certificatesv1.KeyUsage{certificatesv1.UsageDigitalSignature, certificatesv1.UsageClientAuth}
 )
 
+// isTransientCertError determines if a certificate error is transient (e.g., due to
+// certificate rotation) and should be retried, or if it's a permanent error that
+// requires different handling.
+func isTransientCertError(err error) bool {
+	if err == nil {
+		return false
+	}
```
SchSeba, line 64: nit (non-blocking): `errMsg := err.Error()` is computed before the `errors.As` check for `NoCertKeyError`. If that branch returns early, the string conversion was wasted. Consider moving the `err.Error()` call below the `errors.As` check — it's a trivial allocation but keeps the function cleaner.

pkg/k8sclient/kubeconfig.go
```diff
@@ -149,7 +170,18 @@ func PerNodeK8sClient(nodeName, bootstrapKubeconfigFile string, certDuration tim
 	err = wait.PollWithContext(context.TODO(), time.Second, 2*time.Minute, func(_ context.Context) (bool, error) {
 		var currentCert *tls.Certificate
 		currentCert, storeErr = certificateStore.Current()
-		return currentCert != nil && storeErr == nil, nil
+		if storeErr != nil {
+			var noCertKeyErr *certificate.NoCertKeyError
+			if errors.As(storeErr, &noCertKeyErr) {
```
SchSeba: we have the same validation inside the function lets remove it from here

pkg/k8sclient/kubeconfig.go
bpickard22: the errors are wrapped with a %v under the hood so we have to use string matching unfortunately. In principal your suggestion would be better it just wont work in this case

thomasferrandiz (approved):

PR #1490, opened 2026-03-16
Fix thick plugin STATUS and GC handling for plugin-level commands

tsorya:
Opening new PR for https://github.com/k8snetworkplumbingwg/multus-cni/pull/1480

STATUS and GC are plugin-level commands with no pod context per the
CNI 1.1.0 spec. The thick plugin daemon incorrectly required
CNI_CONTAINERID, CNI_NETNS, and K8S_POD_NAME/K8S_POD_NAMESPACE for
these commands, causing failures when invoked by kubelet.


trozet (approved):
thanks. Sorry i missed this in my original PR.

trozet:
@bpickard22 can you PTAL?

bpickard22:
lgtm
we also need to fix this for delegate endpoint, its okay if thats in a follow up pr. I will open an issue so that we keep track of it.

Issue #1454, opened 2025-10-22
one pod failed


13567436138:
**What happend**:
```
025-10-22T11:40:18Z [error] conflistDel: error in getting result from DelNetworkList: plugin type="calico" failed (delete): invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
2025-10-22T11:40:18Z [error] DelegateDel: error invoking ConflistDel - "k8s-pod-network": conflistDel: error in getting result from DelNetworkList: plugin type="calico" failed (delete): invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
2025-10-22T11:40:18Z [verbose] DEL finished CNI request ContainerID:"a6de39d2d8ced7b0194dd1f747cc6d6a677a88e3602ddf52ccaeec0619fc0b45" Netns:"/var/run/netns/isulacni-abee149abc8b9317" IfName:"eth0" Args:"IgnoreUnknown=1;K8S_POD_INFRA_CONTAINER_ID=a6de39d2d8ced7b0194dd1f747cc6d6a677a88e3602ddf52ccaeec0619fc0b45;K8S_POD_NAME=virt-handler-rhwng;K8S_POD_NAMESPACE=kubevirt;K8S_POD_UID=" Path:"", result: "", err: DelegateDel: error invoking ConflistDel - "k8s-pod-network": conflistDel: error in getting result from DelNetworkList: plugin type="calico" failed (delete): invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

```

[root@kamaji-manager-01-a ~]# kubectl get pod -n kube-system
NAME                                          READY   STATUS    RESTARTS      AGE
coredns-857d9ff4c9-77zlc                      1/1     Running   6 (34m ago)   5d1h
coredns-857d9ff4c9-xxwgc                      1/1     Running   6 (34m ago)   5d1h
etcd-kamaji-manager-01-a                      1/1     Running   6 (34m ago)   5d1h
kube-apiserver-kamaji-manager-01-a            1/1     Running   6 (34m ago)   5d1h
kube-controller-manager-kamaji-manager-01-a   1/1     Running   6 (34m ago)   5d1h
kube-multus-ds-7v9nj                          1/1     Running   1 (34m ago)   49m
kube-multus-ds-9dfdf                          0/1     Error     0             49m
[...]

PR #1498, opened 2026-03-31
fix: return success on DEL when CNI cache is missing

kvaps:
## Summary
- When CNI ADD never completes (e.g. sandbox creation fails), no cache file is written
- On cleanup, CmdDel reconstructs delegate configs from the API and calls DEL on each plugin
- Delegate plugins fail because they require runtime state from a successful ADD
- This creates a deadlock: containerd cannot release sandbox name reservations
- Fix: return success immediately when cache file does not exist (ADD never completed)

## Problem
After a node disruption, containerd accumulates stale sandbox name reservations. When it tries
to clean them up, CNI DEL is called through multus. Without a cache file, multus reconstructs
the config from the Kubernetes API, but delegate plugins (e.g. kube-ovn) reject the config
because it lacks runtime state (socket paths, cached results) that only exists after ADD.

This causes:
1. Sandbox name reservations persist indefinitely
2. New pods cannot be created on the affected node
3. DEL retry storm overloads multus, causing ADD timeouts for new pods

## Fix
If the cache file does not exist (`os.IsNotExist`), return `nil` immediately since there is
nothing to clean up. This matches the existing graceful fallback when both cache and pod are
missing (the `else` branch at the former line 978).

## Related issues
- #1454
- #1446
- https://github.com/containerd/containerd/issues/11504

## Test plan
- [ ] Verify existing unit tests pass
- [ ] Manual test: create a pod sandbox failure scenario, verify DEL returns success without cache

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -954,33 +954,19 @@ func CmdDel(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) er
 	}
 
 	if !useCacheConf {
-		// Fetch delegates again if cache is not exist and pod info can be read
-		if os.IsNotExist(err) && pod != nil {
-			if in.ClusterNetwork != "" {
-				_, err = k8s.GetDefaultNetworks(pod, in, kubeClient, nil)
-				if err != nil {
-					return cmdErr(k8sArgs, "failed to get clusterNetwork/defaultNetworks: %v", err)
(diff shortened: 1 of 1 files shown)


mattia-eleuteri:
We had the same issue specified [here](https://github.com/cozystack/cozystack/issues/2310)
and I can confirm that the fix proposed in this MR resolved our issue.

PR #1502, opened 2026-04-17
e2e: Replace EOL CentOS 8 with CentOS Stream 9 in test templates

### LINKED ISSUE #1500: e2e: Replace EOL CentOS 8 image in test templates
## Problem

Several e2e test templates use `quay.io/centos/centos:8` (or previously `centos:8`, being fixed in #1472). CentOS 8 reached end-of-life on December 31, 2021. The image is no longer receiving updates and poses several risks:

1. **Image removal risk** -- `quay.io/centos/centos:8` could be removed from the registry at any time, causing all tests using it to fail with image pull errors.
2. **Known CVEs** -- The image contains unpatched vulnerabilities. CI environments with security scanning may flag or block it.
3. **Stale base** -- Built on glibc 2.28 and 2019-era kernel headers.

### Affected files

- `e2e/templates/default-route1.yml.j2` (2 references)
- `e2e/templates/simple-macvlan1.yml.j2` (2 references)
- `e2e/templates/simple-pod.yml.j2` (1 reference)

All of these pods just run `/bin/sleep 10000` or simple network commands -- they don't depend on CentOS-specific functionality.

## Suggested fix

Replace `quay.io/centos/centos:8` with one of:

- **`quay.io/centos/centos:stream9`** -- Drop-in replacement, actively maintained, lowest friction.
- **`docker.io/library/alpine:<pinned version>`** -- Smallest image, already used in `cni-install.yml.j2`, would unify all test pods on one base.

`centos:stream9` is the simplest path since it requires no other changes (same package manager, same paths, same shell).

## Related

- #1472 -- Fixes the short-name issue for these same templates but retains the EOL image.

ashokpariya0:
Replace quay.io/centos/centos:8 with quay.io/centos/centos:stream9 in e2e test templates to address end-of-life concerns.

Fixes #1500

**Note to reviewers:**
also quay.io/centos/centos:8 does not support the s390x architecture, which leads to failures on s390x arch.
https://quay.io/repository/centos/centos?tab=tags

### DIFF
--- a/e2e/templates/default-route1.yml.j2
+++ b/e2e/templates/default-route1.yml.j2
@@ -32,7 +32,7 @@ metadata:
 spec:
   containers:
   - name: default-route-worker1
-    image: quay.io/centos/centos:8
+    image: quay.io/centos/centos:stream9
     command: ["/bin/sleep", "10000"]
     securityContext:
       privileged: true
@@ -51,7 +51,7 @@ metadata:
 spec:
   containers:
   - name: default-route-worker2
-    image: quay.io/centos/centos:8
+    image: quay.io/centos/centos:stream9
     command: ["/bin/sleep", "10000"]
     securityContext:
       privileged: true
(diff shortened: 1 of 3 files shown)


ashokpariya0:
Please review this PR @bpickard22

PR #1504, opened 2026-04-22
kubeletclient: copy dev.DeviceIds to stop SortDeviceIDs scrambling cached state

### LINKED ISSUE #1495: kubeletclient: getDevicePluginResources aliases dev.DeviceIds causing SortDeviceIDs to mutate cached kubelet state
In `pkg/kubeletclient/kubeletclient.go`, when a new resource name is first seen, `getDevicePluginResources` stores the slice directly without copying:

```go
resourceMap[dev.ResourceName] = &types.ResourceInfo{DeviceIDs: dev.DeviceIds}
```

`dev.DeviceIds` is a slice backed by the same underlying array as the data in `rc.resources` (the cached kubelet API response). When `GetPodResourceMap` subsequently calls `types.SortDeviceIDs(resourceMap)`, the in-place `sort.Strings` call mutates that backing array, silently reordering the data stored in `rc.resources`.

## Impact

- A read-only operation (`GetPodResourceMap`) has a hidden write side-effect on cached state.
- Under concurrent `GetPodResourceMap` calls, this is a data race on the shared backing array.

## Fix

Copy the slice when inserting into `resourceMap`:

```go
// in getDevicePluginResources
resourceMap[dev.ResourceName] = &types.ResourceInfo{
    DeviceIDs: append([]string(nil), dev.DeviceIds...),
}
```

The same pattern should be verified in `getDRAResources`, though `deviceIDs` there is built locally so it is not aliased.

SAY-5:
## Summary

Fixes #1495.

`getDevicePluginResources` stored the kubelet-owned `dev.DeviceIds` slice directly in `resourceMap`. When `GetPodResourceMap` later called `types.SortDeviceIDs(resourceMap)`, `sort.Strings` mutated the slice in place; that slice shares its backing array with `rc.resources` (the cached `ListPodResourcesResponse`). So a read-only-looking `GetPodResourceMap` call ended up reordering the cache, and concurrent callers raced on that same backing array.

## Fix

Copy `dev.DeviceIds` into a fresh slice on first insert so `SortDeviceIDs` acts on map-local state. The append-into-existing branch already owns the slice (built with `append` above), and `getDRAResources` already builds `deviceIDs` locally, so neither needs changes.

## Tests

Adds `TestGetDevicePluginResourcesDoesNotAliasDeviceIds` (plain `go test`, avoids the Ginkgo `BeforeSuite` filesystem setup that is macOS-unfriendly). It pre-populates a `ContainerDevices.DeviceIds` slice with a deliberately unsorted order, runs `getDevicePluginResources` and `SortDeviceIDs`, then asserts the input slice is untouched.

- **On master:** test FAILs, `dev-b dev-a dev-c` becomes `dev-a dev-b dev-c` in the kubelet-side slice.
- **With this fix:** test PASSes and the stored resourceMap slice is the only thing sorted.

`go build ./pkg/kubeletclient/...` and `go vet ./pkg/kubeletclient/...` are clean.

### DIFF
--- a/pkg/kubeletclient/kubeletclient.go
+++ b/pkg/kubeletclient/kubeletclient.go
@@ -152,7 +152,14 @@ func (rc *kubeletClient) getDevicePluginResources(devices []*podresourcesapi.Con
 		if rInfo, ok := resourceMap[dev.ResourceName]; ok {
 			rInfo.DeviceIDs = append(rInfo.DeviceIDs, dev.DeviceIds...)
 		} else {
-			resourceMap[dev.ResourceName] = &types.ResourceInfo{DeviceIDs: dev.DeviceIds}
+			// Copy dev.DeviceIds into a fresh slice. Storing the
+			// kubelet-owned slice directly lets SortDeviceIDs'
+			// in-place sort mutate the cached rc.resources backing
+			// array, which is a data race under concurrent
+			// GetPodResourceMap calls (#1495).
+			resourceMap[dev.ResourceName] = &types.ResourceInfo{
+				DeviceIDs: append([]string(nil), dev.DeviceIds...),
+			}
 		}
 	}
 }
--- a/pkg/kubeletclient/kubeletclient.go
+++ b/pkg/kubeletclient/kubeletclient.go
@@ -149,18 +149,17 @@ func (rc *kubeletClient) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resou
 
 func (rc *kubeletClient) getDevicePluginResources(devices []*podresourcesapi.ContainerDevices, resourceMap map[string]*types.ResourceInfo) {
 	for _, dev := range devices {
-		if rInfo, ok := resourceMap[dev.ResourceName]; ok {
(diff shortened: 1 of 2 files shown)


SAY-5:
Thanks @gemini-code-assist — applied both suggestions:

1. Collapsed the ok/else branches; appending into a nil `DeviceIDs` on first insert still produces a fresh backing array, so the alias guarantee holds with less duplication.
2. Extended `TestGetDevicePluginResourcesDoesNotAliasDeviceIds` to also assert the map-owned slice is sorted (`dev-a`, `dev-b`, `dev-c`) after `SortDeviceIDs`, so the test covers both the alias regression and the sort behaviour in one shot.

SAY-5:
Thanks @gemini-code-assist — the code in this PR already matches the suggested pattern (ensure `ResourceInfo` exists in the map, then a single `append(rInfo.DeviceIDs, dev.DeviceIds...)`); and the regression test at `kubeletclient_internal_test.go:50-61` already asserts that `resourceMap["vendor.example.com/gpu"].DeviceIDs` contains the sorted `{dev-a, dev-b, dev-c}` — exactly the extra check you proposed. Nothing to change.

PR #1505, opened 2026-04-22
pkg/multus: downcast empty CNI result to requested cniVersion

### LINKED ISSUE #1497: Panic could generate with this change when Multus is configured with cniVersion 0.3.1 or 0.4.0
https://github.com/k8snetworkplumbingwg/multus-cni/blob/132c5e7e79fb80993ec8ee1ec11ecf0bdd7b1e26/pkg/multus/multus.go#L769

In case that mulus is configured with cniVersion: 0.3.1 or 0.4.0, this will create a *types100.Result with CNIVersion 0.31 or 0.4.0. Eventually it leads to [github.com/containernetworking/cni/pkg/types/100.convertFrom04x](http://github.com/containernetworking/cni/pkg/types/100.convertFrom04x) throws out a panic when trying to cast *types100.Result to *types040.Result.

The issue was fixed by https://github.com/k8snetworkplumbingwg/multus-cni/commit/ccfd8f5feae846aa32697b90ba7bc12cb8893c9d . But later commit https://github.com/k8snetworkplumbingwg/multus-cni/commit/921191dece729c8f25f6aaad51f7f31b29677a4a reverts the fix.

SAY-5:
## Summary

Fixes #1497.

`emptyCNIResult` always returns a `*cni100.Result`. When the pod-not-found branch set `CNIVersion` to `0.3.1` / `0.4.0` and handed that to skel, the containernetworking/cni conversion chain dispatched to `convertFrom04x`, which type-asserts the incoming `types.Result` to `*types040.Result` and panics because we gave it a `*cni100.Result`.

The earlier fix [ccfd8f5](https://github.com/k8snetworkplumbingwg/multus-cni/commit/ccfd8f5feae846aa32697b90ba7bc12cb8893c9d) hardcoded `1.0.0` to paper over the class mismatch; [921191d](https://github.com/k8snetworkplumbingwg/multus-cni/commit/921191dece729c8f25f6aaad51f7f31b29677a4a) reverted that so the response would round-trip through the user's configured version again. The panic regressed along with the revert.

## Fix

Call `emptyResult.GetAsVersion(n.CNIVersion)` so the returned Result's concrete type matches whatever `convertFrom*`/`convertTo*` the caller will run:

- For `1.x` configs the Result stays `*cni100.Result` (existing `1.1.0` test still asserts on that type).
- For `0.3.x` / `0.4.0` configs it now returns a `*types040.Result` that `convertFrom04x` can safely assert.

If `n.CNIVersion` is unset, we keep the existing behaviour (return `*cni100.Result` as-is).

## Tests

Adds `It("returns empty add result downcast to 0.4.0 cniVersion when pod is not found", ...)` that drives pod-not-found through `CmdAdd` with `cniVersion: 0.4.0` and asserts the returned Result is a `*types040.Result` with `CNIVersion "0.4.0"`. `go vet ./pkg/multus/...` clean under `GOOS=linux`; the existing `1.1.0` variant stays green.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -768,12 +768,13 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 		if stderrors.Is(err, errPodNotFound) {
 			emptyResult := emptyCNIResult(args, n.CNIVersion)
 			logging.Verbosef("CmdAdd: Warning: pod [%s/%s] not found, exiting with empty CNI result: %v", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, emptyResult)
-			// Downcast to the types package matching n.CNIVersion so skel's
-			// response conversion path calls the right convertFrom*/To*
-			// converter. Returning a *cni100.Result with CNIVersion 0.3.1
-			// trips convertFrom04x's type assertion to *types040.Result
-			// and panics (#1497).
-			if n.CNIVersion != "" {
+			// emptyCNIResult already returns a *cni100.Result, so 1.x
+			// configs don't need conversion. For older cniVersions
+			// (0.3.x / 0.4.0) we must downcast so skel's response
+			// conversion path calls the right convertFrom*/To* converter;
+			// otherwise convertFrom04x's type assertion to *types040.Result
+			// panics (#1497).
+			if n.CNIVersion != "" && !strings.HasPrefix(n.CNIVersion, "1.") {
 				versioned, convErr := emptyResult.GetAsVersion(n.CNIVersion)
 				if convErr != nil {
 					return nil, cmdErr(k8sArgs, "failed to downcast empty CNI result to cniVersion %q: %v", n.CNIVersion, convErr)
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -768,6 +768,18 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 		if stderrors.Is(err, errPodNotFound) {
 			emptyResult := emptyCNIResult(args, n.CNIVersion)
 			logging.Verbosef("CmdAdd: Warning: pod [%s/%s] not found, exiting with empty CNI result: %v", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, emptyResult)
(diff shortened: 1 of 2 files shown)


SAY-5:
Thanks @gemini-code-assist — good call. Gated the `GetAsVersion` conversion on `!strings.HasPrefix(n.CNIVersion, "1.")` so 1.x configs (where `emptyCNIResult` already returns a `*cni100.Result`) skip the round-trip. 0.3.x / 0.4.0 callers still downcast to avoid the `convertFrom04x` panic.

SAY-5:
Thanks @gemini-code-assist — applied your suggestion. GetAsVersion is now gated on `!strings.HasPrefix(n.CNIVersion, "1.")` so the CNIVersion 1.x path skips the redundant round-trip while the pre-1.0 path keeps the panic fix.

pkg/multus/multus.go
```diff
@@ -768,6 +768,19 @@ func CmdAdd(args *skel.CmdArgs, exec invoke.Exec, kubeClient *k8s.ClientInfo) (c
 		if stderrors.Is(err, errPodNotFound) {
 			emptyResult := emptyCNIResult(args, n.CNIVersion)
 			logging.Verbosef("CmdAdd: Warning: pod [%s/%s] not found, exiting with empty CNI result: %v", k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_NAME, emptyResult)
+			// emptyCNIResult already returns a *cni100.Result, so 1.x
+			// configs don't need conversion. For older cniVersions
+			// (0.3.x / 0.4.0) we must downcast so skel's response
+			// conversion path calls the right convertFrom*/To* converter;
+			// otherwise convertFrom04x's type assertion to *types040.Result
+			// panics (#1497).
+			if n.CNIVersion != "" && !strings.HasPrefix(n.CNIVersion, "1.") {
```
trozet: This check is not ideal because:
1. it is not future proof, if cni 2.0 comes along this breaks
2. it can produce a false positive if for example the version was 0.1.0

I would suggest using the functions already exposed in the CNI library:
GreaterThanOrEqualTo

https://github.com/containernetworking/cni/blob/7c270076995b6a35f4774ce94dafcf266d1c6925/pkg/version/plugin.go#L123

trozet (changes requested):
should also squash the 2 commits into 1

SAY-5:
Still current from my side, nothing pending. Single commit, rebased on master, and both of trozet's points are in (cniversion.GreaterThanOrEqualTo instead of the string prefix check, and the commits squashed). Just waiting on a maintainer review.

PR #1398, opened 2025-02-27
Only wait for config waitGroup when configManager is used

chdxD1:
fixes #1366

When `MultusConfigFile` is not set to `auto` (configManager is nil) the WaitGroup is initialised and then waited for, immediately exiting Multus. This PR only waits when configManager is not nil, otherwise it waits for the context to complete (which happens from the SIGINT / SIGTERM handlers)

PR #1508, opened 2026-05-06
Only wait for config waitGroup when configManager is used

chdxD1:
reopen of https://github.com/k8snetworkplumbingwg/multus-cni/pull/1398 with rebase on master

When `MultusConfigFile` is not set to `auto` (configManager is nil) the WaitGroup is initialised and then waited for, immediately exiting Multus. This PR only waits when configManager is not nil, otherwise it waits for the context to complete (which happens from the SIGINT / SIGTERM handlers)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon shutdown handling to ensure ongoing operations finish cleanly before the process exits.

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -139,7 +139,9 @@ func main() {
 		}
 	}
 
+	<-ctx.Done()
 	wg.Wait()
+
 	logging.Verbosef("multus daemon is exited")
 }
 


thomasferrandiz:
@chdxD1 can you follow-up on gemini's code review please?

chdxD1:
> @chdxD1 can you follow-up on gemini's code review please?

@thomasferrandiz It greatly simplifies the change and reduces the branch, changed.

junkb:
i happened to run into this same behavior, and submitted #1543 [now closed], which was pretty much identical to this one.

i'm using on a cri-o cluster: the daemon stays up, cni add and del work
through flannel and macvlan delegates, and sigterm exits cleanly.

PR #1509, opened 2026-05-18
Add optional pprof endpoint support to multus-daemon

bpickard22:
Enable multus-daemon to expose Go pprof endpoints for runtime profiling, allowing operators to capture CPU profiles, memory heap dumps, goroutine stacks, and execution traces. This helps identify performance hot spots and memory issues in production.

pprof is gated behind a new "enablePprof" config flag and shares the existing metrics TCP port, so it is not exposed by default.

Daemon config example:

    {
        "metricsPort": 6061,
        "enablePprof": true
    }

Usage (with metricsPort 6061):

    # 30-second CPU profile
    go tool pprof http://<node>:6061/debug/pprof/profile?seconds=30

    # Memory heap dump
    go tool pprof http://<node>:6061/debug/pprof/heap

    # Goroutine stacks
    curl http://<node>:6061/debug/pprof/goroutine?debug=2

    # Execution trace (5 seconds)
    curl -o trace.out http://<node>:6061/debug/pprof/trace?seconds=5
    go tool trace trace.out

Assisted by Claude Opus 4.6

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -22,12 +22,14 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
 	"os/user"
 	"path/filepath"
 	"sync"
 	"syscall"
+	"time"
 
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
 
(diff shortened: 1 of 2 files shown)


thomasferrandiz (approved):

bpickard22:
thanks for review @thomasferrandiz

PR #1520, opened 2026-06-29
Add resourceName field to the NetworkSelectionElement

### LINKED ISSUE #1515: Allow resourceName selection on a NetworkSelectionElement level instead of NAD only

`resourceName` as a NetworkSelectionElement field, to allow different resource types to be connected to the same network, like so
`"k8s.v1.cni.cncf.io/networks": [{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}]`.

**Why is this needed**:
Currently we allow selecting a resource that should be used for NAD attachment with the NAD annotation `k8s.v1.cni.cncf.io/resourceName`. That means that if I want to allow different devices to be connected to the same network, I have to create a copy of a NAD for each device type, and those NADs are identical except for that annotation value. This is cumbersome, especially if we build APIs on top.
Since network doesn't actually change, but only the pod-side device type changes, it may be reasonable to allow setting this attribute on the NSE instead of NAD.

On the implementation side, it is one new filed and a tiny code change because we only get resourceName once here https://github.com/k8snetworkplumbingwg/multus-cni/blob/db30741512795c258b3ba383a571a66850e8baac/pkg/k8sclient/k8sclient.go#L306 so we could get it from the NSE there too.
In case NAD and NSE both request `resourceName` that are different, we can report an error, but it is a new field and all the previous configurations will work exactly as before.

npinaeva:
The resource pool a network attachment should use is normally selected with the `k8s.v1.cni.cncf.io/resourceName` annotation on the NetworkAttachmentDefinition.
The same selection can instead be set per attachment in the network selection element by adding `"resourceName": "<resource>"`. This lets attachments backed by different resource pools reference the same NetworkAttachmentDefinition, without creating a separate NetworkAttachmentDefinition per resource pool.
NAD still should set alowed resourceNames, which is similar to
creating multiple NADs with different resourceNames.

Fixes #1515 

I am not sure if I should also update https://github.com/k8snetworkplumbingwg/network-attachment-definition-client and in which order.

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting a resource pool per network entry in pod network annotations.
  * Added support for restricting selectable resource pools with an allowlist.
  * Documented per-network resource selection, allowlists, and pod launch examples.

* **Bug Fixes**
  * Resource selection conflicts now return an error and warning event.
  * Requests for pools outside the configured allowlist are rejected with a warning event.
  * Existing behavior is preserved when requested and configured resource values match.

### DIFF
--- a/docs/how-to-use.md
+++ b/docs/how-to-use.md
@@ -431,6 +431,40 @@ spec:
 EOF
 ```
 
+#### Launch pod with json annotation with resourceName
+
+The resource pool a network attachment should use is normally selected with the
+`k8s.v1.cni.cncf.io/resourceName` annotation on the NetworkAttachmentDefinition.
+The same selection can instead be set per attachment in the network selection
+element by adding `"resourceName": "<resource>"`. This lets attachments backed by
+different resource pools reference the same NetworkAttachmentDefinition, without
+creating a separate NetworkAttachmentDefinition per resource pool.
+
+```
+# Execute following command at Kubernetes master
+cat <<EOF | kubectl create -f -
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-case-07
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[
+            { "name" : "sriov-net",
+              "resourceName": "intel.com/sriov" }
+    ]'
+spec:
+  containers:
+  - name: pod-case-07
+    image: docker.io/centos/tools:latest
+    command:
+    - /sbin/init
+EOF
+```
+
+If the NetworkAttachmentDefinition also carries a `k8s.v1.cni.cncf.io/resourceName`
+annotation and it differs from the `resourceName` set in the network selection
+element, Multus reports an error. Setting the same value in both places is allowed.
+
 ### Verifying pod network
 
 Following the example of `ip -d address` output of above pod, "pod-case-06":
--- a/pkg/k8sclient/k8sclient.go
+++ b/pkg/k8sclient/k8sclient.go
@@ -301,12 +301,26 @@ func getKubernetesDelegate(client *ClientInfo, net *types.NetworkSelectionElemen
 		return nil, resourceMap, logging.Errorf("getKubernetesDelegate: %s", errMsg)
 	}
 
-	// Get resourceName annotation from NetworkAttachmentDefinition
+	// Get resourceName from the NetworkAttachmentDefinition annotation. It may
+	// also be set per-attachment by the NetworkSelectionElement, which lets
+	// different device pools attach to the same network without duplicating the
+	// NAD. If both are set and disagree, that is a misconfiguration.
 	deviceID := ""
-	resourceName, ok := customResource.GetAnnotations()[resourceNameAnnot]
-	if ok && pod != nil && pod.Name != "" && pod.Namespace != "" {
-		// ResourceName annotation is found; try to get device info from resourceMap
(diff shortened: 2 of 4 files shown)


npinaeva:
cc @SchSeba I think this should be ready for review :)

docs/how-to-use.md
```diff
@@ -431,6 +431,40 @@ spec:
...
+# Execute following command at Kubernetes master
+cat <<EOF | kubectl create -f -
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pod-case-07
+  annotations:
+    k8s.v1.cni.cncf.io/networks: '[
+            { "name" : "sriov-net",
+              "resourceName": "intel.com/sriov" }
```
trozet, line 453: does this violate any security concerns that having the resourceName on the NAD protected against? For example, if the NAD did not have resourceA, then previously I could not use resourceA while attempting to attach to the NAD, but with this now I can?

Should the NAD still need to contain the allowed resourceNames to have some level of enforcement? Or  was that type of security never the goal in the first place @SchSeba ?

npinaeva: This is kind of a new feature that allows different resourceNames to be connected to the same NAD.
If we want to add network-side limitation on which resourceNames are allowed to be connected, we can add another annotation on the NAD, e.g. "k8s.v1.cni.cncf.io/allowedResourceNames"
We could also ensure security constraints on a higher-level APIs, like UDN, to not overload NAD itself

npinaeva: I update the code with the `k8s.v1.cni.cncf.io/allowedResourceNames` annotation

docs/how-to-use.md
```diff
@@ -431,6 +431,79 @@ spec:
...
+
+```
+# Execute following command at Kubernetes master
+cat <<EOF | kubectl create -f -
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  name: sriov-net
+  annotations:
+    k8s.v1.cni.cncf.io/allowedResourceNames: intel.com/sriov,nvidia.com/sriov
```
SchSeba, line 454: I am trying to think should this be a list or maybe we want an annotation for every network instead, I am not sure what is the right way to model this.

if we have one for every allow resource I think it will allow to patch without a risk of race if two controllers try to append remove from the list.

another point now you will need a controller to append/remove this annotation, is it not the same as creating new net-attach-def?

trying to understand the use-case and how much this will make it easier instead of creating new NAD's


let me know what do you think.

npinaeva: The difference is mostly on the API level. NADs are supposed to represent networks, and current resourceName lets us select the device type for the whole network. Now if I want to allow multiple device types to connect to the same network, I shouldn't need to create new networks (which are actually the same network) for each device type. Instead I can simply say this is a list of devices I allow, and it is less confusing than having copies of the same network with different names.
The original use case is coming from the ovn-kubernetes UserDefinedNetwork implementation, which translates into NADs, and didn't seem optimal to translate a single network into a bunch of NADs just because of the resourceName. A single NAD (e.g. vs 3 NADs) requires much less space in etcd and on the informer side, and doesn't generate unneeded events for the same network split among multiple NADs, which will be a problem at scale.

Now to the list vs annotation per network, I think from the api point of view, a list makes more sense. From the user point of view, I expect a single NAD to be most of the time changed by a single controller/person. If a conflict resolution among multiple users is ever needed (not sure if that use case even exists), it can be easily implemented with the resourceVersion guard. But we can discuss other formats if you see some other potential disadvantages

PR #1521, opened 2026-06-30
fix(types): preserve conflist delegate config so DEL does not leak IPs

chent1996:
## Summary

This PR fixes an **IP leak** that occurs when a Multus delegate is a CNI **conflist** (`plugins: [...]`). On `CMD_DEL`, the delegate was invoked with a *lossy* copy of its configuration, so IPAM plugins that need CNI-specific fields (e.g. Calico, which needs `kubeconfig`/`datastore_type` to reach its datastore) failed to release the allocated IP. Over time, pod churn exhausts the IP pool.

## Symptom

In a cluster using Calico as a Multus delegate (conflist form), we observed that IPs allocated on pod creation were **never released on pod deletion**. The delegate's `DEL` ran against an incomplete config, Calico IPAM could not authenticate to its datastore, and the IP allocation was orphaned. The pool slowly drained until new pods could no longer get an address.

## Root cause

When loading a conflist delegate, `LoadDelegateNetConfFromConfList` stored `DelegateNetConf.Bytes` by marshaling the **structured** `types.NetConfList` (built from `cnitypes.PluginConf`):

```go
pluginsBytes, err := json.Marshal(netConfList)
```

`cnitypes.PluginConf` only models a fixed subset of keys (`cniVersion`/`name`/`type`/`capabilities`/`ipam.type`/`dns`). Any other field — Calico's `kubeconfig`, `datastore_type`, and similar plugin-specific options — is **silently dropped** during that round-trip.

The same lossy marshal happened again on the DEL path in `CmdDel`, where the code backfilled `cniVersion` with:

```go
v.Bytes, err = json.Marshal(v.ConfList)
```

So the config handed to the delegate's `DEL` was missing exactly the fields the plugin needs to clean up its IPAM state — hence the leak. (`ADD` happened to work in many setups because the on-disk conf was read more directly, masking the asymmetry.)

## Fix

1. **Preserve the original conflist bytes losslessly.** `LoadDelegateNetConfFromConfList` now rebuilds `DelegateNetConf.Bytes` from the libcni raw bytes via a new `rawConfListBytes` helper: it takes the list-level keys from `confList.Bytes` and rebuilds the `plugins` array from each plugin's own raw bytes. Per-plugin raw bytes are used (rather than `confList.Bytes` as a whole) because libcni appends plugins loaded from a subdirectory chain into `confList.Plugins` without updating `confList.Bytes`; relying on `confList.Bytes` alone would drop those appended plugins.

2. **Backfill `cniVersion` without data loss on DEL.** `CmdDel` now uses a new `InjectCNIVersionInConfList` helper that unmarshals the raw conflist into a generic map, sets only `cniVersion`, and re-marshals — instead of marshaling the lossy structured `NetConfList`. DEL now receives the same complete config that ADD did.

3. **Remove the now-redundant representation.** The `CNINetworkConfigList libcni.NetworkConfigList` field on `DelegateNetConf` is dropped, and `conflistAdd` is simplified to always parse from the (now-correct) raw bytes. This unifies ADD and DEL on a single source of truth — the lossless `Bytes` — and removes the duplicate config carried on the struct.

## Testing

Added unit tests in `pkg/types/conf_test.go`:

- `LoadDelegateNetConfFromConfList preserves CNI-specific fields so DEL does not leak` — verifies plugin-specific fields (e.g. `kubeconfig`/`datastore_type`) survive into `DelegateNetConf.Bytes`.
- `LoadDelegateNetConfFromConfList keeps plugins appended from a subdirectory chain` — verifies plugins appended by libcni (not present in `confList.Bytes`) are retained.
- `InjectCNIVersionInConfList sets cniVersion without dropping fields` — verifies `cniVersion` backfill is lossless.

## Compatibility

No API or manifest changes. The fix only changes how delegate config bytes are reconstructed internally; delegates now receive the complete config on both ADD and DEL, which is the intended behavior.

## Summary by CodeRabbit

* **New Features**
  * Delegated CNI configuration handling now preserves plugin-specific settings more reliably, including nested Calico-related fields and appended plugins.
* **Bug Fixes**
  * Improved CNI version injection so `cniVersion` is applied when needed, including during delete, without stripping existing nested fields.
  * Added clearer error handling when conflist parsing fails.
* **Tests**
  * Added coverage for delegated conflist preservation and CNI version injection behavior.

### DIFF
--- a/pkg/multus/multus.go
+++ b/pkg/multus/multus.go
@@ -296,25 +296,16 @@ func confStatus(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.
 	return err
 }
 
-func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, cniConfList *libcni.NetworkConfigList, multusNetconf *types.NetConf, exec invoke.Exec) (cnitypes.Result, error) {
+func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *types.NetConf, exec invoke.Exec) (cnitypes.Result, error) {
 	logging.Debugf("conflistAdd: %v, %s", rt, string(rawnetconflist))
 	// In part, adapted from K8s pkg/kubelet/dockershim/network/cni/cni.go
 	binDirs := filepath.SplitList(os.Getenv("CNI_PATH"))
 	binDirs = append([]string{multusNetconf.BinDir}, binDirs...)
 	cniNet := libcni.NewCNIConfigWithCacheDir(binDirs, multusNetconf.CNIDir, exec)
 
-	var confList *libcni.NetworkConfigList
-	var err error
-
-	// This may wind up being set during parsing the default network config.
-	// In this case -- we'll use it as passed. Otherwise, we'll recalculate it.
-	if len(cniConfList.Plugins) > 0 {
-		confList = cniConfList
-	} else {
-		confList, err = libcni.NetworkConfFromBytes(rawnetconflist)
-		if err != nil {
-			return nil, logging.Errorf("conflistAdd: error converting the raw bytes into a conflist: %v", err)
-		}
+	confList, err := libcni.NetworkConfFromBytes(rawnetconflist)
+	if err != nil {
+		return nil, logging.Errorf("conflistAdd: error converting the raw bytes into a conflist: %v", err)
 	}
 
 	result, err := cniNet.AddNetworkList(context.Background(), confList, rt)
@@ -436,8 +427,7 @@ func DelegateAdd(exec invoke.Exec, kubeClient *k8s.ClientInfo, pod *v1.Pod, dele
 	var result cnitypes.Result
 	var err error
 	if delegate.ConfListPlugin {
-		// TODO: why are we passing bytes here? don't we have a better representation of it?
-		result, err = conflistAdd(rt, delegate.Bytes, &delegate.CNINetworkConfigList, multusNetconf, exec)
+		result, err = conflistAdd(rt, delegate.Bytes, multusNetconf, exec)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


FlyGoat:
Possibly duplication with #1450

pkg/multus/multus.go
```diff
@@ -296,25 +296,18 @@ func confStatus(rt *libcni.RuntimeConf, rawNetconf []byte, multusNetconf *types.
 	return err
 }
 
-func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, cniConfList *libcni.NetworkConfigList, multusNetconf *types.NetConf, exec invoke.Exec) (cnitypes.Result, error) {
-	logging.Debugf("conflistAdd: %v, %s", rt, string(rawnetconflist))
+func conflistAdd(rt *libcni.RuntimeConf, rawnetconflist []byte, multusNetconf *types.NetConf, exec invoke.Exec) (cnitypes.Result, error) {
+	// Log only the byte size, not the raw conflist: it now preserves
+	// plugin-specific fields that may contain sensitive delegate config.
+	logging.Debugf("conflistAdd: %v, raw conflist bytes length: %d", rt, len(rawnetconflist))
```
wizhaoredhat: I think we should keep the original log function. I am unsure what sensitive delegate config there is and since this is a debug log and is logged not towards the pod but on Multus; I believe we should log as much as possible that is useful.

wizhaoredhat (changes requested):

wizhaoredhat:
@chent1996  Could you sign your commits? (e.g. git commit --signoff)

pkg/multus/multus.go
chent1996: Agreed — restored the original debug log in `conflistAdd` (the change came from a CodeRabbit suggestion, but the full payload is indeed more useful for a node-side debug log). Also reworded that commit message accordingly.

chent1996:
@wizhaoredhat Done — rebased with `--signoff`, both commits now carry the Signed-off-by trailer.

wizhaoredhat (approved):

wizhaoredhat:
LGTM

PR #1512, opened 2026-06-02
Bump Kubernetes dependencies to 1.36.2

miheer:
Bump Kubernetes to 1.36.2 and Go to 1.26
bump kubernetes to 1.36.2 and Go to 1.26

commands run:

go get k8s.io/api@v0.36.2 k8s.io/apimachinery@v0.36.2 k8s.io/client-go@v0.36.2 k8s.io/kubelet@v0.36.2

go mod tidy

go mod vendor

go test ./cmd/cert-approver

go test ./pkg/k8sclient

go test ./pkg/checkpoint ./pkg/cmdutils ./pkg/logging

GOOS=linux make build

podman run --rm --privileged -v /Users/misalunk/go/src/github.com/k8snetworkplumbingwg/multus-cni:/workspace:Z -w /workspace -e GO111MODULE=on -e GOFLAGS="-mod=vendor -p=1" -e GOCACHE=/tmp/go-build-cache registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 bash -lc 'mkdir -p /etc/cni/net.d /var/lib/cni/multus/results && printf "package main\nfunc main(){}\n" > /tmp/true.go && CGO_ENABLED=0 go build -o /tmp/truebin /tmp/true.go && cp /tmp/truebin /usr/bin/true && ./hack/test-go.sh'

PR #1524, opened 2026-07-21
Followup PR for k8snetworkplumbingwg/multus-cni#1512 to validate command file paths after Kubernetes bump

miheer:
This is a follow-up to k8snetworkplumbingwg/multus-cni/#1512 for issues found while carrying the Kubernetes and Go bump into downstream.

The dependency verifier reported vendor drift for vendor/k8s.io/api/scheduling/v1alpha2/generated.proto after go mod tidy and go mod vendor. Re-running those commands on this upstream branch produced no vendor diff, so no vendor file change is included.

The security scan reported path traversal findings in kubeconfig_generator and multus-daemon because CLI-controlled paths were passed directly to file operations. This change validates those paths, requires absolute paths, rejects empty, relative, and parent-directory paths, and uses os.Root with local file names for file reads, writes, creates, and removes.

Commands run:

go mod tidy

go mod vendor

go test ./pkg/cmdutils ./cmd/kubeconfig_generator

GOOS=linux go test -c -o /private/tmp/multus-daemon.test ./cmd/multus-daemon

podman run --rm -v /Users/misalunk/go/src/github.com/k8snetworkplumbingwg/multus-cni:/workspace:Z -w /workspace -e GOCACHE=/tmp/go-build-cache -e GOFLAGS=-mod=vendor registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 go test ./pkg/cmdutils ./cmd/kubeconfig_generator ./cmd/multus-daemon

## Summary by CodeRabbit

* **Security and Reliability**
  * Improved validation and safe handling of kubeconfig and CNI configuration paths.
  * Rejected unsafe, relative, empty, and path-traversal inputs.
  * Improved configuration file copying, resource cleanup, and shutdown handling.
* **Bug Fixes**
  * Added clearer error diagnostics when configuration files cannot be read or written.
  * Ensured generated kubeconfig files use secure file permissions.
* **Tests**
  * Added coverage for secure path handling, file copying, kubeconfig generation, file permissions, and default socket directory behavior.

### DIFF
--- a/cmd/kubeconfig_generator/main.go
+++ b/cmd/kubeconfig_generator/main.go
@@ -21,13 +21,13 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"syscall"
 	"text/template"
 	"time"
 
 	"github.com/spf13/pflag"
 
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/cmdutils"
 	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/k8sclient"
 
 	"k8s.io/client-go/tools/clientcmd"
@@ -63,16 +63,17 @@ func main() {
 	certDurationString := pflag.StringP("cert-duration", "", "10m", "specify certificate duration")
 	helpFlag := pflag.BoolP("help", "h", false, "show help message and quit")
 
-	kubeconfigPath, err := filepath.Abs(*kubeconfigPathRaw)
-	if err != nil {
-		klog.Fatalf("illegal path %s in kubeconfigPath %s: %v", kubeconfigPath, *kubeconfigPathRaw, err)
-	}
-
 	pflag.Parse()
 	if *helpFlag {
 		pflag.PrintDefaults()
 		os.Exit(1)
 	}
+
+	kubeconfigPath, err := cmdutils.NewRootedFile(*kubeconfigPathRaw)
+	if err != nil {
+		klog.Fatalf("illegal path in kubeconfigPath %s: %v", *kubeconfigPathRaw, err)
+	}
+	defer kubeconfigPath.Close()
 
 	// check variables
 	if _, err := os.Stat(*bootstrapConfig); err != nil {
@@ -108,9 +109,9 @@ func main() {
 		klog.Fatalf("failed to start cert manager: %v", err)
 	}
 
-	fp, err := os.OpenFile(kubeconfigPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	fp, err := kubeconfigPath.Root.OpenFile(kubeconfigPath.FileName, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
-		klog.Fatalf("cannot create kubeconfig file %q: %v", kubeconfigPath, err)
+		klog.Fatalf("cannot create kubeconfig file %q: %v", kubeconfigPath.Path(), err)
 	}
 
 	// render kubeconfig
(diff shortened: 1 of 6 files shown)


thomasferrandiz:
@miheer can you fix coderabbit's comment?

Then I thing we're good to merge the PR.

miheer:
> @miheer can you fix coderabbit's comment?
> 
> Then I thing we're good to merge the PR.

Done

miheer:
@thomasferrandiz PTAL

cmd/kubeconfig_generator/main.go
```diff
@@ -63,17 +63,18 @@ func main() {
...
 	}
 
+	kubeconfigPath, err := cmdutils.NewRootedFile(*kubeconfigPathRaw)
+	if err != nil {
+		klog.Fatalf("illegal path in kubeconfigPath %s: %v", *kubeconfigPathRaw, err)
+	}
+	defer kubeconfigPath.Close()
+
 	// check variables
 	if _, err := os.Stat(*bootstrapConfig); err != nil {
```
wizhaoredhat: Why are we still using os.* calls here. "os.Stat(*bootstrapConfig)". Shouldn't we have a unified way to access the FS and not a mix of os.* and filepath.*

cmd/kubeconfig_generator/main.go
```diff
@@ -108,9 +109,9 @@ func main() {
 		klog.Fatalf("failed to start cert manager: %v", err)
 	}
 
-	fp, err := os.OpenFile(kubeconfigPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
+	fp, err := kubeconfigPath.Root.OpenFile(kubeconfigPath.FileName, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
-		klog.Fatalf("cannot create kubeconfig file %q: %v", kubeconfigPath, err)
+		klog.Fatalf("cannot create kubeconfig file %q: %v", kubeconfigPath.Path(), err)
 	}
```
wizhaoredhat: Shouldn't we defer closing the file pointer?

miheer: Good catch. I updated this in a separate commit.

The kubeconfig write path now uses writeKubeconfig(), which opens the file through the rooted path and immediately defers fp.Close(). That way the file is closed even if template parsing or execution returns an error.

I also added a focused test for the kubeconfig write helper.

cmd/kubeconfig_generator/main.go
miheer: Good point. I updated this to use the same rooted filesystem pattern for the remaining CLI paths in kubeconfig_generator.

Specifically:
1. Reused NewRootedFile for bootstrap-config
2. Added NewRootedDir for certdir
3. Replaced the direct os.Stat calls with Root.Stat

So kubeconfig, bootstrap-config, and certdir are now all validated through cmdutils before filesystem access.

wizhaoredhat (approved):
LGTM

wizhaoredhat:
@miheer Could you look at the code rabbit suggestions?

miheer:
> @miheer Could you look at the code rabbit suggestions?

done @wizhaoredhat @bpickard22 PTAL

cmd/kubeconfig_generator/main_test.go
```diff
@@ -14,79 +14,102 @@
...
 
+// disable dot-imports only for testing
+//revive:disable:dot-imports
 import (
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 
+	. "github.com/onsi/ginkgo/v2"
```
wizhaoredhat: Why do we have dot imports? Can't we have named imports?

miheer: Fixed in a separate commit.

I replaced the Ginkgo/Gomega dot imports with named imports and removed the dot-import lint exception.

So instead of importing the packages with "." and calling the helpers directly, the test now uses explicit package names, for example:

ginkgo.Describe(...)
ginkgo.It(...)
ginkgo.DeferCleanup(...)
gomega.Expect(...)
gomega.HaveOccurred(...)
gomega.Succeed(...)

This keeps the test behavior the same, but makes it clear where each test helper comes from.

wizhaoredhat (approved):
LGTM

thomasferrandiz (approved):

miheer:
@thomasferrandiz @bpickard22 can we please re-run this test or over-ride it ? the error does not seems related to the fix in this PR.

PR #1525, opened 2026-07-24
Thick plugin graceful termination

wizhaoredhat:
This revives the stale PR https://github.com/k8snetworkplumbingwg/multus-cni/pull/1338

This PR introduces graceful shutdown functionality to the Multus daemon by adding a /readyz endpoint alongside the existing /healthz. The /readyz endpoint starts returning 500 once a SIGTERM is received, indicating the daemon is in shutdown mode. During this time, CNI requests can still be processed for a short window. The daemonset configs have been updated to increase terminationGracePeriodSeconds from 10 to 30 seconds, ensuring we have a bit more time for these clean shutdowns.

This addresses a race condition during pod transitions where the readiness check might return true, but a subsequent CNI request could fail if the daemon shuts down too quickly. By introducing the /readyz endpoint and delaying the shutdown, we can handle ongoing CNI requests more gracefully, reducing the risk of disruptions during critical transitions.

Major thanks to @deads2k for the find, identification, fix, and of course, the explanations. Appreciate it.

## Summary by CodeRabbit

* **New Features**
  * Added a readiness endpoint that reports when the service is entering graceful shutdown.
  * Readiness polling now uses the dedicated readiness check, while health checks remain available for ongoing operations.

* **Bug Fixes**
  * Improved graceful shutdown handling by allowing active operations time to complete.
  * Extended the pod termination grace period from 10 to 30 seconds for more reliable shutdowns.
  * Cleanup operations can continue while the service is gracefully shutting down.

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -66,11 +66,9 @@ func main() {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	sigTermCtx, sigTermCancel := context.WithCancel(ctx)
+	// Used by the server readiness handler to report not-ready once shutdown is initiated.
 	isInGracefulShutdownMode := func() bool {
-		if sigTermCtx.Err() == nil {
-			return false
-		}
-		return true
+		return sigTermCtx.Err() != nil
 	}
 
 	daemonConf, err := cniServerConfig(*configFilePath)
@@ -138,6 +136,7 @@ func main() {
 	go func() {
 		for sig := range signalCh {
 			logging.Verbosef("caught %v, stopping...", sig)
+			// Mark graceful-shutdown mode first so readiness probes fail before canceling server context.
 			sigTermCancel()
 			<-time.After(SigTermCancelAfter)
 			cancel()
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -44,6 +44,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
+// SigtermCancelAfter sets the wait time to cancel after sig term
+// TODO: This could be a configuration option
+const SigTermCancelAfter = 10 * time.Second
+
 func main() {
 	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
@@ -61,6 +65,13 @@ func main() {
 
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
+	sigTermCtx, sigTermCancel := context.WithCancel(ctx)
+	isInGracefulShutdownMode := func() bool {
+		if sigTermCtx.Err() == nil {
+			return false
+		}
+		return true
+	}
 
 	daemonConf, err := cniServerConfig(*configFilePath)
 	if err != nil {
@@ -109,7 +120,7 @@ func main() {
 		}
 	}
 
-	if err := startMultusDaemon(ctx, daemonConf, ignoreReadinessIndicator); err != nil {
+	if err := startMultusDaemon(ctx, daemonConf, ignoreReadinessIndicator, isInGracefulShutdownMode); err != nil {
 		logging.Panicf("failed start the multus thick-plugin listener: %v", err)
 		os.Exit(3)
 	}
@@ -127,6 +138,8 @@ func main() {
 	go func() {
 		for sig := range signalCh {
 			logging.Verbosef("caught %v, stopping...", sig)
+			sigTermCancel()
+			<-time.After(SigTermCancelAfter)
 			cancel()
 		}
 	}()
@@ -143,7 +156,7 @@ func main() {
 	logging.Verbosef("multus daemon is exited")
 }
 
-func startMultusDaemon(ctx context.Context, daemonConfig *srv.ControllerNetConf, ignoreReadinessIndicator bool) error {
+func startMultusDaemon(ctx context.Context, daemonConfig *srv.ControllerNetConf, ignoreReadinessIndicator bool, isInGracefulShutdownMode func() bool) error {
 	if user, err := user.Current(); err != nil || user.Uid != "0" {
 		return fmt.Errorf("failed to run multus-daemon with root: %v, now running in uid: %s", err, user.Uid)
 	}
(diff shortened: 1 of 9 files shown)


cmd/multus-daemon/main.go
```diff
@@ -61,6 +65,13 @@ func main() {
 
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
+	sigTermCtx, sigTermCancel := context.WithCancel(ctx)
+	isInGracefulShutdownMode := func() bool {
+		if sigTermCtx.Err() == nil {
+			return false
+		}
+		return true
+	}
```
SchSeba, line 72: lets simplify this by 

```
isInGracefulShutdownMode := func() bool {
    return sigTermCtx.Err() != nil
}
```

also please add a comment to explain how we use this

cmd/multus-daemon/main.go
```diff
@@ -127,6 +138,8 @@ func main() {
 	go func() {
 		for sig := range signalCh {
 			logging.Verbosef("caught %v, stopping...", sig)
+			sigTermCancel()
```
SchSeba, line 140: add a comment here to explain why we do this please

cmd/multus-daemon/main.go
```diff
@@ -127,6 +138,8 @@ func main() {
...
+			<-time.After(SigTermCancelAfter)
 			cancel()
```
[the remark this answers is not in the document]
SchSeba: I think we can skip this commit this is running inside a pod so I don't see when we can get a SIGINT

pkg/server/server.go
```diff
@@ -351,6 +351,23 @@ func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec, s
 			w.Header().Set("Content-Type", "application/json")
 		})))
 
+	// handle for '/readyz'
+	router.HandleFunc(api.MultusReadyAPIEndpoint, promhttp.InstrumentHandlerCounter(s.metrics.requestCounter.MustCurryWith(prometheus.Labels{"handler": api.MultusHealthAPIEndpoint}),
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if r.Method != http.MethodGet && r.Method != http.MethodPost {
+				http.Error(w, fmt.Sprintf("Method not allowed"), http.StatusMethodNotAllowed)
```
SchSeba: we don't need `fmt.Sprintf` here

pkg/server/api/api.go
```diff
@@ -106,7 +109,7 @@ func CreateDelegateRequest(cniCommand, cniContainerID, cniNetNS, cniIFName, podN
 // WaitUntilAPIReady checks API readiness
 func WaitUntilAPIReady(socketPath string) error {
 	return utilwait.PollImmediate(APIReadyPollDuration, APIReadyPollTimeout, func() (bool, error) {
-		_, err := DoCNI(GetAPIEndpoint(MultusHealthAPIEndpoint), nil, SocketPath(socketPath))
+		_, err := DoCNI(GetAPIEndpoint(MultusReadyAPIEndpoint), nil, SocketPath(socketPath))
```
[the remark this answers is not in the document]
SchSeba: this is expected as we don't want to accept any new CNI_ADD calls but still allow CMD_DEL to pass

SchSeba (commented):
in general the PR looks good but can you please add some until tests that will validate this new function

also please check the CMD_DEL is using the other entry is that expected?

this means that we can have new CNI_DEL calls that will get break in the middle of the run, should we also use the reject entry for that?

wizhaoredhat:
> also please check the CMD_DEL is using the other entry is that expected?

>this means that we can have new CNI_DEL calls that will get break in the middle of the run, should we also use the >reject entry for that?

@SchSeba 
I think it is intentional keeping DEL on /healthz since we never want to reject DEL even during shutdown to minimize orphaned netdevs.

I will add a comment

pkg/server/server.go
wizhaoredhat: Done!

cmd/multus-daemon/main.go
wizhaoredhat: Done!

cmd/multus-daemon/main.go
wizhaoredhat: Done!

pkg/server/server.go
```diff
@@ -351,6 +351,23 @@ func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec, s
...
+				return
+			}
+
+			if !isInGracefulShutdownMode() {
+				w.WriteHeader(http.StatusOK)
+				w.Header().Set("Content-Type", "application/json")
+			} else {
+				w.WriteHeader(http.StatusInternalServerError)
+				w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
wizhaoredhat: Done!

pkg/server/server.go
```diff
@@ -351,6 +351,23 @@ func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec, s
 			w.Header().Set("Content-Type", "application/json")
 		})))
 
+	// handle for '/readyz'
+	router.HandleFunc(api.MultusReadyAPIEndpoint, promhttp.InstrumentHandlerCounter(s.metrics.requestCounter.MustCurryWith(prometheus.Labels{"handler": api.MultusHealthAPIEndpoint}),
```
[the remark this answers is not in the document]
wizhaoredhat: Done!

pkg/server/api/api_test.go
```diff
@@ -0,0 +1,68 @@
...
+	tmpDir := t.TempDir()
+	pathCh := make(chan string, 1)
+
+	listener, err := net.Listen("unix", SocketPath(tmpDir))
+	if err != nil {
+		t.Fatalf("failed to listen on unix socket: %v", err)
+	}
+	t.Cleanup(func() {
+		_ = listener.Close()
+	})
```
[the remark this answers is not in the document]
wizhaoredhat: done.

vinnie1110 (approved):
/lgtm

pkg/server/server.go
```diff
@@ -343,14 +343,31 @@ func newCNIServer(rundir string, kubeClient *k8s.ClientInfo, exec invoke.Exec, s
...
+				http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
+				return
+			}
+
+			if !isInGracefulShutdownMode() {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusOK)
+			} else {
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusInternalServerError)
```
dasanind, line 367: Nice to have: The `/readyz` handler returns `http.StatusInternalServerError` with no body. I think it will be nice to have a response body with a JSON message like `{"status": "graceful shut down is in progress"}`. This will help in debugging.

dasanind (commented):
The PR looks good to me. Minor nice to have comment.

pkg/server/server.go
wizhaoredhat: The readyz endpoint only has 2 possible outcomes, i think it is ok to not write anything. If in error, it is in shutdown. If not, everything is normal.

bpickard22: agree

PR #1539, opened 2026-08-27
Sort device id per container

cathy-zhou:
## Summary by CodeRabbit

* **Bug Fixes**
  * Device IDs are now sorted consistently within each container while preserving container order.
  * Device assignments from multiple entries are correctly combined without reordering devices across containers.
  * Input device lists remain unchanged during processing.

* **Documentation**
  * Added documentation explaining device ID collection, ordering, and delegate configuration.
  * Added the device resource assignment guide to the comprehensive documentation index.

### DIFF
--- a/pkg/checkpoint/checkpoint.go
+++ b/pkg/checkpoint/checkpoint.go
@@ -102,10 +102,11 @@ func (cp *checkpoint) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resource
 				entry = &types.ResourceInfo{}
 				resourceMap[pod.ResourceName] = entry
 			}
+			var ids []string
 			for _, v := range pod.DeviceIDs {
-				// already exists; append to it
-				entry.DeviceIDs = append(entry.DeviceIDs, v...)
+				ids = append(ids, v...)
 			}
+			entry.DeviceIDs = append(entry.DeviceIDs, types.CopyAndSortDeviceIDs(ids)...)
 		}
 	}
 	return resourceMap, nil
--- a/pkg/checkpoint/checkpoint_test.go
+++ b/pkg/checkpoint/checkpoint_test.go
@@ -132,12 +132,130 @@ var _ = Describe("Kubelet checkpoint data read operations", func() {
 			Expect(len(resourceInfo.DeviceIDs)).To(BeEquivalentTo(2))
 		})
 
-		It("should have \"0000:03:02.3\" in deviceIDs[0]", func() {
-			Expect(resourceInfo.DeviceIDs[0]).To(BeEquivalentTo("0000:03:02.3"))
+		// Single-container allocation is sorted; container order is preserved across entries.
+		It("should have \"0000:03:02.0\" in deviceIDs[0] (per-container sorted)", func() {
+			Expect(resourceInfo.DeviceIDs[0]).To(BeEquivalentTo("0000:03:02.0"))
 		})
 
-		It("should have \"0000:03:02.0\" in deviceIDs[1]", func() {
-			Expect(resourceInfo.DeviceIDs[1]).To(BeEquivalentTo("0000:03:02.0"))
+		It("should have \"0000:03:02.3\" in deviceIDs[1] (per-container sorted)", func() {
+			Expect(resourceInfo.DeviceIDs[1]).To(BeEquivalentTo("0000:03:02.3"))
+		})
+	})
+
+	Context("Using a checkpoint file with multiple containers sharing a resource name", func() {
+		const multiContainerFile = "/tmp/kubelet_internal_checkpoint_multi_container"
+
+		var (
+			fakeCp      *fakeCheckpoint
+			resourceMap map[string]*types.ResourceInfo
+		)
+
+		BeforeEach(func() {
+			// appcntr1 and appcntr2 both get devices from intel.com/sriov_net_A,
+			// and a third entry belongs to a different pod.
+			sampleData := `{
+				"Data": {
+					"PodDeviceEntries": [
+					{
+						"PodUID": "8f5ba1b4-cc11-11e8-89df-408d5c537d23",
+						"ContainerName": "appcntr1",
+						"ResourceName": "intel.com/sriov_net_A",
+						"DeviceIDs": {"-1": [
+							"0000:03:02.3",
+							"0000:03:02.0"
+							]
+						},
+						"AllocResp": ""
+					},
+					{
+						"PodUID": "8f5ba1b4-cc11-11e8-89df-408d5c537d23",
+						"ContainerName": "appcntr2",
+						"ResourceName": "intel.com/sriov_net_A",
+						"DeviceIDs": {"-1": [
+							"0000:03:02.5",
+							"0000:03:02.1"
+							]
+						},
+						"AllocResp": ""
+					},
+					{
+						"PodUID": "8f5ba1b4-cc11-11e8-89df-408d5c537d23",
+						"ContainerName": "appcntr2",
+						"ResourceName": "intel.com/sriov_net_B",
+						"DeviceIDs": {
+							"1": ["0000:03:06.3"],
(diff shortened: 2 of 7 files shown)


cathy-zhou:
@coderabbitai please review

pkg/kubeletclient/kubeletclient.go
```diff
@@ -140,16 +140,19 @@ func (rc *kubeletClient) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resou
...
+		// Copy+sort this container's IDs only, then append. That keeps
+		// container order for NAD assignment while making order inside
+		// one container deterministic (e.g. UDN "last device").
+		deviceIDs := types.CopyAndSortDeviceIDs(dev.DeviceIds)
 		if rInfo, ok := resourceMap[dev.ResourceName]; ok {
-			rInfo.DeviceIDs = append(rInfo.DeviceIDs, dev.DeviceIds...)
+			rInfo.DeviceIDs = append(rInfo.DeviceIDs, deviceIDs...)
 		} else {
-			resourceMap[dev.ResourceName] = &types.ResourceInfo{DeviceIDs: dev.DeviceIds}
+			resourceMap[dev.ResourceName] = &types.ResourceInfo{DeviceIDs: deviceIDs}
```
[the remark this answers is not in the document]
cathy-zhou: it's not the contract we're aiming for here.
Kubelet almost always emits one ContainerDevices row per (container, resourceName). We sort that row, then append in container order so:
1. VFs on one container are in a stable order (UDN "last device" when they're all on one container).
2. We do not re-sort the pod-wide merged list (that was #1481 and broke multi-container SR-IOV vs PCIDEVICE_*).
If kubelet ever splits one container's allocation into multiple rows for the same ResourceName, we keep each row sorted and concatenate. That order is still deterministic; it just isn't a single lexicographic sort of the whole container.

cathy-zhou:
@coderabbitai please take a look of all your comments are addressed, mark them as resolved if yes.

pkg/checkpoint/checkpoint.go
```diff
@@ -102,10 +102,11 @@ func (cp *checkpoint) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resource
 				entry = &types.ResourceInfo{}
 				resourceMap[pod.ResourceName] = entry
 			}
+			var ids []string
 			for _, v := range pod.DeviceIDs {
-				// already exists; append to it
-				entry.DeviceIDs = append(entry.DeviceIDs, v...)
+				ids = append(ids, v...)
 			}
+			entry.DeviceIDs = append(entry.DeviceIDs, types.CopyAndSortDeviceIDs(ids)...)
```
wizhaoredhat, line 109: The `for _, pod := range cp.podEntires` is iterating through every checkpoint entry. Not every container in a pod, this function doesn't do that. For the checkpoint everything should be sorted accordingly right?

pkg/kubeletclient/kubeletclient.go
```diff
@@ -144,11 +144,26 @@ func (rc *kubeletClient) GetPodResourceMap(pod *v1.Pod) (map[string]*types.Resou
...
-		if rInfo, ok := resourceMap[dev.ResourceName]; ok {
-			rInfo.DeviceIDs = append(rInfo.DeviceIDs, dev.DeviceIds...)
+		if _, ok := seen[dev.ResourceName]; !ok {
+			seen[dev.ResourceName] = struct{}{}
+			resourceOrder = append(resourceOrder, dev.ResourceName)
+		}
+		aggregated[dev.ResourceName] = append(aggregated[dev.ResourceName], dev.DeviceIds...)
+	}
+	for _, name := range resourceOrder {
+		deviceIDs := types.CopyAndSortDeviceIDs(aggregated[name])
```
wizhaoredhat, line 162: I don't understand how this works... isn't dev.ResourceName key at the pod level? There is nothing here at the container level.

pkg/checkpoint/checkpoint.go
wizhaoredhat, line 90: The issue is "map[string]*types.ResourceInfo", we need a new data type to handle this case. Same with `addDevicePluginResources` function below, it needs a new datatype and not the flat list.

https://github.com/k8snetworkplumbingwg/multus-cni/compare/master...wizhaoredhat:k8snetworkplumbingwg-multus-cni:sort_devices_per_container is my attempt at this changing the flat list. But as you can tell this is a massive change.

pkg/kubeletclient/kubeletclient.go
cathy-zhou: The container boundary is in the caller: GetPodResourceMap calls getDevicePluginResources(cnt.Devices, ...) once per container. So devices is a single container's devices — aggregated/seen just sort that container's IDs per resource, and the outer container loop preserves container order by appending.

pkg/checkpoint/checkpoint.go
cathy-zhou: Each checkpoint entry is keyed by (PodUID, ContainerName, ResourceName), so one entry == one container's allocation. Per-entry sort here is therefore the same per-container sort as kubeletclient.go. A global sort would reorder across containers (the reverted behavior) and diverge from the kubelet backend, so I kept it per-entry to keep the two backends consistent.

pkg/checkpoint/checkpoint.go
cathy-zhou: I'll take a look of your changes and it is probably longer term model. 

But this PR is intentionally minimal: restore deterministic ordering for the single-container primary-UDN case (#1481) without re-triggering the revert regression. That regression came from pod-wide sorting crossing container boundaries; here we sort only within each container and preserve container order, so no device crosses a boundary and each NAD still gets a VF from its own container (for single-VF-per-container it's a no-op = pre-#1481 order).

Do you have a specific multi-container case where the per-container-sorted flat list still breaks (e.g. multiple VFs per container where within-container order is load-bearing)?

wizhaoredhat: I should have reviewed it more deeply. But I think your answers above addresses my concerns.

pkg/checkpoint/checkpoint_test.go
```diff
@@ -132,12 +132,13 @@ var _ = Describe("Kubelet checkpoint data read operations", func() {
 			Expect(len(resourceInfo.DeviceIDs)).To(BeEquivalentTo(2))
 		})
 
-		It("should have \"0000:03:02.3\" in deviceIDs[0]", func() {
-			Expect(resourceInfo.DeviceIDs[0]).To(BeEquivalentTo("0000:03:02.3"))
+		// Single-container allocation is sorted; container order is preserved across entries.
```
wizhaoredhat, line 135: Can we also add a new unit test where the checkpoint file has multiple contains with multiple devices. (the use case that regressed in the past).

wizhaoredhat (changes requested):
I think in general this approach is the way to go after considering the entire E2E path with OVN-K.

Couple things we want to do.
1) Let's document this behavior. I think a new md file in docs. Give an explanation of the use case (Let's not mention OVN-K here). I believe there are cases where multi networking falls short of the current APIs. E.g. where default networks and secondary networks fall short.
2) Please try to attend the next community meeting Thursday, 3 September at 9:30 – 10:30am EST. The purpose is to inform the community of this multus behavior change.

pkg/checkpoint/checkpoint_test.go
cathy-zhou: @wizhaoredhat updated the PR, please take a look. I will try to attend the community meeting Thursday, 3 September, but probably only the first 30 minutes.

wizhaoredhat (approved):

wizhaoredhat:
@SchSeba This PR doesn't have an effect on DRA. If we want OVN-K to move to DRA from DP then we probably need to sort the same way this PR does. draClient.GetPodResourceMap() would need to be changed which remains unchanged ATM.

Issue #1366, opened 2024-12-10
Multus pods always restart when the multus cni config has been specify


cyclinder:
**What happend**:

```
root@10-20-1-10:/home/cyclinder/multus4.0# kubectl get po -n kube-system
NAME                                           READY   STATUS             RESTARTS        AGE
calico-kube-controllers-86447f4445-pjhnz       1/1     Running            0               4d4h
calico-node-qc5cp                              1/1     Running            0               4d4h
calico-node-vn5bw                              1/1     Running            0               4d4h
coredns-76f75df574-9drmz                       1/1     Running            0               4d5h
coredns-76f75df574-xzkcz                       1/1     Running            0               4d5h
etcd-spider-control-plane                      1/1     Running            0               4d5h
kube-apiserver-spider-control-plane            1/1     Running            0               4d5h
kube-controller-manager-spider-control-plane   1/1     Running            0               4d5h
kube-multus-ds-jctrl                           0/1     CrashLoopBackOff   6 (2m26s ago)   8m18s
kube-multus-ds-jpgdr                           0/1     CrashLoopBackOff   6 (2m20s ago)   8m18s
kube-proxy-6csq8                               1/1     Running            0               4d5h
kube-proxy-z9dhd                               1/1     Running            0               4d5h
kube-scheduler-spider-control-plane            1/1     Running            0               4d5h
root@10-20-1-10:/home/cyclinder/multus4.0# kubectl logs -f -n kube-system kube-multus-ds-jctrl
Defaulted container "kube-multus" out of: kube-multus, install-multus-binary (init)
2024-12-10T08:55:01Z [verbose] multus-daemon started
2024-12-10T08:55:01Z [verbose] server configured with chroot: /hostroot
2024-12-10T08:55:01Z [verbose] Filtering pod watch for node "spider-worker"
2024-12-10T08:55:02Z [verbose] API readiness check
2024-12-10T08:55:02Z [verbose] API readiness check done!
2024-12-10T08:55:02Z [verbose] multus daemon is exited
```

**What you expected to happen**:

multus pods running well/

**How to reproduce it (as minimally and precisely as possible)**:

```
~# kubectl apply -f multus.yaml
~# cat multus.yaml
# Note:
#   This deployment file is designed for 'quickstart' of multus, easy installation to test it,
#   hence this deployment yaml does not care about following things intentionally.
#     - various configuration options
#     - minor deployment scenario
#     - upgrade/update/uninstall scenario
[...]

PR #1543, opened 2026-09-08
keep thick daemon running when multusConfigFile is a path

junkb:
setting multusConfigFile to a path makes the thick daemon exit 0
immediately after startup. see the commit message.

verified on a cri-o cluster: the daemon stays up, cni add and del work
through flannel and macvlan delegates, and sigterm exits cleanly.

#1366 may describe the same problem.

## Summary by CodeRabbit

- **Bug Fixes**
  - The daemon now remains running until it receives a termination signal, allowing it to shut down cleanly instead of exiting prematurely.

### DIFF
--- a/cmd/multus-daemon/main.go
+++ b/cmd/multus-daemon/main.go
@@ -151,6 +151,9 @@ func main() {
 		}
 	}
 
+	// run until signalled (the WaitGroup only waits for the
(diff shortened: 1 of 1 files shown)


junkb:
i just happened to notice #1508, which this looks to be a duplicate of, so i'll close this one

PR #1548, opened 2026-09-16
daemon: fix nil pointer dereference in the root privilege check

rollandf:
## What this PR does

`startMultusDaemon()` checks that it is running as root using `os/user.Current()`. When that call fails it returns a **nil** `*User` alongside the error, but the guard dereferences `user.Uid` while building the error message. The daemon therefore crashes with SIGSEGV instead of reporting the problem:

```
[verbose] multus-daemon started
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
main.startMultusDaemon(...)
        /usr/src/multus-cni/cmd/multus-daemon/main.go:160
```

This replaces the check with `os.Geteuid()`.

## Why `user.Current()` fails

The daemon is built with `CGO_ENABLED=0` (`hack/build-go.sh:60`), so `os/user` uses the pure-Go implementation, which resolves the uid by parsing `/etc/passwd` and only falls back to `$USER`/`$HOME` if that lookup misses. Any image whose `/etc/passwd` has no entry for the running uid hits the nil dereference. Note that setting only `$HOME` is not enough, the fallback requires both `$USER` and `$HOME`.

## Reproduction

This reproduces on the currently shipped thick image base, so it is not specific to any particular distribution:

```
$ podman run --rm --user 1000:1000 --passwd=false \
    --entrypoint /multus-daemon debian:stable-slim
user.Current() -> user=<nil> err=user: Current requires cgo or $USER, $HOME set in environment
panic: runtime error: invalid memory address or nil pointer dereference
```

(`--passwd=false` stops podman from synthesising a passwd entry, which matches how CRI-O and containerd set the uid.)

## Why `os.Geteuid()`

- It is a syscall, so it has no `/etc/passwd` or environment dependency.
- It returns the *effective* uid, which is what actually governs the `chroot` performed in `pkg/server/exec_chroot.go`, rather than whether a username happens to resolve.
- It removes the nil dereference, so the intended message is printed:

```
failed to run multus-daemon with root, now running in uid: 1000
```

## Testing

- `gofmt`, `go vet ./cmd/multus-daemon/...` and `go build ./cmd/...` are clean.
- Built the thick image and confirmed the daemon starts as uid 0 and is cleanly rejected (no panic) as uid 1000.
- No new unit test: exercising the failure path requires running the process as non-root, which does not fit the existing suite.

## Summary by CodeRabbit

* **Bug Fixes**
  * The Multus daemon now clearly reports an error when started without root privileges, including the detected user ID.


almaslennikov (approved):

e0ne (approved):
LGTM

